PR #19, opened 2018-07-03
Using os.sep instead of "\\"

bb511:
This is the first PR

Problem with importing the file is fixed if os.sep is used instead of "\\" at https://github.com/mantidproject/fitbenchmarking/blob/29eaecbf601dd392dc82d2d8844f871ebb5ba78e/fitbenchmarking/input_parsing.py#L226

To test:
1. Make sure that run_data = "neutron" in example_runScript.py
2. Run example_runScript in the mantidpython console
3. Check if the right neutron files are imported (i.e. compare console output with the files in the fitbenchmarking/benchmark_problems/Neutron_data)

### DIFF
--- a/fitbenchmarking/example_runScripts.py
+++ b/fitbenchmarking/example_runScripts.py
@@ -1,4 +1,3 @@
-
 # script for running fit benchmarking and comparising the relative performance of local minimzers on
 # fit problems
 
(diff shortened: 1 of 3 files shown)


fitbenchmarking/input_parsing.py
```diff
@@ -218,16 +218,20 @@ def load_neutron_data_fitting_problem_file(fname):
 
     @param fname :: name of the file to load
     """
+
     with open(fname) as probf:
+
         entries = get_neutron_data_problem_entries(probf)
-        # get the path to the data
```
Anders-Markvardsen: Please keep comment. In fact I would add to it explaining in code that data in data_files is will be added to the data search path

fitbenchmarking/example_runScripts.py
```diff
@@ -7,25 +6,22 @@
 
 import os
 
-#import mantid.simpleapi as msapi
+# Import mantid.simpleapi as msapi
```
Anders-Markvardsen: probably delete?

fitbenchmarking/input_parsing.py
```diff
@@ -218,16 +218,20 @@ def load_neutron_data_fitting_problem_file(fname):
...
-        k=-1
-        prefix =""
-        k = fname.rfind("\\")
+
+        k = -1
```
Anders-Markvardsen: explain what 'k' is?

Issue #22, opened 2018-07-04
Neutron visual display page produces a crash


bb511:
### Expected behavior
When calling build_visual_display_page with group_name = 'neutron_data', the script should produce a visual display page.
https://github.com/mantidproject/fitbenchmarking/blob/730de95477d74c34fd5bdad893fe643b6820dbb1/fitbenchmarking/results_output.py#L145-L154
### Actual behavior
Instead of producing a visual display page, the script crashes.
### Steps to reproduce the behavior
In `example_runScript.py` set:

- run_data = "neutron"
- group_names = "Neutron data"
- group_suffix_names = "neutron_data"

Run `example_runScript` in the mantidpython console. The application should crash after a while.
### Platforms affected
Ubuntu 16.04, Windows

PR #24, opened 2018-07-05
Output nan in the table when fitting fails

bb511:
This comes after #18 
The ValueError that is raised by the Simplex minimizer in problem WISH17701 panel 103 tube 3 calibration, peak 9 1 is now caught and a nan value is outputted in the table.

To test:

1. In example_runScript.py make sure
    - minimizers contains 'Simplex'
    - run_data = 'neutron'
2. Run example_runScript.py in the mantidpython terminal
3. Check the table comparison_weighted_v3.8_acc_nist_lower.html under column Simplex, last row, should display nan

Notes: 
- The table name is wrong, this is a known issue, changing it to neutron causes a crash #22 .
- This only catches RuntimeErrors and ValueErrors, if the fit algorithm returns any other kind of error, the problem is still there. Doing an except that catches everything is generally bad practice so I refrained from doing so here. In this case, other kind of error should be added as they are discovered.

### DIFF
--- a/fitbenchmarking/fitting_benchmarking.py
+++ b/fitbenchmarking/fitting_benchmarking.py
@@ -182,9 +182,10 @@ def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0,
                     best_fit=data(minimizer_name,tmp.readX(1),tmp.readY(1))
                     min_sum_err_sq=sum_err_sq
             else:
-                sum_err_sq = float("inf")
+                sum_err_sq = float("nan")
                 print(" WARNING: no output fit workspace")
-            print("   sum sq: {0}".format(sum_err_sq))
+            print("sum sq: {0}".format(sum_err_sq))
+
             result = test_result.FittingTestResult()
             result.problem = prob
             result.fit_status = status
@@ -196,7 +197,9 @@ def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0,
             result.runtime = t_end - t_start if not np.isnan(chi2) else np.nan
             print("Result object: {0}".format(result))
             results_problem_start.append(result)
+
         results_fit_problem.append(results_problem_start)
+
         # make plots
         fig=plot()
         best_fit.markers=''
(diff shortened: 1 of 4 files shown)


fitbenchmarking/fitting_benchmarking.py
```diff
@@ -168,35 +168,40 @@ def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0,
...
-                    min_sum_err_sq=sum_err_sq
-            else:
-                sum_err_sq = float("inf")
-                print(" WARNING: no output fit workspace")
-            print("   sum sq: {0}".format(sum_err_sq))
+            
+            sum_err_sq = -1
+            if not status == 'failed':
+                print("   params: {0}, errors: {1}".format(params, errors))
+                def sum_of_squares(values):
```
AnthonyLim23: can we have this function definition outside of this function? I was a little surprised to find it in here.

fitbenchmarking/results_output.py
```diff
@@ -99,6 +99,7 @@ def print_group_results_tables(minimizers, results_per_test, problems_obj, group
                                metric_type=FILENAME_SUFFIX_ACCURACY, file_extension=FILENAME_EXT_HTML)
 
         # print out accuracy summary table for this group of fit problems
+        # THIS IS BROKEN
```
AnthonyLim23: what is broken?

bb511: The accuracy summary table.

bb511: The accuracy summary table.

AnthonyLim23: what is it doing and what should it be doing?

bb511: It should output a table with the most accurate minimser etc., instead it outputs gibberish.

AnthonyLim23:
@Anders-Markvardsen  and @bb511 I just ran this and found that the `comparison_weighted_v3.8_acc_summary.html` (and the runtime summary) did not work. Do we have an issue for this?

bb511:
Just checked, don't think we do. I will create one today.

AnthonyLim23:
@bb511 is this the bit of code you commented? 

Other than that just move the function and I am happy with it

bb511:
Yes, that would be the bit I commented.

PR #27, opened 2018-07-11
Fixed issue with master when parsing neutron data

bb511:
Fixed https://github.com/mantidproject/fitbenchmarking/blob/ee3307bd8f3e9210074b55765b44f45dab591f42/fitbenchmarking/input_parsing.py#L249

This line was causing an error: `global name 'k' is not defined`
Change k to sep_idx


AnthonyLim23:
just tested it and it seemed to run fine, I did get an error about the summary tables but we are aware of that bug.

PR #30, opened 2018-07-12
Best_fit not defined solved

### LINKED ISSUE #29: do_fitting_benchmark_one_problem broken in some cases
In fitting_benchmarking.py, do_fitting_benchmark_one_problem function, parameter best_fit is not defined  before allocation in some cases. More specifically, when the fit fails https://github.com/mantidproject/fitbenchmarking/blob/e3190c063de98703a83d224ada23b0de63533dd9/fitbenchmarking/fitting_benchmarking.py#L181
will not execute. However, the line https://github.com/mantidproject/fitbenchmarking/blob/e3190c063de98703a83d224ada23b0de63533dd9/fitbenchmarking/fitting_benchmarking.py#L204
does, leading to an error message. It is interesting that the current algorithm runs without being impeded by this error.

bb511:
**Description of work:**
Fixed #29, i.e. best_fit not being defined before assignment sometimes, and moved the plotting algorithm that was initially embedded into do_fitting_benchmark_one_problem function into its own separate function.

**To test:**
1. Open the mantidpython console
2. Run example_runScript.py in the console by typing
`run /your_path/fitbenchmarking/fitbenchmarking/example_runScript.py`
3. Check if you are getting the same results as master

### DIFF
--- a/fitbenchmarking/fitting_benchmarking.py
+++ b/fitbenchmarking/fitting_benchmarking.py
@@ -135,7 +135,6 @@ def do_fitting_benchmark_group(group_name, problem_files, minimizers, use_errors
     return problems, results_per_problem
 
 
-
 def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0, previous_name="none"):
     """
     One problem with potentially several starting points, returns a list (start points) of
@@ -170,6 +169,7 @@ def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0,
 
             print("*** with minimizer {0}, Status: {1}, chi2: {2}".format(minimizer_name, status, chi2))
 
+            best_fit = None
             sum_err_sq = -1
             if not status == 'failed':
                 print("   params: {0}, errors: {1}".format(params, errors))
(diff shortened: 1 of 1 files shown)


fitbenchmarking/fitting_benchmarking.py
```diff
@@ -170,6 +169,7 @@ def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0,
 
             print("*** with minimizer {0}, Status: {1}, chi2: {2}".format(minimizer_name, status, chi2))
 
+            best_fit = None
```
Anders-Markvardsen: To further improve code readability. Could you add a comment to explain when new variables are first introduced to the code (expect where it is very clear what a variable). Please also don't hesitate to add such where they are currently missing

bb511: Yes, I was going to create another branch where I would add comments and do a general clean up of the code as soon as I finished with the unit tests.

bb511: Yes, I was going to create another branch where I add comments and do a general clean up of the code as soon as I finished with the unit tests.

Anders-Markvardsen: OK

PR #40, opened 2018-07-23
Code now uses poisson distributed y data to fake observational errors

### LINKED ISSUE #31: Fake observational errors are erroneous
The way the observational errors are obtained for data sets that do not have them already currently use X data to calculate the said errors. However, they should use Y data.

https://github.com/mantidproject/fitbenchmarking/blob/9a64a061f5b53b570193916d6d0ce7cac4cee6d2/fitbenchmarking/fitting_benchmarking.py#L345

bb511:
Additionally, changed the order of the minimizers to increase readability

Fixes #31 

#### To test:

1. Make sure that in example_runScripts.py `run_data` is set to "nist".
2. Run example_runScripts.py in the mantidpython console.
3. Compare the results with the output you get from following the same steps in master branch.


Anders-Markvardsen:
thanks for spotting this bug

PR #41, opened 2018-07-23
Made IF statement less convoluted.

### LINKED ISSUE #36: IF statement in prepare_wks_cost_function
The if statement https://github.com/mantidproject/fitbenchmarking/blob/9a64a061f5b53b570193916d6d0ce7cac4cee6d2/fitbenchmarking/fitting_benchmarking.py#L343

represents a weird and potentially dangerous way of testing if observational errors exist for the current problem. Should replace this with a simpler test that also makes the code more readable.

bb511:
Fixes #36 

#### To test:

1. Make sure that in example_runScripts.py `run_data` is set to "nist".
2. Run example_runScripts.py in the mantidpython console.
3. Compare the results with the output you get from following the same steps in master branch.

### DIFF
--- a/fitbenchmarking/fitting_benchmarking.py
+++ b/fitbenchmarking/fitting_benchmarking.py
@@ -340,7 +340,7 @@ def prepare_wks_cost_function(prob, use_errors):
     """
     if use_errors:
         data_e = None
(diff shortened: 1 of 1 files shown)


Anders-Markvardsen:
I agree this code maintenance change is sensible

PR #46, opened 2018-07-23
Results output tests

bb511:
#### Description of Work:
Added unit tests for functions that display the results of the fits.

#### To test:
Run the test files in the mantidpython console and check if the tests are successful (if tests with fail in their name do fail, that is considered expected behaviour).


Anders-Markvardsen:
nice first set of tests. all passed when tested on window

PR #47, opened 2018-07-23
Deleted the redundant chi_squared calculations 

bb511:
#### Description of Work
Deleted the redundant chi_squared calculations along with fixing some minor issues. Restructured the test_result object such that it only records one chi squared value. Added a function that calculates chi squared separately by using a vector of the differences between the actual data and the fit data (better when the project is extended beyond mantid).

#### Testing Instructions

1. In `example_runScripts.py` make sure that run_data is `"neutron"`.
2. Run the script in the mantidpython console.
3. Compare the results with the ones from master, they should be the same.
4. Change run_data to `"nist"`.
5. Repeat the process.

### DIFF
--- a/fitbenchmarking/fitting_benchmarking.py
+++ b/fitbenchmarking/fitting_benchmarking.py
@@ -158,7 +158,7 @@ def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0,
     # Loop over the different starting points
     for user_func in function_defs:
         # search for lowest chi2
-        min_chi_sq = 1.e100  # used to store min chi-sq value found. initialised to large value
+        min_chi_sq = 1.e20  # used to store min chi-sq value found. initialised to large value
         best_fit = None
         results_problem_start = []
         for minimizer_name in minimizers:
--- a/fitbenchmarking/fitting_benchmarking.py
+++ b/fitbenchmarking/fitting_benchmarking.py
@@ -158,7 +158,7 @@ def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0,
     # Loop over the different starting points
     for user_func in function_defs:
         # search for lowest chi2
-        min_chi_sq = 1.e20
+        min_chi_sq = 1.e100  # used to store min chi-sq value found. initialised to large value
         best_fit = None
         results_problem_start = []
         for minimizer_name in minimizers:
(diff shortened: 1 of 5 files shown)


fitbenchmarking/fitting_benchmarking.py
```diff
@@ -154,45 +154,45 @@ def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0,
 
     # Get function definitions for the problem - one for each starting point
     function_defs = get_function_definitions(prob)
-    # search for lowest chi2
-    min_sum_err_sq = 1.e20
+
     # Loop over the different starting points
     for user_func in function_defs:
```
Anders-Markvardsen: Looking at this code in more detail. Can you confirm this is a loop over different starting points. Asking before this is not clear from 'for user_func in function_defs'

fitbenchmarking/fitting_benchmarking.py
```diff
@@ -154,45 +154,45 @@ def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0,
...
+        # search for lowest chi2
+        min_chi_sq = 1.e100  # used to store min chi-sq value found. initialised to large value
+        best_fit = None
```
Anders-Markvardsen: what does best_fit store?

fitbenchmarking/fitting_benchmarking.py
```diff
@@ -154,45 +154,45 @@ def do_fitting_benchmark_one_problem(prob, minimizers, use_errors=True, count=0,
 
     # Get function definitions for the problem - one for each starting point
     function_defs = get_function_definitions(prob)
-    # search for lowest chi2
-    min_sum_err_sq = 1.e20
+
     # Loop over the different starting points
     for user_func in function_defs:
+        # search for lowest chi2
+        min_chi_sq = 1.e100  # used to store min chi-sq value found. initialised to large value
```
Anders-Markvardsen: instead of 1.e100 it would be even better to set to max floating point value

fitbenchmarking/fitting_benchmarking.py
bb511: Yes, this loop over different starting points. Tested it in one of the FittingBenchmarkOneProblem tests.

fitbenchmarking/fitting_benchmarking.py
bb511: For a function_def, best_fit stores the y and x data of the best fit after testing against all the minimisers provided by the user.

fitbenchmarking/fitting_benchmarking.py
bb511: Yes, I think I changed it by mistake. The maximum value should be 1e20.

bb511: Hmm, this change was not present on my local branch.

bb511: Fixed in commit 80decd4

Anders-Markvardsen:
As discussed

PR #51, opened 2018-07-24
Fake observational errors now can deal with negative values

### LINKED ISSUE #50: Fake observational errors problem when encountering negative values
The code in question
https://github.com/mantidproject/fitbenchmarking/blob/517f790c98131e33f9bc0bfebd396349e13c4b6a/fitbenchmarking/fitting_benchmarking.py#L342-L344
presents errors when y is a negative number (i.e. in the case of Bennett5.dat nist problem file).

bb511:
Fixes #50

#### Description of Work
Faking observational errors now takes the square root of the absolute value of y data.

#### Testing Instructions
1. In `example_runScript.py` make sure that `run_data = "nist"`.
2. In the mantidpython console, run `example_runScript.py`.
3. Check the acc_high_difficulty nist table that is produced, the Bennett5 rows should now contain numbers instead of `nans`.

### DIFF
--- a/fitbenchmarking/fitting_benchmarking.py
+++ b/fitbenchmarking/fitting_benchmarking.py
@@ -340,8 +340,10 @@ def prepare_wks_cost_function(prob, use_errors):
     if use_errors:
         data_e = None
         if prob.data_pattern_obs_errors is None:
-            # Fake observational errors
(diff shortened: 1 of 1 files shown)


Anders-Markvardsen:
As discussed

PR #65, opened 2018-07-26
Run all tests at once added

### LINKED ISSUE #55: Create a way to run all the tests at once
Instead of running the neutron data, nist data etc. individually, create a way to run them all at once.

bb511:
#### Description of Work

Fixes #55
Changed `example_runScript` such that it now runs all tests at once.

#### Testing Instructions

1. Run `example_runScript.py` in the mantidpython console.
2. Check if all the output is in the results folder.

**Note:** This was giving an error after completing execution: `RuntimeError maximum recursion depth reached` so I increased the recursion depth here:
https://github.com/mantidproject/fitbenchmarking/blob/acf175d8f33b932b73a653eef98ce6d0d6cb341a/fitbenchmarking/example_runScripts.py#L5-L6
I have not found any other way that fixes the problem.

### DIFF
--- a/fitbenchmarking/example_runScripts.py
+++ b/fitbenchmarking/example_runScripts.py
@@ -60,6 +60,10 @@ for run_data in ["neutron", "nist"]:
                                                       minimizers=minimizers, use_errors=use_errors,
                                                       results_dir=results_dir)
 
+    else:
+        raise RuntimeError("Invalid run_data, please check if the array"
+                            "contains the correct names!")
+
     for idx, group_results in enumerate(results_per_group):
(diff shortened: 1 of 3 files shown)


fitbenchmarking/example_runScripts.py
```diff
@@ -2,6 +2,8 @@
 # fit problems
 
 import os
+import sys
+sys.setrecursionlimit(10000)
```
Anders-Markvardsen: Please add a comment to why this is needed - i.e. a version of what you explain in the issue

Anders-Markvardsen:
and re-tested it after the most recent commit

Issue #48, opened 2018-07-24
Code Cleanup - Add Comments and Improve Variable Names


bb511:
In this instance of cleanup one needs to add comments where code gets ambiguous. Additionally, make the variable names more intuitive. The aim of this is to make the code easier to understand overall.

PR #70, opened 2018-07-26
Improved how the script handles results_dir input

### LINKED ISSUE #66: Improve how setting a result directory handles edge cases
Improve how the functionality added in #59 handles non-sense names and not giving the full path.

bb511:
#### Description of Work

Fixes #66 
Code can now handle results_dir input better.

#### Testing Instructions

1. Run `example_runScript.py` with different `results_dir`. These should include:
    - results_dir = None
    - results_dir = "Something"
    - results_dir = 12
    - results_dir = os.sep + "Something"
2. Check if the output is the same as master.

### DIFF
--- a/fitbenchmarking/fitting_benchmarking.py
+++ b/fitbenchmarking/fitting_benchmarking.py
@@ -508,10 +508,14 @@ def get_data_group_problem_files(grp_dir):
     return probs
 
 
-def empty_contents_of_folder(results_dir):
+def empty_contents_of_folder(directory):
+    ''' Deletes everything in the directory given by directory.
 
-    for file in os.listdir(results_dir):
-        file_path = os.path.join(results_dir, file)
+    @param directory :: path to the directory that gets wiped
+    '''
+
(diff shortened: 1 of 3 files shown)


fitbenchmarking/fitting_benchmarking.py
```diff
@@ -516,3 +517,22 @@ def empty_contents_of_folder(results_dir):
             os.unlink(file_path)
         elif os.path.isdir(file_path):
             shutil.rmtree(file_path)
+
+
+def setup_results_directory(results_dir):
```
Anders-Markvardsen: please add documentation for this method

bb511: I am often skipping documentation as I will add it all in #48 , but can add it now if that's preferred.

Anders-Markvardsen: Yes please add as you add new functions since this is point in time where it is most fresh in your mind what the method is doing

PR #81, opened 2018-08-03
Inverted x and y

### LINKED ISSUE #80: NIST data x and y are inverted
When parsing nist data, x data is read as y and vice versa.

bb511:
#### Description of Work

Fixes #80 
Fixes issue in which the nist data was inverted (x was y and y was x).

#### Testing Instructions

1. Run `example_runScripts.py` in the mantidpython terminal.
2. Check the nist tables and the visual display pages of the nist problems.


Anders-Markvardsen:
thanks for self-correcting yourself:)

Issue #92, opened 2018-08-09
EVS14188-90_processed Gaussian peaks 1 visual display page is broken


bb511:
As the title suggests, this visual display page is broken. I suspect this is because of how function definitions are treated in the neutron case as opposed tot the nist case. The two cases are treated very separately at the moment. Solve this issue and created a more integrated approach (since I think this will help in solving the bug).

Issue #88, opened 2018-08-06
Add function descriptions


bb511:
Add @params and @returns to all the functions in the tool. Also add good descriptions of what the functions do.

PR #93, opened 2018-08-09
Added new directory structure

### LINKED ISSUE #77: Code Cleanup - Organize functions into classes
As the title says, organize similar functions into classes as this will make the program easier to understand.

bb511:
#### Description of Work

Fixes #77 
Added a new directory struncture in which functions are organized into modules and subdirectories. Prepared the way to extending the tool to use other minimizers outside mantid.

#### Testing Instructions

1. In the `mantidpython` terminal, run `/example_scripts/example_runScripts.py` .
2. Check if the output tables are as expected (i.e. produce the same results as master).

**Notes:**

* EVS14188-90_processed Gaussian peaks 1 visual display page does not work. This is a known issue, check #92 
* The functions lack descriptions in the current state. They will be added in #88 , as discussed.


fitbenchmarking/fitbenchmark_one_problem.py
```diff
@@ -0,0 +1,119 @@
...
+
+from fitting import fit_algorithms, mantid_utils
+from fitting.plotting import plots
+from utils import fit_misc, test_result
+from utils.logging_setup import logger
+
+MAX_FLOAT = sys.float_info.max
+
+
+def fitbm_one_problem(prob, minimizers, use_errors=True, group_results_dir=None):
```
Anders-Markvardsen: Please explain what this method is doing. E.g. it is not immediately obvious how it differ from fitbm_one_problem_def

fitbenchmarking/fitting/minimizers.json
```diff
@@ -0,0 +1,8 @@
+{
```
Anders-Markvardsen: Hard coding minimizer names in files within the fitbenchmarking package, unless there is a very good reason not to, should be avoided. This minimizer list may change with future versions of Mantid and hence a future user of fitbenchmarking would potentially have to change the guts of fitbenchmarking as I read this json file. Having such a file outside the fitbenchmarking package would be fine, e.g. in example_scripts

Anders-Markvardsen:
Tested and results as expected.

Re-packaging / re-structuring probably in the right direction - speak later about this

bb511:
1. Every function gets a description in the next PR
2. Empty lines removed
3. Markdown file only requires little changes, I was planning to write some proper documentation though.

fitbenchmarking/fitting/minimizers.json
bb511: I am trying to find a way such that the user can get a list of available minimizer names compatible with a certain algorithm and choose which minimizers they want/all minimizers. The implementation is not yet complete but the json file is a start.

PR #95, opened 2018-08-10
Added function descriptions

### LINKED ISSUE #88: Add function descriptions

bb511:
#### Description of Work

Fixes #88 
Added function descriptions.

#### Testing Instructions

1. Look through all the files and see if the function descriptions are satisfying.

### DIFF
--- a/fitbenchmarking/fitbenchmark_one_problem.py
+++ b/fitbenchmarking/fitbenchmark_one_problem.py
@@ -40,8 +40,10 @@ MAX_FLOAT = sys.float_info.max
 
 def fitbm_one_problem(prob, minimizers, use_errors=True, group_results_dir=None):
     """
-    Call all the functions and perform a fit benchmarking for
-    only one problem.
+    Sets up the workspace, cost function and function definitons for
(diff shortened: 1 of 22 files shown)


fitbenchmarking/fitbenchmark_one_problem.py
```diff
@@ -40,6 +40,19 @@
 
 def fitbm_one_problem(prob, minimizers, use_errors=True, group_results_dir=None):
     """
+    Sets up the workspace, cost function and function definitons for
+    a particular problem and fits the models provided in the problem
+    object. The best fit, along with the data and a starting guess
+    is then plotted on a visual display page.
```
Anders-Markvardsen: Does this function actually do this "The best fit, along with the data and a starting guess
    is then plotted on a visual display page."

bb511: Yes, it calls make_plots which then plots the results.

Anders-Markvardsen:
thanks for the adding of this additional code document

PR #101, opened 2018-08-14
Created second table on the visual display page

### LINKED ISSUE #90: Create a second table with final parameter values
Have a second table on the visual display pages, after the solution found subtitle, where the final values of the parameters are listed.

bb511:
#### Description of Work

Fixes #90 
Added a second table on the visual display page that lists the functional form and the final parameters of the best fit.

#### Testing Instructions

1. In the `mantidpython` terminal run `example_runScripts.py`.
2. Check if the results are like expected.


AnthonyLim23:
looking at the neutron table, I only see 1 table.

AnthonyLim23:
sorry, I was unclear. I open the neutron table. Then click one of the runs, I see 3 plots and one table

bb511:
Looking into that.

PR #109, opened 2018-08-20
Documented the WISH problem adhoc exception

### LINKED ISSUE #38: WISH problems ad hoc exception
The WISH problems are considered with an ad hoc exception at
https://github.com/mantidproject/fitbenchmarking/blob/9a64a061f5b53b570193916d6d0ce7cac4cee6d2/fitbenchmarking/fitting_benchmarking.py#L300-L305
As the comments say, this should be investigated and changed.

bb511:
#### Description of Work

Fixes #38 
Documented the WISH problem ad hoc exception. The problem itself is much deeper and it does not concern the fitting benchmarking tool. The issue has to do with the WISH nexus files.

#### Testing Instructions

Look through the code and check if the comments are good enough.

### DIFF
--- a/fitbenchmarking/fitting/mantid_utils.py
+++ b/fitbenchmarking/fitting/mantid_utils.py
@@ -130,6 +130,9 @@ def ignore_invalid(prob, cost_function):
     """
 
     ignore_invalid = cost_function == 'Least squares'
+
+    # The WISH data presents some issues
+    # For which this adhoc if must is present
     if 'WISH17701' in prob.name:
         ignore_invalid = False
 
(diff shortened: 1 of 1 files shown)


fitbenchmarking/fitting/mantid_utils.py
```diff
@@ -130,6 +130,9 @@ def ignore_invalid(prob, cost_function):
     """
 
     ignore_invalid = cost_function == 'Least squares'
+
+    # The WISH data presents some issues
+    # For which this adhoc if must is present
```
AnthonyLim23: is it possible to have a little more explanation as to why this error occurs and why the if statement works?

bb511: Unfortunately, I also have no idea what the specific cause is. I tired looking into it and got nowhere pretty quickly as there seems to be some special thing about the WISH nexus files.

AnthonyLim23: ok, can you change the is to a to a be?

PR #116, opened 2018-08-29
All code depending on mantid now contained in mantid.py file

### LINKED ISSUE #5: Extend fit benchmarking script to be callable with any minimizers
_From @Anders-Markvardsen on December 5, 2016 12:59_

Extend the fit benchmarking scripts introduced with #17654 so that the benchmarking scripts can be called with non-Mantid minimizers also.

There are several motivations for this:

* To make it easy to test if e.g. any of the minimizers in scipy or Matlab are better than the once we currently offer to our Mantid users 
* To enable someone to test new idea for a minimizer within needing to first learn how to add a minimizer in Mantid (Although IMPORTANTLY note testing against an entirely different code is not 100% the same as comparing minimizers within the mantid framework - although this - should! - not be that different - if there are differences this would be something Mantid could learn from also)
* Motivate more to use our benchmarking framework and extending it with more fit problems

_Copied from original issue: mantidproject/mantid#18172_

bb511:
#### Description of Work

Fixes #5 
Laid the base for extending to other fitting algorithms/minimizers.

#### Testing Instructions

1. In the `mantidpython` terminal, run `example_runScripts.py`.
2. Check if the results are the same as before


Anders-Markvardsen:
As discussed

PR #159, opened 2019-04-08
Restructuring workflow

### LINKED ISSUE #154: Investigation into redesign of FittingProblem class
Investigation into redesign of FittingProblem and surrounding helper functions. 

Create diagrams docs and discuss these

* look into making helping functions member functions
* look into making FittingProblem a composite class

### LINKED ISSUE #155: Investigate larger restructuring of workflow of FitBenchmarking
@wathen could you add a copy of your proposed restructuring please?

wathen:
#### Description of Work

I have moved utils/fitbm_problem.py to parsing and updated the base problem class. From the base problem class I changed parsing/parse_{nist,fitbenchmark}_data.py to be child classes of the base class and added the data type specific functions to these classes.  Under testing instructions I have added new documentation to add new data types  

Note: start_x, end_x and data_e are not initialized by `parse_nist_data.FittingProblem(prob_file)`, this is consistent with how it was before

Fixes #154
Fixes #155 

#### Testing Instructions

1. New documentation for NIST and FitBenchmark fitting classes
2. Assess readability of code
3. New documentation about adding new data types

Documentation:

# Supported fitting problem definition types

## Fitting problem definition types currently supported
At the time of writing, the types (formats) that are currently support are:
* Native (Fitbenchmark)
* NIST

An example of the native and NIST formats can be seen in `benchmark_problems/Neutron_data/` and `benchmark_problems/NIST/`, respectively.

## Adding new fitting problem definition types

Follow the following steps
1. Create `parse_{type}_data.py` which contains a child class of `BaseFittingProblem` in `parsing/base_fitting_problem.py` that processes the type (format) and initialise the class with appropriate attributes (examples can be found in `parse_{nist/fitbenchmark}_data.py`)
2. In `parsing/parse.py` alter the function `determine_problem_type()` such that it determines the new type
3. In `parsing/parse.py` add a new if statement to `parse_problem_file()`  to call the user defined `parse_{type}_data.py` class


fitbenchmarking/fitting/mantid/tests/test_prepare_data.py
```diff
@@ -17,11 +17,27 @@
 from fitting.mantid.prepare_data import setup_errors
 from fitting.mantid.prepare_data import convert_back
 
-from utils import fitbm_problem
+from parsing.fitbm_problem import BaseFittingProblem
 
 
 class MantidTests(unittest.TestCase):
 
+    def misra1a_file(self):
```
Anders-Markvardsen: Could misrela_file, neutron_peak_19_file and similar be factored out into separate python file. If only used within mantid/tests I suggest put these in python file named data_for_testing.py

Anders-Markvardsen:
Tested example scripts works on windows

fitbenchmarking/parsing/fitbm_problem.py
```diff
@@ -0,0 +1,150 @@
...
+        - floats: start_x, end_x, ref_residual_sum_sq
+        - numpy arrays: data_x, data_y, data_e
+        - arrays: starting_values
+    """
+    __metaclass__ = ABCMeta
+
+    def __init__(self, fname):
+
+        # Initializes base class with file and problem type
+        self.fname = fname
```
Anders-Markvardsen: why is fname attribute without _ when others are not, not clear from just quickly looking at the code

fitbenchmarking/parsing/fitbm_problem.py
```diff
@@ -0,0 +1,150 @@
...
+        self._type = None
```
Anders-Markvardsen: Can we get rid of self._type with now base and child classes? Since the type of child class should tell which type it is

fitbenchmarking/parsing/fitbm_problem.py
```diff
@@ -0,0 +1,150 @@
...
+        self._data_x = None
+        self._data_y = None
+        self._data_e = None
+
+        # The 'certified' or reference sum of squares, if provided
+        # (for example in NIST tests).
+        self.ref_residual_sum_sq = None
+
+    def read_file(self):
+        self.contents = open(self.fname, "r")
```
Anders-Markvardsen: This attribute is created on the fly rather than up front. I suggest either needs to be declared up front like the other attributes or avoided (since in effect it will then store the same information twice in the class)

Anders-Markvardsen:
Check start_x, end_x and data_e for NIST problem set, private attributes, rename base_fitting_problem.py then correct tests

Anders-Markvardsen:
https://github.com/mantidproject/mantid/blob/master/scripts/test/Muon/utilities/muon_pair_test.py#L29 for assert

Anders-Markvardsen:
I like your updated description in https://github.com/mantidproject/fitbenchmarking/wiki/Problem-Groups. 

I also like you section for adding a new type. I have made changed 'data type' to 'fitting problem definition type' reason being is that scientist will associate 'data type' with the raw data file type (i.e. if nexus data file or xye data file or other), hence to avoid this confusing the somewhat longer 'fitting problem definition type' I think is better - but please don't hesitate to disagree or suggest otherwise

Anders-Markvardsen:
@wathen also please update wiki with your new section above on types, thanks

PR #164, opened 2019-04-25
User defined minimizer file

### LINKED ISSUE #153: minimise files
I think it would be good if we had a 3 possible ways to define the minimizers we are interested in. 

1. At the command line e.g. `.\fitbenchmarking ../../minimizers/BUMPS.json`

2. In the run script itself

3. The default fall back

wathen:
#### Description of Work

Enables users to customize define separate json file which contains minimizers using command line arguments. Also, added option of defining minimizer dict within example script.

Fixes #153 

#### Testing Instructions

1. Check tests work
2. Documentation changed in `fitbenchmarking/example_scripts`


fitbenchmarking/utils/misc.py
```diff
@@ -31,21 +31,28 @@
 from utils import user_input
 
 
-def get_minimizers(software):
+def get_minimizers(software, minimizers_list, json_file):
```
AnthonyLim23: as far as I can tell you always use this function after extracting the arguments from `software_options`. It would be neater to just pass the `software_options` and extract it here

fitbenchmarking/utils/misc.py
```diff
@@ -31,21 +31,28 @@
...
     """
     Gets an array of minimizers to fitbenchmark from the json file depending
     on which software is used.
     @param software :: string defining the software used
+    @param miimizers_list :: dictionary containing minimizers
```
AnthonyLim23: typo

example_scripts/README.md
```diff
@@ -1,3 +1,13 @@
 # Example Scripts
 
 Folder that holds all the example scripts for fitBenchmarking. The user can modify these files and run a benchmarking on their machine. For more details, please read the comments in the code of each file.
+
```
AnthonyLim23: can we make these a bit easier for a user to read?

AnthonyLim23:
I have had a look at the code and it has a couple of minor things. Would someone else have time to test it?

example_scripts/example_runScripts.py
```diff
@@ -43,9 +44,25 @@
 from fitting_benchmarking import do_fitting_benchmark as fitBenchmarking
 from results_output import save_results_tables as printTables
 
-
 # SPECIFY THE SOFTWARE/PACKAGE CONTAINING THE MINIMIZERS YOU WANT TO BENCHMARK
-software = 'mantid'
+software = 'scipy'
```
Anders-Markvardsen: @wathen thanks for this first go. Thinking about it more I would suggested changes which we can discuss at 3pm today:

* remove from software_options 'minimizer_list' and rename 'json_file' to 'minimizer_list_file'
* rename minimzers.json to minimers_list_default.json
* To discussed today: should we insist on example_scripts to load minimizers_list info from file?

Alternative is to rename minimizers.json to fitbenchmarking_config.json, where the later becomes a placeholder for all configurations which can be overwritten

Anders-Markvardsen:
@wathen noticed you attempted to change travis to point to Python 3. Because of Mantid we have so far been restricted to use Python 2.7 (See Fitbenchmarking README). @AnthonyLim23 out of curiosity what is the timeline for Mantid moving to Python 3?

AnthonyLim23:
Mantid cannot move to Python3 until redhat moves over, which wont be for a while yet.

example_scripts/example_runScripts.py
```diff
@@ -43,9 +44,25 @@
...
+software_options = {'software': software}
+
+if len(sys.argv) > 1:
+    software_options['minimizer_file'] = current_path + sys.argv[1]
```
Anders-Markvardsen: Looks a lot better overall, but still have an issue with software_options have three attributes. If we keep this, these three options needs to be transparent in all methods they are used and documented

PR #165, opened 2019-05-15
Enabling different comparison between minimizers in different software packages

### LINKED ISSUE #132: Allow users to comparing minimizers from scipy, mantid, other in one table
Allow user to run a fit benchmarking with a mixture of different sources of minimizers (e.g. mantid and scipy) and where the relative scores of these against selected benchmarking problems is shown in a table. 

At present FitBenchmarking only create tables which shows comparisons in scores between minimizers available from one software, and separate tables are produced for each software

wathen:
#### Description of Work

Adding new feature to enable the comparison of minimizers from different software packages.

Currently expert example script compares the two software packages. Working on incorporating this into both example scripts.

#### Testing Instructions

1. Code documentation
2. Ease of use

Fixed #132


fitbenchmarking/fitting_benchmarking.py
```diff
@@ -61,13 +61,43 @@ def do_fitting_benchmark(group_name, software_options, data_dir,
...
 
   return prob_results, results_dir
 
 
+def do_benchmarking(user_input, problem_groups, group_name):
+  """
+  Loops through software and benchmarks each problem within the problem
+  group.
+
+  @param user_input :: all the information specified by the user
```
Anders-Markvardsen: @wathen your comment for user_input is correct. However, separately to this PR we could discuss how user_input may benefit from being broken down, i.e. to reduce the need for the code reader to have to look elsewhere in the code to check what user_input might contain

Issue #163, opened 2019-04-12
NIST data class do not initialize with start_x, end_x and data_e 


wathen:
Observed in #159 that start_x, end_x and data_e are not initialized for the NIST data set.

wathen:
More the default values for start_x, end_x and data_e out of fitbenchmarking/fitting and set defaults in fitbenchmarking/parsing/parse_{fitbenchmark,nist}_data.py automatically

PR #170, opened 2019-06-19
Removing fit_parameter bug in EVS14188* data

wathen:
#### Description of Work

Adds condition to check if fit_parameters is present in native fitbenchmark data format.

#### Testing Instructions

1. Check to see if example scripts run

### DIFF
--- a/fitbenchmarking/parsing/parse_fitbenchmark_data.py
+++ b/fitbenchmarking/parsing/parse_fitbenchmark_data.py
@@ -61,8 +61,9 @@ class FittingProblem(base_fitting_problem.BaseFittingProblem):
         self._type = "FitBenchmark"
 
         self._starting_values = None
-        self._start_x = entries['fit_parameters']['StartX']
-        self._end_x = entries['fit_parameters']['EndX']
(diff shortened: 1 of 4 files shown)


fitbenchmarking/parsing/parse_fitbenchmark_data.py
```diff
@@ -61,8 +61,9 @@ def __init__(self, fname):
         self._type = "FitBenchmark"
 
         self._starting_values = None
-        self._start_x = entries['fit_parameters']['StartX']
-        self._end_x = entries['fit_parameters']['EndX']
+        if 'fit_parameters' in entries:
```
Anders-Markvardsen: Thanks for spotting this and I agree it is reasonable to make StartX and EndX optional

Anders-Markvardsen:
@wathen would you say this PR also fixes #163 ?

wathen:
> @wathen would you say this PR also fixes #163 ?

Nope, that would need to be coded into the two classes seperately

Anders-Markvardsen:
@wathen could you add a sentense to #163 of how you propose this issue fixed for future reference please

PR #176, opened 2019-07-03
Remove type attribute

### LINKED ISSUE #162: remove type attribute in fittingproblem classes
In BaseFittingProblem currently have attribute self._type = None. 

In this issue remove this type and instead figure out the type from the child name. The issue is that currently the type is in effect represented twice, which is not needed.

AtomAnu:
#### Description of Work

Fixes #162 

The type name is now obtained by the child name (file name).

#### Testing Instructions

1. Run `example_runScripts.py` and investigate the correctness of the output


.idea/workspace.xml
```diff
@@ -0,0 +1,316 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ChangeListManager">
```
Anders-Markvardsen: Please remove this file

Anders-Markvardsen:
Hi @AtomAnu, could you alternatively look into using https://stackoverflow.com/questions/11408148/how-to-get-derived-class-name-from-base-class and see if you can use this to remove _type attribute entirely from the base class

fitbenchmarking/parsing/parse_fitbenchmark_data.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, fname):
 
         self._name = entries['name']
         self._equation = entries['function']
-        self._type = "FitBenchmark"
+        self._type = ((os.path.basename(__file__)).split('_')[1]).upper()
```
Anders-Markvardsen: I think you can remove self._type from here and in equivalent file for NIST by using suggestion here for getting module name in https://stackoverflow.com/questions/2020014/get-fully-qualified-class-name-of-an-object-in-python in func_def.py

AtomAnu: I will look into that.

AtomAnu:
@Anders-Markvardsen 
I have finished solving this issue.

Anders-Markvardsen:
Could you check the build - looks like some tests are not passing

PR #183, opened 2019-07-11
Initialisation of start_x, end_x and data_e for NIST problems

### LINKED ISSUE #163: NIST data class do not initialize with start_x, end_x and data_e

AtomAnu:
#### Description of Work

Fixes #163 

#### Testing Instructions

1. Run `example_runScripts.py` with NIST problems selected
2. Run `parsing/test_parse_nist_data.py`
3. Add a new NIST problem definition fiile with data errors (data_e) included in the same row as x and y data.


fitbenchmarking/parsing/parse_nist_data.py
```diff
@@ -303,3 +302,31 @@ def get_startvals_floats(self, startval_str):
...
+
+        start_x = sorted_x_data[0]
+        end_x = sorted_x_data[-1]
+
+        return start_x, end_x
+
+    def get_data_e(self, data_pattern):
+        """
+
+        :param data_pattern: numpy array of the data points of the problem
```
Anders-Markvardsen: did you mean to write "@param" instead of ":param"?

AtomAnu: Yes I did mean `@param`

AtomAnu: @Anders-Markvardsen I have now fixed the comment.

Anders-Markvardsen:
could you merge conflict please?

AtomAnu:
@Anders-Markvardsen I will do that now.

AtomAnu:
@Anders-Markvardsen It's now ready to be merged with master.

wathen:
Sorry I have been a bit out of the changes here but adding these new functions within fitbenchmarking will change the results for the NIST/CUTEst data sets? If you look at fitbenchmarking/fitting/prerequisites.py in the last function the default values are given as:

>     if problem.start_x == None and problem.end_x == None:
>         problem.start_x = - np.inf
>         problem.end_x = np.inf
>     return data, cost_function, function_definitions

This seems fairly important that the results between this branch and the master would presumiably be different?

AtomAnu:
@wathen I have just verified the results from this branch with that of the master branch. The results appear to be identical to each other. 

Therefore, would it be fine to add the changes in this branch to master?

wathen:
@AtomAnu Did you check the raw number and not the normalised ones? I can see that the normalised output which is currently what fitbenchmarking produces could be the same but not when you’re using completely different starting points to what the default was before.

@Anders-Markvardsen and @tyronerees  do you think this is an issue?

tyronerees:
If I understand correctly, this PR adds default starting and ending values for the range of X we're interested in, as required by #183.   By default we should fit all the data, as there's no prior knowledge to exclude any of the X's, and this is what this PR implements. The issue is whether changing the defaults from +/- Inf to min/max(X) should change the output.  Please shout someone if I've got this wrong.

As far as I can see, there should be no difference in the SciPy case, as the StartX/EndX values are just used to set constraints, and +/- Inf will capture everything (see [here](https://github.com/fitbenchmarking/fitbenchmarking/blob/17c223c4e6186234aa326c8fc7bfdd34b491a66e/fitbenchmarking/fitting/scipy/prepare_data.py#L66)) .

The question then is what Mantid does under the hood when you send it StartX and EndX values ([here](https://github.com/fitbenchmarking/fitbenchmarking/blob/17c223c4e6186234aa326c8fc7bfdd34b491a66e/fitbenchmarking/fitting/mantid/main.py#L93))?  My suspicion is that it does the same as SciPy, but I can't find where this is in the mantid code at the moment -- @Anders-Markvardsen / @AnthonyLim23, do you know off the tops of your heads?

Anders-Markvardsen:
If start_x and end_x are not specified then Mantid simply just pick the start and end of the x-range, and the same applied if start_x = - infinity and end_x = + infinity, since it will only attempt to fit against x-values for which we have provided y-values.

Hence @wathen for the above reason I would be surprised if the changes in the PR would change any of the FitBenchmarking outputs - does this seem right?

Just speaking to @AtomAnu . He will run some extra check for different combinations of Mantid/Scipy against Neutron/NIST date. If all these relative number are identical I will would expect the same applied for there absolute counterparts

wathen:
@Anders-Markvardsen I think it might be worth checking the actual values just to make sure? It’s easy to check and then we can be sure both run to the same minimiser

But anyway, wouldn’t it make more sense to set start_x and end_x in the base class so when more examples are added you can remove the references to default values later in the code?

AtomAnu:
@wathen I will check the actual values. However, I think it would be better to set start_x and end_x  in the classes (FittingProblem) that derive from the base class (BaseFittingProblem) because the way to set those values is different between NIST and FitBenchmark problems.

PR #193, opened 2019-07-31
Function Evaluation Method 

AtomAnu:
#### Description of Work

Fixes:

1. In every problem object created from either `parse_nist_data.py` or `parse_fitbenchmark_data.py`, there are now two additional modules. 
    1. The first module `eval_f(x, param_list)` is a function evaluation method which takes in x       value(s) and the function parameters, and returns the y value(s) accordingly.
    2. The second module 'get_function()` returns a list of the used function in the fitting process  and the initial parameters

2. For fitting using mantid, the method `mantid.simpleapi.fit` called in `fitbenchmarking/fitting/mantid/main.py` now takes a mantid-native function as a FunctionWrapper object and a user-defined function as a string that describes a function called `fitFunction`. The function `fitFunction` is created from a class `fitFunction` defined in `fitbenchmarking/fitting/mantid/func_def.py`. The class `fitFunction` is a child class of IFunction1D (from mantid.api) and is defined from the function evaluation method of the problem object. All the fitting parameters are declared in the method`init` of the class and the function evaluation method is used in the method `function1D`. 

3. For fitting using scipy, the method `scipy.optimize.curve_fit` called in `fitbenchmarking/fitting/scipy/main.py` now takes the function and the initial parameters from the problem object method, get_function().

#### Testing Instructions

1. Run the example scripts and compare the results to the results produced from the master branch
2. Run the test files to check the correctness of the newly added methods


wathen:
There seems to be an error in the test set, could you fix this?

fitbenchmarking/fitting/mantid/func_def.py
```diff
@@ -41,45 +44,106 @@ def function_definitions(problem):
 
     problem_type = extract_problem_type(problem)
```
wathen: Do we need this anymore? We check the instance of the function and then by default use parse_function_definitions function

AtomAnu:
@wathen The problem is solved now.

AtomAnu:
@wathen For the function `extract_problem_type`, I still need it in the case where invalid problem type is inserted. Do you agree?

fitbenchmarking/fitting/mantid/func_def.py
wathen: Same goes for the scipy/func_def.py

wathen:
> @wathen For the function `extract_problem_type`, I still need it in the case where invalid problem type is inserted. Do you agree?

I think not now that we have a structure for how the function definitions are defined

AtomAnu:
@wathen Sure, would it be possible for me to complete this in a separate branch on a new issue? This is because many of the files and test files will have to be changed as a result.

wathen:
Can you remove all committed code and the benchmark_problems/test folder?

> @wathen Sure, would it be possible for me to complete this in a separate branch on a new issue? This is because many of the files and test files will have to be changed as a result.

Sure

fitbenchmarking/fitting/scipy/func_def.py
```diff
@@ -35,20 +34,8 @@ def function_definitions(problem):
     """
     problem_type = extract_problem_type(problem)
```
wathen: Can you add a test for this new function?

AtomAnu:
@wathen The test folder has been removed

wathen:
Can you also add a couple of tests for the eval_f functions?

fitbenchmarking/fitting/scipy/func_def.py
AtomAnu: `scipy/func_def.py` currently does not have its test file. It will try to add that accordingly.

AtomAnu:
@wathen How would you like the tests for eval_f functions to be?

wathen:
> @wathen How would you like the tests for eval_f functions to be?



> @wathen How would you like the tests for eval_f functions to be?

Well just make up a few points and make sure that the evaluation is correct

AtomAnu:
@wathen Sure I will add some tests in `test_fitbenchmark_data.py` and `test_nist_data.py`

AtomAnu:
Regarding the issue #195 

@wathen I have just added a test file for `scipy/func_def.py`

AtomAnu:
@wathen All the test files have been added. It is now ready for testing.

wathen:
I've been checking the results of this branch compared to master and all look identical apart from the results in the ENSO data set, do you know why this could be the case? I had a look at the function in ENSO and it is made up of trigonometric functions, do you think this is the issue?

AtomAnu:
@wathen I have just had a look at the difference in the results that you spotted. It appears to me that the results from the master branch is incorrect. Take a look at the two results here.

The first one is the ENSO plot produced by master.
Screen Shot 2019-08-06 at 2 50 38 PM

The second one is the ENSO plot produced by the functionEvaluationMethod branch.
Screen Shot 2019-08-06 at 2 52 29 PM

As you can see, on the first plot, **pi** happens to be one of the fitting parameters. This occur as the formula is passed into mantid fit as a string.

wathen:
Ok, great! I can just see one final thing before merging into master. Is it easy to change the form back to b1+b2*.... instead of fitFunction?

AtomAnu:
@wathen Sure! I will do that now. I have just finished integratig SasView further to allow SasView models to be used in mantid and scipy fitting modules.

AtomAnu:
@wathen I have just finished change the table display format.

Screen Shot 2019-08-07 at 11 40 58 AM

PR #220, opened 2019-09-10
Issue #134: Add option to options file for displaying absolute value,…

### LINKED ISSUE #134: Add absolute time column to tables measuring speed
Needs #167 to have been fixed.

Add option to fitbenchmarking_default_options.json, see #167, which allow user to output tables also in absolute numbers (some decision on unit of time need to be made).

Also, optionally, allow this option to be in 'both mode' which will result in tables that display both relative and absolution numbers in format: 1.2 (0.000034 s) where the number in bracket is the absolution number here in seconds.

AndrewLister-STFC:
… relative, or both

#### Description of Work
Add an option 'comparison_mode' to the options file. This will change the value reported in the results table.
If comparison_mode is:
- 'abs' : it will report the absolute value from the run (note: the colour will still be relative to the other results)
- 'rel' : it will function as it did before this change
- 'both' : it will display the absolute value followed by the relative in parentheses. e.g. 0.001274 (1)

Fixes #134

#### Testing Instructions

1. Set 'comparison_mode' in options file
2. Run example scripts
3. Check output is as expected

### DIFF
--- a/example_scripts/example_runScripts.py
+++ b/example_scripts/example_runScripts.py
@@ -37,15 +37,16 @@ custom_minimizers = None
 
 # SPECIFY THE MINIMIZERS YOU WANT TO BENCHMARK, AND AS A MINIMUM FOR THE SOFTWARE YOU SPECIFIED ABOVE
 if len(sys.argv) > 1:
-  # Read custom minimizer options from file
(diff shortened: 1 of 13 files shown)


wathen:
@AndrewLister-STFC can you change the width of the first column so that it is about the length of the name of the longest data set?
Table

Issue #219, opened 2019-09-10
Move documentation to Sphinx


tyronerees:
Move the docs from the GitHub wiki to Sphinx. 
- [x] Set up sphinx, with placeholders for user and developer documentation
- [x] Set up to compile and display on readthedocs.com
- [ ] Move across the existing documentation to the new format

AndrewLister-STFC:
To set up a readthedocs account for the project. The easiest method would be to reuse the FitBenchmarking account from GitHub.
@Anders-Markvardsen can you create the account [here](https://readthedocs.org/accounts/signup/)? There is a useful video [here](https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html) which mentions the readthedocs steps from around 14:20 to the end.
Alternately, I can do it if you give me the details.

Anders-Markvardsen:
Hi Andrew I have made your owner hence you then have the same permissions as I have. Can you now setup readthedocs?

AndrewLister-STFC:
Yes, I should be able to

AndrewLister-STFC:
Should be set up to automatically update here:
https://fitbenchmarking.readthedocs.io/en/latest/

tyronerees:
Reopening as the existing wiki hasn't been moved to the read-the-docs format

PR #221, opened 2019-09-11
Issue #219: Adds sphinx style docs to be hosted on readthedocs

AndrewLister-STFC:
#### Description of Work

Prep for #219
Adds rst files to create the start of the structure for the readthedocs documentation.
Also adds a makefile so that documentation can be generated locally using:
`cd docs`
`make html` 

#### Testing Instructions

Attempt to make the docs locally and check that it works correctly


docs/source/conf.py
```diff
@@ -0,0 +1,169 @@
...
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = '.rst'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = u'FitBenchmarking'
+copyright = u'2019, STFC'
```
tyronerees: Just flagging that this should be amended to be consistent with whatever is said in the new licence, when that appears.

PR #244, opened 2019-09-23
close plots when finished with

### LINKED ISSUE #227: Too many plots open error
Plots should be closed after they are saved.

Also need to solve reason for tihs symptom
```# Avoid reaching the maximum recursion depth by setting recursion limit
# This is useful when running multiple data set benchmarking
# Otherwise recursion limit is reached and the interpreter throws an error
sys.setrecursionlimit(10000)
```

Form GitHub history this was to do with a tkinter issue so matplotlib could be the cause of this.

tyronerees:
#### Description of Work

Fixes #227 

#### Testing Instructions

1. Run example_scripts/example_runScripts.py
2. Observe output: there should no longer be any complaints of too many plots open


tyronerees:
I've just created a new issue to deal with this warning.

I still need to investigate the recursion warning mentioned in #227 to see if that goes away now

AndrewLister-STFC:
> I've just created a new issue to deal with this warning.
> I still need to investigate the recursion warning mentioned in #227 to see if that goes away now

The recursion statement mentioned in that issue is just an extra line of code in the example scripts. We arent sure it's necessary, but it was added because of something with matplotlib according to the git history, so we thought it could be related.

tyronerees:
I think you're hunch is probably right @AndrewLister-STFC -- I'm currently running all the tests on this branch with the recursion limit line removed.  If that's successful, I'll remove this from the example scripts.

Issue #214, opened 2019-09-10
Simple tests


tyronerees:
Add some tests were we can work out the answer analytically.  These should include:
- [x] An unerdetermined linear least squares problem (known chi-squared for weighted and not, global min)
- [x] An overdetermined linear least squares problem (known chi-squared for weighted and not, global min)
- [x] An exact quadratic data fitting problem (chi-squared is zero)
- [x] An exact gaussian fitting problem (chi-squared is zero)

tyronerees:
Closed by #248

PR #248, opened 2019-09-24
214 Adding linear least squares problems

wathen:
#### Description of Work

#214 Adds linear least squares problems in `benchmark_problems/simple_tests`.

#### Testing Instructions

1. Check example scripts run with new problem set


tyronerees:
This PR Closed #214

PR #250, opened 2019-09-24
Issue #208: Add tests to run the example scripts

### LINKED ISSUE #208: Add automated testing of user facing example scripts
Add automated testing for the user facing example scripts in example_scripts folder.

These sits outside the main code in fitbenchmarking folder. These should be tested against

* that they are run with the expected software and minimizers
* that they don't fail

AndrewLister-STFC:
#### Description of Work

Fixes #208

#### Testing Instructions

1. Check Travis has run the additional tests
2. Check modified example scripts still run as standalone scripts


AndrewLister-STFC:
Travis failed due to timeout with Mantid. Perhaps we should take some problems out of the Mantid example?

example_scripts/example_runScripts_mantid.py
```diff
@@ -36,81 +36,86 @@
...
-                color_scale=color_scale, results_dir=results_dir)
-
-  print('\nCompleted benchmarking for {} problem set\n'.format(sub_dir))
+def main(argv):
+    # SPECIFY THE SOFTWARE/PACKAGE CONTAINING THE MINIMIZERS YOU WANT TO BENCHMARK
+    software = ["mantid"]
+    software_options = {'software': software}
+
+    # User defined minimizers
+    custom_minimizers = {"mantid": ["Simplex"],
```
tyronerees: Simplex is one of the slowest methods -- if we change this to levenberg marquardt, does this fix the timeout issue?

AndrewLister-STFC: I'll give it a go.

AndrewLister-STFC:
Travis didn't start automatically with my last commit.... I've restarted a previous failed test which should use the latest version, but the red/green dot will be on the wrong commit

example_scripts/example_runScripts_mantid.py
AndrewLister-STFC: Not sure if its Travis not picking up the new code or if it didn't work... I think closing a PR and opening it normally kicks off a test though?

Issue #242, opened 2019-09-20
Redundant plotting tools (sorting, extra plots created,..)


wathen:
**Description of the error**

Plots data, initial guess and best performing minimizer every time there is a better chisq value. 

Sorting function in in  fitting/plotting/plot_helper.py on lines 60 - 65 is of order n^2 and is called each time a new plot is created. This should be moved to the parsers.

**Describe the expected result**
Dramatic speed up

**Describe the actual result**
What was the actual result.

wathen:
Observation:

```
Changing to use argsort in plot_helper.py
mantidpython example_scripts/example_runScripts.py  329.26s user 14.22s system 106% cpu 5:22.58 total

Sorting in parsing
mantidpython example_scripts/example_runScripts.py  332.22s user 14.46s system 106% cpu 5:25.90 total

Master
mantidpython example_scripts/example_runScripts.py  445.65s user 16.82s system 104% cpu 7:20.80 total
```

wathen:
Sorting fixed in #261 and there is no redundant plotting at the moment

PR #261, opened 2019-09-30
Moving data sorting from plotting to data input

wathen:
### Description of Work

Fixes part of #242

Moves data sorting out of `fitbenchmarking/fitting/plotting` to `fitbenchmarking/parsing` so that the data is sorted once.

#### Testing Instructions

1. Check example results from master


fitbenchmarking/parsing/parse.py
```diff
@@ -41,6 +42,14 @@ def parse_problem_file(prob_file):
 
     logger.info("* Testing fitting of problem {0}".format(problem.name))
 
+    if problem.sort == True:
+        index = argsort(problem.data_x)
+        problem.data_x = problem.data_x[index]
+        problem.data_y = problem.data_y[index]
+        if isinstance(problem.data_e, ndarray):
+            problem.data_e = problem.data_e[index]
+
```
tyronerees: What if `sort==False`: if this for when we know the data is pre-sorted?

If this is meant for a user (or someone putting in a new model) to change, then we should mention it in the docs somewhere -- even if it just hangs there for now.  Otherwise, I think we should just sort by default, and add in a switch if and when we think it's necessary.

example_scripts/example_runScripts.py
```diff
@@ -66,7 +66,7 @@ def main(args):
     # problem_sets = ["CUTEst", "Muon", "Neutron", "NIST/average_difficulty", "NIST/high_difficulty", "NIST/low_difficulty",
     #                "SAS_modelling/1D"]
 
-    problem_sets = ["simple_tests"]
+    problem_sets = ["CUTEst", "Muon"]
```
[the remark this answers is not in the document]
wathen: I changed it to check both the nist and fitbenchmarking data types, shall I revert it back to the simple tests examples?

fitbenchmarking/parsing/parse.py
wathen: Lets add the switch later when we have done the cleaning in the problem definitions

example_scripts/example_runScripts.py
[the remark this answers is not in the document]
tyronerees: Yes please -- you raise a valid point about whether this should be the default for `problem_sets`, but we should only put changes that address the issue in a PR.

tyronerees:
For some reason this change slowed things down quite a bit on my system. I didn’t have time to investigate, and won’t be able to follow until I’m back in a week. I’ll pass over reviewing duty to @AndrewLister-STFC so this doesn’t get held up

fitbenchmarking/fitbenchmark_one_problem.py
```diff
@@ -131,10 +131,11 @@ def benchmark(controller, minimizers):
...
         if chi_sq < min_chi_sq:
             min_chi_sq = chi_sq
+            index = controller.sorted_index
             best_fit = plot_helper.data(name=minimizer,
-                                        x=controller.data_x,
-                                        y=controller.results,
-                                        E=controller.data_e)
+                                        x=controller.data_x[index],
+                                        y=controller.results[index],
+                                        E=controller.data_e[index])
```
AndrewLister-STFC: If use_errors is false this will fail?

AndrewLister-STFC:
> For some reason this change slowed things down quite a bit on my system. I didn’t have time to investigate, and won’t be able to follow until I’m back in a week. I’ll pass over reviewing duty to @AndrewLister-STFC so this doesn’t get held up

I had a look on mine, and found that the branch gives timings which are similar but slower.
As an example, I ran scipy against SasModelling/1D, simple_tests, and Muon. On the branch I get ~1min 10s and on master it's ~1min 3s

PR #279, opened 2019-10-30
Issue #277: Add documentation for options

### LINKED ISSUE #277: Add documentation on fitbenchmarking options
The docs for fitbenchmarking options are missing and should be added so they can be updated as the options evolve.

AndrewLister-STFC:
#### Description of Work

Fixes #277 

Adds user documentation explaining the options.

Can be viewed [here](https://fitbenchmarking.readthedocs.io/en/277-options-docs/users/options.html)

#### Testing Instructions

1. Read document and check that it makes sense.
2. Check formatting
3. Check for completeness

### DIFF
--- a/docs/source/users/index.rst
+++ b/docs/source/users/index.rst
@@ -9,4 +9,5 @@ Welcome to the documentation for the FitBenchmarking software.
 
     getting_started
     examples
+    options
 
--- a/docs/source/users/options.rst
+++ b/docs/source/users/options.rst
@@ -1,0 +1,146 @@
+.. _options:
+
+#######################
+FitBenchmarking Options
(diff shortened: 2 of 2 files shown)


AndrewLister-STFC:
> I've pushed a few changes directly.  I thought code inlining might be helpful to separate things in the code, but I'm not convinced by the red colour.  What do you think?  I'll approve, but won't merge until you've had a chance to look at it.

I don't mind either. The red text with boxes do make it a bit jarring, but they highlight the key bits.
Some of the things you've changed were in quotes to show that they were strings so I can add those in again and put the inline markup on the 3 headings at the bottom which now look out of place.

PR #293, opened 2019-11-12
273 controller factories

### LINKED ISSUE #273: Factory pattern for Controllers
When a new `Controller` is added, 4 additions need to be made (see `extending_fitbenchmarking.rst` for details).  This puts unnecessary burden on the user, and maximises the chance that a newly added `Controller` will not be correctly implemented. 

Implement a Factory pattern for `Controller` initialisation.  This will remove at least one of the additions (2. the Controller needs to imported and added to a dictionary in `fitbenchmarking/fitbenchmark_one_problem.py`.).  This could also be used to remove the minimiser lists from `fitbenchmaking_default_options.json` and include them in the specific Controller (although ideally you would just populate these lists directly from the software package).

**Additional context**
Once this is complete it would probably also be worth looking at parametrising the `Controller` tests, so that the user also isn't require to add basic tests (e.g. those which are fundamentally the same for each minimiser).

AndrewLister-STFC:
#### Description of Work

Fixes #273

Adds a factory implementation for the Controllers

#### Testing instructions

### DIFF
--- a/fitbenchmarking/fitting/controllers/controller_factory.py
+++ b/fitbenchmarking/fitting/controllers/controller_factory.py
@@ -11,7 +11,7 @@ from fitbenchmarking.fitting.controllers.base_controller import Controller
 
 class ControllerFactory:
     """
-    A factory for creating controllers.
+    A factory for creating software controllers.
     This has the capability to select the correct controller, import it, and
     generate an instance of it.
     Controllers generated from this must be a subclass of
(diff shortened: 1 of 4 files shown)


fitbenchmarking/fitting/controllers/controller_factory.py
```diff
@@ -0,0 +1,47 @@
...
+
+from importlib import import_module
+from inspect import isclass, isabstract, getmembers
+
+from fitbenchmarking.fitting.controllers.base_controller import Controller
+
+
+class ControllerFactory:
+    """
+    A factory for creating controllers.
```
wathen: Creating software controllers?

wathen:
```suggestion
    A factory for creating software controllers.
```

Issue #286, opened 2019-11-05
Additional RALFit options


wathen:
**Is your feature request related to a problem? Please describe.**
There are different options within `RALFit` (see [here](https://ralfit.readthedocs.io/projects/Python/en/latest/method.html)) which can be added to `fitbenchmarking_default_options.json`. 

**Describe the solution you'd like**
Pass additional options from json file to the `RALFit` controller.

tyronerees:
closed by PR #302

PR #302, opened 2019-11-18
RALFit weights and options

### LINKED ISSUE #297: RALFit with weights
**Description of the error**
RALFit doesn't use the weights at the moment.  Running `simple_tests` with `use_errors = True` gives the wrong answers:
ralfit_bad

**Suggested fix**
Change the `_prediction_error` function in the RALFit controller to take into account errors.

tyronerees:
#### Description of Work

Fixes #297, #286.  Also:
- returns a result when an exception is thrown by RALFit due to the solution not being found
- minor adjustments to docs

#### Testing Instructions

1. add ralfit to the `software` list in the `example_runScripts.py` or similar
2. run on a range of example cases
3. check results are as expected (particularly with `simple_tests`)

**Function**: Does the change do what it's supposed to?

**Tests**: Does it pass? Is there adequate coverage for new code?

**Style**: Is the coding style consistent? Is anything overly confusing?

**Documentation**: Is there a suitable change to documentation for this change?

**Bonus**: I'm not sure the exception handling is done in the best way.  Could the reviewer(s) look closely at this -- suggestions for improvements very welcome!

### DIFF
--- a/fitbenchmarking/fitting/controllers/ralfit_controller.py
+++ b/fitbenchmarking/fitting/controllers/ralfit_controller.py
@@ -21,7 +21,7 @@ class RALFitController(Controller):
         """
         super(RALFitController, self).__init__(problem, use_errors)
 
-        self._x = None
+        self._popt = None
         self._inform = None
 
     def setup(self):
@@ -59,18 +59,18 @@ class RALFitController(Controller):
         self.success = False
         try:
             options = {"model" : 1}
-            (self._x, self._inform) = ral_nlls.solve(self.initial_params,
-                                                     self._prediction_error,
-                                                     self._jac,
-                                                     params=(self.data_x,
-                                                             self.data_y),
-                                                     options=options)
+            (self._popt, self._inform) = ral_nlls.solve(self.initial_params,
+                                                        self._prediction_error,
+                                                        self._jac,
+                                                        params=(self.data_x,
+                                                                self.data_y),
+                                                        options=options)
         except Exception as err:
             # clear the exception
             sys.exc_clear()
             pass
         
-        self.success = (self._x is not None)
+        self.success = (self._popt is not None)
 
     def cleanup(self):
         """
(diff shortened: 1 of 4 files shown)


fitbenchmarking/fitting/controllers/ralfit_controller.py
```diff
@@ -20,42 +21,63 @@ def __init__(self, problem, use_errors):
...
+        j = approx_derivative(self._prediction_error,
+                              p)
+        return j
 
     def fit(self):
         """
         Run problem with RALFit.
         """
         self.success = False
         try:
```
AndrewLister-STFC: Looking at this now, I think this try, except is superfluous.

Since solve is throwing an exception, I wouldn't have thought it would be able to return something. So, unfortunately, there won't be any way to get the result.
```
self.success = False
self._popt = ral_nlls….
self.success = True
```
Would give a correct value for success.

tyronerees: I'll dig around and see if I can find an example where I get an exception for too many iterations (I was seeing it as a problem more when we were solving the wrong problem with the wrong Jacobian/not weighting correctly).  

A (nearly correct) answer was being returned with a try/catch on too many iterations, but not without -- I think this might have been to do with the try/catch that wraps `fit`, though?  Again, I'll try to find an example to show this behaviour.

tyronerees: Actually -- I take that back...with the code as it is, NaN is returned for both scipy and ralfit if the max number of iterations are hit -- NIST/average_difficulty has examples of both.  I therefore agree with @AndrewLister-STFC that this try/catch isn't needed -- I'll remove.

fitbenchmarking/fitting/controllers/ralfit_controller.py
```diff
@@ -4,6 +4,7 @@
 """
 
 import ral_nlls
+import sys
```
wathen: I can't see where sys is used, is there a reason for importing it?

tyronerees: Yes -- thanks!  It was left over from the exception handling that I took out yesterday.

PR #307, opened 2019-11-20
Adding CERN's minuit solver

### LINKED ISSUE #166: Investigate work involved to add support for Minuit
This was suggested by Miguel Gonzales after presentation at SINE2020 workshop in Lund, May 2019.

This issue is to investigate the work involved in adding support for minimizers https://www.cern.ch/minuit at the level of how this tool can be tested with SciPy and Mantid minimizers. Discuss this outcome and if go ahead create the relevant issue(s) to do this

tyronerees:
#### Description of Work

Fixes #166 

#### Testing Instructions

1. Add `minuit` as a software in, e.g., `example_runScripts.py`
2. Check results are consistent with other minimizers

### DIFF
--- a/fitbenchmarking/fitting/controllers/minuit_controller.py
+++ b/fitbenchmarking/fitting/controllers/minuit_controller.py
@@ -1,0 +1,72 @@
+"""
+Implements a controller for the CERN pacakage Minuit
+https://seal.web.cern.ch/seal/snapshot/work-packages/mathlibs/minuit/
+using the iminuit python interface
+http://iminuit.readthedocs.org
+"""
+from iminuit import Minuit
+import numpy as np
+
+from fitbenchmarking.fitting.controllers.base_controller import Controller
+
+class MinuitController(Controller):
+    """
+    Controller for the Minuit fitting software
+    """
+
+    def __init__(self, problem, use_errors):
+        """
+        Initializes variable used for temporary storage.
+        """
+        super(MinuitController, self).__init__(problem, use_errors)
+        self._popt = None
+        self._initial_step = None
+
+    def setup(self):
+        """
+        Setup for Minuit
+        """        
+        # minuit requires an initial step size.
+        # The docs say
+        # "A good guess is a fraction of the initial 
+        #  fit parameter value, e.g. 10% 
+        #  (be careful when applying this rule-of-thumb
+        #  when the initial parameter value is zero "
+        self._initial_step = 0.1 * np.array(self.initial_params)
+        # set small steps to something sensible(?)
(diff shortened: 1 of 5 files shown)


fitbenchmarking/fitting/controllers/minuit_controller.py
```diff
@@ -0,0 +1,72 @@
...
+        """        
+        # minuit requires an initial step size.
+        # The docs say
+        # "A good guess is a fraction of the initial 
+        #  fit parameter value, e.g. 10% 
+        #  (be careful when applying this rule-of-thumb
+        #  when the initial parameter value is zero "
+        self._initial_step = 0.1 * np.array(self.initial_params)
+        # set small steps to something sensible(?)
+        self._initial_step[self._initial_step < 1e-12] = 1e-12
```
wathen: Is `1e-12` what they suggest as a default tolerance too?

fitbenchmarking/fitting/controllers/minuit_controller.py
```diff
@@ -0,0 +1,72 @@
...
+            f = f / self.data_e
+
+        return np.dot(f,f)
+
+    def fit(self):
+        """
+        Run problem with Minuit
+        """
+        self.success = False
+        m = Minuit.from_array_func(self._prediction_error,
```
wathen: Does this make sense to put in the `setup` function so we just time `m.migrad()`?

fitbenchmarking/fitting/controllers/minuit_controller.py
tyronerees: No -- it says:

> You can set initital step sizes with keywords, `error_<name> = <step size>`. If you don't know what to pick, don't worry too much. Usually, the choice is not important. In the worst case, iminuit will use a few more function evaluations to find the minimum. A good guess is a fraction of the initial fit parameter value, e.g. 10 % (be careful when applying this rule-of-thumb when your the initial parameter value is zero).

`1e-12` is plucked out of thin air, in the absence of any other information.

wathen: Maybe once the options file is completed we have a tolerance for things being close to zero

We kinda use it here as well
https://github.com/fitbenchmarking/fitbenchmarking/blob/1c6d59583ce7c7b75922dbed8235998bf4768c83/fitbenchmarking/fitting/controllers/base_controller.py#L74-L75

fitbenchmarking/fitting/controllers/minuit_controller.py
tyronerees: Fixed.  I also moved the bit that saves as a scipy array into `cleanup`

PR #327, opened 2019-11-28
Issue #322: Refactor functions var in fitting_problem

### LINKED ISSUE #322: Restructure FittingProblem.functions to avoid repeating the callable
Currently, fitting_problems.FittingProblem.functions is of the form:
```
[[callable_0, starting_parameters_0],
 [callable_1, starting_parameters_1],
 ...]
```
However all of the callables are expected to be the same.

This should be changed so that starting parameters are separated from the callables, and we should change it to only allow a single callable until multiple are needed.

e.g.
```
FittingProblem.callable = callable_0
FittingProblem.starting_parameter_sets = [starting_parameters_0,
                                          starting_parameters_1,
                                          …]
```

AndrewLister-STFC:
#### Description of Work

Fixes #322 
Swaps from using a list of (callable, parameter) pairs to having a single callable, and changes how parameters are stored by the fitting problem.

This allows removing the function_id from calls to eval_f.

#### Testing Instructions

1. Verify accuracy does not change between master and branch. (I checked this already but it's good to be sure)


fitbenchmarking/fitting/controllers/scipy_controller.py
```diff
@@ -33,7 +33,7 @@ def fit(self):
         """
         popt = None
 
-        popt, _ = curve_fit(f=self.functions[self.function_id][0].__call__,
+        popt, _ = curve_fit(f=self.problem.function,
```
wathen: Should this not be `problem.eval_f`?

AndrewLister-STFC: This doesn't work with eval_f, but I'll add a comment

PR #338, opened 2019-11-29
294 move reused methods out of controllers

### LINKED ISSUE #294: Create a Jacobian and residual evaluation function in the fitting problem
When adding RALFit (#215) you have to supply Jacobians and residuals to the software. Currently the Jacobians is done by using `SciPy` in the RALFit controller. We should create two new functions `eval_j` and `eval_r` which evaluate the Jacobian and residual (`y - eval_f(x)`) in the fitting problem, similar to `eval_f`. 

Blocked by #283

AndrewLister-STFC:
#### Description of Work

Fixes #294 
Creates eval_r, eval_r_norm, and eval_j for evaluationg the residual, the square of the l2 norm for the residual and an approximation to the jacobian respectively.

Note: During this I observed that Minuit fails when running Muon, however this is also happening on master.

#### Testing Instructions

1. Check consistency with master

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -30,7 +30,8 @@ sys.path.insert(0, os.path.abspath('../../'))
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = ['sphinx.ext.autodoc',
-              'sphinx.ext.todo']
+              'sphinx.ext.todo',
+              'sphinx.ext.imgmath']
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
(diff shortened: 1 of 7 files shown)


AndrewLister-STFC:
> This all looks good.
> My only comment is that this should be explicit in the docs -- you don't want to force users to dig through code to find that these exist.  I'll add something...

Good point. In fact there's no documentation at all for the FittingProblem.
Should this be a separate issue?

tyronerees:
@AndrewLister-STFC -- could you build and look at the docs, and see if this is ok?  As discussed earlier, we don't have to fix on a format now, but I do think that some docs at this point is better than no docs.

Once you're happy with the docs, I'll approve and merge.

PR #342, opened 2019-12-10
Issue #287: Replace example script with bin/fitbenchmarking and add simple argparser

### LINKED ISSUE #287: Example run scripts should be renamed and should have argparser
The example run scripts run from main and take command line arguments.  As these currently use sys.argv, they have to be in a specific order, can be passed as any type, and have no associated documentation - the user has to look at the code in order to determine what command line arguments they can pass, what form they should take and what they do.

Use `argparser` to manage these command line arguments.  This should include:
- short and long tags for all arguments (e.g. `-f` and `--file`)
- types
- nargs
- help messages

AndrewLister-STFC:
#### Description of Work

Fix #287 

Moves example script to a fitbenchmarking script in a bin directory.
Updates the setup.py to install this script.
Moves code to a lib directory.
Implements very basic argparser.

#### Testing Instructions

1. Try with custom options and check results match expectations
2. Ensure selection of problem files behaves in a sensible way


bin/fitbenchmarking
```diff
@@ -0,0 +1,103 @@
...
+    if options_file != '':
+        # Read custom minimizer options from file
+        options_file = glob.glob(options_file)
+        options = Options(options_file)
+    else:
+        options = Options()
+
+    for sub_dir in problem_sets:
+        # generate group label/name used for problem set
+        label = sub_dir.replace('/', '_')
```
wathen: Will this `fitbenchmarking benchmark_problems/simple_tests benchmark_problems/Muon` create label names `benchmark_problems_simple_tests` and `benchmark_problems_Muon`? We should try and think of a way to keep the original names `simple_tests` and `Muon`

bin/fitbenchmarking
```diff
@@ -0,0 +1,103 @@
...
+
+
+def get_parser():
+    """
+    Creates and returns a parser for the args.
+    """
+
+    epilog = '''Usage Examples:
+
+    $ fitbenchmarking benchmark_problems/*
```
wathen: Will this work with the `nist` examples?

AndrewLister-STFC:
Huge drop in coverage is due to the untested files now affecting the rating... I checked against the latest master, and these are:
- fitbenchmarking/fitting/plotting/*
- fitbenchmarking/resproc/visual_pages.py
- fitbenchmarking/utils/fitbm_result.py

Also utils/logging_setup.py seems to now be tested but already had tests.

bin/fitbenchmarking
AndrewLister-STFC: You're right, I've updated it so that the example is ` $ fitbenchmarking benchmark_problems/NIST/*` although in future it would be nice to flatten this

AndrewLister-STFC:
This style of bin/script doesn't work on windows with the current setup.py. I'm looking up some options to get it installing on both again, so please don't merge until I do that (although I'm sure I'll have changes to make anyway)

AndrewLister-STFC:
In order to get it working on windows I think I will need to use this file structure:
```
fitbenchmarking
|\_ fitbenchmarking
|   |\_ cli
|   |   \__ fitbenchmarking.py
|   |\_ core
|   |   |\_ fitbenchmark_one_prob.py
|   |   \__ ...
|   |\_ parsers
|   |   |\_ sasview.py
|   |   |\_ mantid.py
|   |   \__ ...
|   \__ controllers
|       \_ ...
|\_ examples
|   \__ ...
|\_ setup.py
\__ ...
```
Then I can make the cli have entry points.

As opposed to:
```
fitbenchmarking
|\_ bin
|   \__ fitbenchmarking
|\_ lib
|   \__ fitbenchmarking
|       |\_ fitbenchmark_one_prob.py
|       |\_ ...
|       |\_ parsers
|       |   \__ ...
|       \__ ...
|\_ examples
|   \__ ...
|\_ setup.py
\__ ...
```

This also lines up with the advice [here](https://gehrcke.de/2014/02/distributing-a-python-command-line-application/)

AndrewLister-STFC:
Tested on windows with py3 (something in visual pages didn't work, but that's not related to this pr)

fitbenchmarking/cli/main.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def get_parser():
+    """
+    Creates and returns a parser for the args.
+    """
+
+    epilog = '''Usage Examples:
+
+    $ fitbenchmarking benchmark_problems/NIST/*
```
wathen: change `fitbenchmarking benchmark_problems/NIST/*` to change `fitbenchmarking examples/benchmark_problems/NIST/*` and the same below

AndrewLister-STFC: done

bin/fitbenchmarking
AndrewLister-STFC: For the scope of this issue I think we should leave it as is. @tyronerees suggested having an optional META.txt or similar in each problem set which could give info including the label. Otherwise default to the uderscores as is currently done. I'll create an issue this afternoon to address it.

AndrewLister-STFC: see #345

setup.py
```diff
@@ -30,5 +36,20 @@
           'externals': InstallExternals,
           'help': Help,
       },
-      package_data={'fitbenchmarking': ['resproc/color_definitions.txt']}
+      package_data={'fitbenchmarking': ['utils/default_options.ini']}
```
TOFarmer: I'm not sure what the distribution plan is for fitbenchmarking, but if we are going to make both source and built distributions, it is worth noting that apparently `package_data` [only includes files for built distributions](https://stackoverflow.com/questions/7522250/how-to-include-package-data-with-setuptools-distribute).

setup.py
```diff
@@ -1,16 +1,22 @@
...
 setup(name='FitBenchmarking',
       version='0.1.dev2',
       description='FitBenchmarking software',
       author='FitBenchmarking Team',
       url='http://github.com/fitbenchmarking/fitbenchmarking',
       license='GPL-3.0',
-      packages=find_packages(),
+      entry_points={
+        "console_scripts": [
+            'fitbenchmarking = fitbenchmarking.cli.fitbenchmarking:main'
```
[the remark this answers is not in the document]
AndrewLister-STFC: You're right, I forgot to update this when I changed the filename

setup.py
AndrewLister-STFC: Neither am I... but thanks for the pointer! It looks like problems in that thread are only relevant in the case of having a "MANIFEST.in" file? So this won't be a problem for us (until we add one).

Do you have any recommendations for alternatives, I see that there are a few ways people are claiming this should be done...?

TOFarmer: I think actually it suggests that the MANIFEST.in file is the way to include files for source distributions, so if we are going to have both source and built distributions we might need to include this as well.  This probably isn't the best forum for this, just thought I'd mention it!

fitbenchmarking/cli/main.py
```diff
@@ -0,0 +1,109 @@
+"""
```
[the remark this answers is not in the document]
AndrewLister-STFC: It wasn't intended to be, although I could change it if it would be better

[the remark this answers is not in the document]
TOFarmer: Well I'm not entirely sure what the convention is when executing with a CLI, but if you were executing with `python -m` if would run `__main__.py`, so I was wondering if it was an idea to be consistent with this.

Thinking about it a little more, what is the reason why `main.py` is in the cli subpackage?  Is it actually necessary?  As otherwise you could just put it in the directory above and name it `__main__`.  Then you could also run fitbenchmarking with `python -m`.

setup.py
AndrewLister-STFC: After some more digging I found that if there's no MANIFEST.in file, setuptools will add the package_data to the MANIFEST file. (https://docs.python.org/2/distutils/setupscript.html#installing-package-data)

TOFarmer: Ah excellent, well discovered!

fitbenchmarking/cli/main.py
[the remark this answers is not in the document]
AndrewLister-STFC: Yes, I guess it depends on what we want to do with this, if we want any other CLI commands, moving it out would be counter-productive. The reason I made a CLI dir is as an analogue to the bin dir, as well as separating it from the core code that we already have in fitbenchmarking.

Would it be better to add an `__main__.py` which is a pass-through to this?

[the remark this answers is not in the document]
TOFarmer: Agreed, that would probably make the most sense.  I think it would be worth doing that, as I think generally having a cli implies being executable as a module too.  It would also allow debugging with `python -m pdb -m fitbenchmarking` (at least when we support python 3.7).

PR #365, opened 2020-01-03
Add defaults if no x, y, e passed in

### LINKED ISSUE #364: `eval_r_norm` defaults
**Description of the error**
Defaults for `x`, `y` and `e` are not set in `eval_r_norm` 
https://github.com/fitbenchmarking/fitbenchmarking/blob/fe94b04e8a5b560784bc9005209f962f188c960c/fitbenchmarking/parsing/fitting_problem.py#L143-L161

If `x`, `y` and `e` are not passed in, then these should pick up values from 
`self`, as `eval_r` does.

`None` is used

**Suggested fix**
Implement this as in `eval_r`

tyronerees:
#### Description of Work

Fixes #364 

#### Testing Instructions

1. run for a solver that uses eval_r_norm (minuit or gsl), and check behaviour is as expected.

### DIFF
--- a/fitbenchmarking/parsing/fitting_problem.py
+++ b/fitbenchmarking/parsing/fitting_problem.py
@@ -157,6 +157,11 @@ class FittingProblem:
                  given parameters
         :rtype: numpy array
         """
+        if x is None and y is None and e is None:
+            x = self.data_x
+            y = self.data_y
+            e = self.data_e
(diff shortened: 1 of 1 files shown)


tyronerees:
Closing -- as pointed out by @wathen, this is done by the call to `eval_r`

PR #368, opened 2020-01-06
Refactoring html pages to use a template

### LINKED ISSUE #309: Refactor how supporting html and rst pages are created
Currently the supporting pages are hard coded in rst format. This should be refactored to use a software package which enables more customisation of these pages.

wathen:
#### Description of Work

Fixes #309 

Removes hard coded rst tables and then uses a template for the supporting HTML tables

#### Testing Instructions

1. Check supporting links work and pages are consistent with those before.


fitbenchmarking/resproc/visual_pages.py
```diff
@@ -52,133 +51,76 @@ def create(prob_results, group_name, results_dir, count):
...
-        rst_link = "<" + rst_file_path + "." + FILENAME_EXT_HTML + ">`__"
+    template = env.get_template('results_template.html')
+    html_link = "{0}.html".format(file_path)
 
-    return rst_link
+    with open(html_link, 'w') as fh:
+        fh.write(template.render(
+            title=prob_name,
+            equation=best_result.problem.equation,
+            initial_guess=best_result.ini_function_def.split("|")[1],
```
AndrewLister-STFC: This "eq | params" format was exclusively used by the old version of this file... we should change where this is saved to the result rather than splitting it here. Same for the "min_params"

wathen: done

fitbenchmarking/fitbenchmark_one_problem.py
```diff
@@ -157,17 +157,17 @@ def benchmark(controller, minimizers, options):
                                         sorted_index=index)
 
         problem = controller.problem
-        if 'fitFunction' in init_function_def:
-            init_function_def = init_function_def.replace(
+        if 'fitFunction' in init_function_params:
```
AndrewLister-STFC: These lines can be removed

wathen: done

fitbenchmarking/resproc/visual_pages.py
```diff
@@ -67,100 +65,33 @@ def create(prob_results, group_name, results_dir, count):
...
-        setup_detail_page_tbls(best_result.ini_function_def,
-                               best_result.fin_function_def)
+    prob_name = best_result.problem.name
+    prob_name = prob_name.replace(',', '')
+    prob_name = prob_name.replace(' ', '_')
 
+    support_pages_dir, file_path, see_also_link = \
+        setup_page_misc(group_name, prob_name,
+                        best_result, results_dir, count)
     fig_data, fig_fit, fig_start = \
```
AndrewLister-STFC: Did you say you removed the data plot? In that case this function might fail to find it? Or are we still plotting it and just not using it?

fitbenchmarking/resproc/visual_pages.py
```diff
@@ -67,100 +65,33 @@ def create(prob_results, group_name, results_dir, count):
...
-    rst_link = generate_rst_link(file_path)
-
-    ini_details_tbl, fin_details_tbl = \
-        setup_detail_page_tbls(best_result.ini_function_def,
-                               best_result.fin_function_def)
+    prob_name = best_result.problem.name
+    prob_name = prob_name.replace(',', '')
+    prob_name = prob_name.replace(' ', '_')
 
+    support_pages_dir, file_path, see_also_link = \
```
AndrewLister-STFC: If we're not using the `see_also_link` this could be removed in the function, and the function could be given a more descriptive name ("get_filename_and_path" or something rather than "misc")

fitbenchmarking/resproc/visual_pages.py
wathen: Good point, I have now removed the data plot now

fitbenchmarking/resproc/visual_pages.py
wathen: done

PR #373, opened 2020-01-07
334 add cutest

### LINKED ISSUE #334: Call cutest

tyronerees:
#### Description of Work

Fixes #334

#### Testing Instructions

1. Run with the SIF file problems and check it works
2. Some of the SIF and NIST definitions are the same -- do these give the same results?


tyronerees:
@AndrewLister-STFC -- as discussed, I'm not sure how to get the tests working with the new parser.  Could you please take a look?  So far:
- `basic.SIF` works as expected if run fitbenchmarking on the folder, but fails with `pytest`
- `start_end_x.SIF` doesn't work, due to the final `x` being disregarded by cutest.  Making this file just contain `NA` also didn't seem to work for me.  I've added a file `StartEndX.SIF` which sets the problem name as `NA`, but again I'm failing to get the test working with this.

Any hints would be appreciated!

AndrewLister-STFC:
Still TODO:
- [ ] Allow for errors (` IE E<num>` lines)
- [ ] Check for MLOWER|UPPER entries and adapt when rewriting X/Y for plotting

fitbenchmarking/parsing/cutest_parser.py
```diff
@@ -0,0 +1,240 @@
...
+        :rtype: lists of floats
+        """
+
+        if self.file.closed:
+            with open(self._filename, 'r') as f:
+                lines = f.readlines()
+        else:
+            lines = self.file.readlines()
+
+        if x is not None and len(x.shape) == 0:
```
TOFarmer: Can `len(x.shape) == 0` ever be `True`?  Even `len(np.array([]))` is 1.   I'm also a bit confused as to what you are trying to achieve here with the cast + additional list.  It feels like you are trying to convert a numeric type like `float` or `int` to a numpy array, although if that is the case `x.shape` would raise an `AttributeError`.

fitbenchmarking/parsing/cutest_parser.py
```diff
@@ -0,0 +1,240 @@
...
+                if x is None:
+                    self._m = int(line.split()[2])
+                    # this will always come before x/y data
+                    # so allocate space for these now
+                    data_x = np.zeros(self._m)
+                    data_y = np.zeros(self._m)
+                    data_e = np.zeros(self._m)
+                    # initialize index parameters for x and y
+                else:
+                    line = line[:col_width-1] + str(len(x))
```
TOFarmer: I could be wrong as I haven't paid especially close attention to the file format, however wouldn't you have to determine the amount of the line that you kept based on the length of the str you are adding to it e.g. `len(str(len(x)))`?

fitbenchmarking/parsing/cutest_parser.py
```diff
@@ -0,0 +1,240 @@
...
+                    self._m = len(x)
+
+            elif "IE MLOWER" in line and x is not None:
+                line = line[:col_width-1] + '1\n'
+
+            elif "IE MUPPER" in line and x is not None:
+                line = line[:col_width-1] + str(len(x))
```
TOFarmer: Same question as above.

fitbenchmarking/parsing/cutest_parser.py
```diff
@@ -0,0 +1,240 @@
...
+                    y_idx += 1
+                    line = line[:col_width-1] + '0.0'
+                else:
+                    continue
+
+            elif 'RE E' in line:
+                if x is None:
+                    data_e[e_idx] = float(line.split()[2])
+                    e_idx += 1
+                    line = line[:col_width-1] + '1.0'
```
TOFarmer: Similar to above re line lengths

fitbenchmarking/parsing/cutest_parser.py
```diff
@@ -0,0 +1,240 @@
...
+            with open(self._filename, 'r') as f:
+                lines = f.readlines()
+        else:
+            lines = self.file.readlines()
+
+        if x is not None and len(x.shape) == 0:
+            x = np.array([x])
+
+        x_idx, y_idx, e_idx = 0, 0, 0
+        data_x, data_y, data_e = None, None, None
```
TOFarmer: Not sure this assignment is necessary, as I'm pretty sure you reassign them if `x is None` and only use them if `x is None`.

fitbenchmarking/parsing/tests/test_parsers.py
```diff
@@ -142,6 +142,11 @@ def test_parsers(self, file_format, test_file, expected):
         assert (test_file is not None), \
             'No test file for {}'.format(file_format)
 
+        with open(test_file) as f:
```
TOFarmer: Unless I've missed something obvious, maybe this could do with a comment to explain why you are doing this.

fitbenchmarking/parsing/cutest_parser.py
```diff
@@ -0,0 +1,240 @@
+"""
+This file calls the pycutest interface for SIF data
```
TOFarmer: So from my memories of our weekly meetings, the complexity of the cutest parser is that in a lot of instances you need to write need files, which explains what's going on with the `TemporaryDirectory` and the writing in `_setup_data`.  I would recommend that you add something about this at the module level, as this is a specific quirk of the cutest parser that it would be useful to document - I imagine anyone with no cutest knowledge who has to work on this at a later date will thank you for it!

fitbenchmarking/parsing/cutest_parser.py
```diff
@@ -0,0 +1,240 @@
...
+
+        fp.function = self._function  # self._p.obj
+        fp.equation = None
+        fp.starting_values = self._get_starting_values()
+        fp.start_x = None
+        fp.end_x = None
+
+        # Use the object id as a hash to the function.
+        # In order to prevent the id being reused also store the x_data
+        self._cache = {id(fp.data_x): {'f': self._p.objcons, 'x': fp.data_x}}
```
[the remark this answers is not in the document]
AndrewLister-STFC: The issue we had is that x data could be very large and doing a comparison would have a time overhead (not good for benchmarking). So I was trying to use a faster comparison.

Ideally this would be very fast to compute so using the object id is pretty good.

We also know that the data object doesn't change throughout the main benchmarking (it does change when plotting/testing but time is not an issue then), so I thought this would be suitable.

fitbenchmarking/parsing/cutest_parser.py
```diff
@@ -0,0 +1,240 @@
...
+
+        return fp
+
+    def __del__(self):
```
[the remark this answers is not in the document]
AndrewLister-STFC: It does, I was hoping to remove the warning but I don't think this helps so I'll remove it.

fitbenchmarking/parsing/cutest_parser.py
AndrewLister-STFC: `np.array(5)` would have shape `()` so this is just making sure it's a 1-vector instead

fitbenchmarking/parsing/cutest_parser.py
AndrewLister-STFC: As far as I'm aware it's tabulated columns that are left justified:
```
col1                     col2
a_value                  another_value
third_value              fourth
```
So it's fine to replace everything from that point onwards

fitbenchmarking/parsing/cutest_parser.py
AndrewLister-STFC: Good point!

fitbenchmarking/parsing/cutest_parser.py
TOFarmer: Ah, fair enough. I didn't even realise that was a possibility, I thought the argument had to be a sequence.  In which case it would be more pythonic (and stop pylint complaining) if you were to do: `if x is not None and len(x.shape)`

fitbenchmarking/parsing/cutest_parser.py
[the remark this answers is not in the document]
TOFarmer: I guess the point I'm confused by is when would you be doing this comparison?  What is the series of events that needs to occur for `_function` to have already been called on `x`?

fitbenchmarking/parsing/cutest_parser.py
TOFarmer: Ah my mistake, I mistook col_width for a line width

fitbenchmarking/parsing/cutest_parser.py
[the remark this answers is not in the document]
AndrewLister-STFC: It happens whenever the software runs a new set of parameters. We fit `| y - f(x, p) |` where `x` and `y` are fixed, but for the purpose of plotting/testing, it's necessary to be able change `x`. During the fitting, `x` does not change and is passed in as `fitting_problem.data_x` which will always be the same object, however `p` varies until a local minima is reached.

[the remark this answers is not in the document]
AndrewLister-STFC: I'm more than happy to use a different way of achieving a fast comparison, however I can't think of a better alternative.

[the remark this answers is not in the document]
TOFarmer: Ok so basically within a single problem definition it will never enter the `Exception` because x is always constant?  But when changing to a new problem definition it will always change because it will use a different x (or the same x but with a different location in memory)?  Is that correct?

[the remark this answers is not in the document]
AndrewLister-STFC: Not quite, different problem definitions use separate parsers (and thereby they have their own caches).

During benchmarking, every minimizer uses the same fitting problem (and the x from it), so during benchmarking for a single problem it will not change. It will change when plotting as we sort x before evaluating to ensure the line plot is sensible. In future, we might over-sample x in plotting as well in order to get smooth curves.

As an example of the cache when fitting problems `t1.SIF` and `t2.SIF` with minimizers a, b, and c would give:
```
parse problem t1.SIF (add fitting_problem1.x to t1 cache)
fit t1.SIF using a (cache hit on fitting_problem1.x) <- function could be called many times as part of the fitting, but always with the same object
fit t1.SIF using b (cache hit on fitting_problem1.x)
fit t1.SIF using c (cache hit on fitting_problem1.x)
parse problem t2.SIF (add fitting_problem2.x to t2 cache)
fit t2.SIF using a (cache hit on fitting_problem2.x)
fit t2.SIF using b (cache hit on fitting_problem2.x)
fit t2.SIF using c (cache hit on fitting_problem2.x)
plot result for t1.SIF using a (t1 cache miss, add sorted(fitting_problem1.x)) <- sorted(fitting_problem1.x) stored in plotting object so the same object will be used each time
plot result for t1.SIF using b (cache hit on sorted(fitting_problem1.x))
plot result for t1.SIF using c (cache hit on sorted(fitting_problem1.x))
plot result for t2.SIF using a (t2 cache miss, add sorted(fitting_problem2.x))
plot result for t2.SIF using b (cache hit on sorted(fitting_problem2.x))
plot result for t2.SIF using c (cache hit on sorted(fitting_problem2.x))
```

fitbenchmarking/parsing/cutest_parser.py
AndrewLister-STFC: There's a comment above where I define col_width that I can try to make clearer

fitbenchmarking/parsing/cutest_parser.py
```diff
@@ -0,0 +1,240 @@
...
+            x = data_x
+
+        file_path = os.path.join(self.mastsif_dir.name,
+                                 '{}.SIF'.format(str(id(x))[-10:]))
+
+        with open(file_path, 'w') as f:
+            f.writelines(to_write)
+
+        # check the data is right
+        # TODO: turn into an exception
```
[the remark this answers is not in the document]
AndrewLister-STFC: We now have FitBenchmarking exceptions, so I'll just add it to this.

fitbenchmarking/parsing/cutest_parser.py
```diff
@@ -0,0 +1,240 @@
...
+        :return: The fully parsed fitting problem
+        :rtype: fitbenchmarking.parsing.fitting_problem.FittingProblem
+        """
+        self.mastsif_dir = TemporaryDirectory()
+
+        # set the MASTSIF environment variable so that pycutest
+        # can find the sif files
+        os.environ["MASTSIF"] = self.mastsif_dir.name
+
+        self._n = None
```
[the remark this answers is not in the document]
AndrewLister-STFC: I think this was written by Tyrone, I'll ask if we need to add it as a attribute as I'm not sure it's used after this func

fitbenchmarking/parsing/cutest_parser.py
AndrewLister-STFC: This doesn't get flagged by my linter... but I'll change it

PR #379, opened 2020-01-08
Optional plotting

### LINKED ISSUE #378: Making plotting optional
Adds option to the options file whether or not to produce plots.

Create new `make_plots` option in options file.

wathen:
#### Description of Work

Fixes #378 

Adds option to make plots

#### Testing Instructions

1. Check option is correctly set and executed


fitbenchmarking/resproc/visual_pages.py
```diff
@@ -79,19 +79,21 @@ def create(prob_results, group_name, results_dir, count, options):
...
+                          options=options,
+                          count=count,
+                          group_results_dir=directory)
+        plot.plot_initial_guess()
+        plot.plot_best_fit(best_result.minimizer, best_result.params)
+
+        fig_fit, fig_start = \
+            get_figure_paths(support_pages_dir, prob_name, count)
+    else:
+        fig_fit = fig_start = "Plot_option_is_turned_off"
```
tyronerees: Does this need to have underscores?  Can we just set this to "Plot option is turned off" (or, possibly, "Re-run with make_plots set to yes in the ini file to generate plots")

fitbenchmarking/utils/options.py
```diff
@@ -43,6 +43,7 @@ def __init__(self, file_name=None):
         self.use_errors = fitting.getboolean('use_errors')
 
         plotting = config['PLOTTING']
+        self.make_plots = plotting.getboolean('make_plots')
```
tyronerees: can we wrap this around a try-catch?  At the moment the code falls over with a cryptic error message if we send anything other than a boolean.

wathen: A try-catch around `make_plots`? I'm not sure what you exactly mean and would we not have to do that with self.use_errors as well?

fitbenchmarking/resproc/visual_pages.py
wathen: How it was before you needed the underscores otherwise it wouldn't print the whole message. I have changed the template slightly so that it checks `make_plots` and then either returns the plots or the message

fitbenchmarking/utils/options.py
wathen: I added a try-catch around `make_plots`, let me know what you think about the message

fitbenchmarking/utils/options.py
```diff
@@ -43,6 +47,11 @@ def __init__(self, file_name=None):
         self.use_errors = fitting.getboolean('use_errors')
 
         plotting = config['PLOTTING']
+        try:
+            self.make_plots = plotting.getboolean('make_plots')
+        except ValueError:
```
tyronerees: Can you add a test that hits this exception?  That will keep the coverage percentage up

wathen: Done

PR #389, opened 2020-01-10
Adding style sheet to tables

### LINKED ISSUE #388: Results tables different style to rest of html pages
It looks like the css style sheet is not used for the results tables

wathen:
#### Description of Work

Fixes #388 

Added blank template which is used to generate the results tables with the style sheet attached.

#### Testing Instructions

1. Check the new table style is consistent with the other html pages

### DIFF
--- a/fitbenchmarking/core/results_output.py
+++ b/fitbenchmarking/core/results_output.py
@@ -217,9 +217,8 @@ def render_pandas_dataframe(table_dict, minimizers, html_links,
         output_file = os.path.join(name + 'html')
 
         with open(output_file, "w") as f:
-            f.write(template.render(
-                css_style_sheet=style_css))
-            f.write(table_style.render(table_styles=style_css))
+            f.write(template.render(css_style_sheet=style_css))
+            f.write(table_style.render())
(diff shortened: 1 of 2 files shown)


fitbenchmarking/core/results_output.py
```diff
@@ -208,14 +208,17 @@ def colour_highlight(value):
...
-        with open(name + 'html', "w") as f:
-            f.write(style)
-            f.write(table_style.render(table_styles=style_css))
+        root = os.path.dirname(inspect.getfile(fitbenchmarking))
+        html_page_dir = os.path.join(root, 'HTML_templates')
+        style_css = os.path.join(html_page_dir, 'style_sheet.css')
+        env = Environment(loader=FileSystemLoader(html_page_dir))
+        template = env.get_template("blank_page.html")
+
+        output_file = os.path.join(name + 'html')
```
AndrewLister-STFC: This is a no-op. I think join just puts a '/' in between each of it's args (or on other os' could put a different symbol in).
`name + 'html'` should be fine.

PR #352, opened 2019-12-17
Refactor systems tests

AndrewLister-STFC:
#### Description of Work

Fixes #291 

Implements the systems tests as described in my comment in #291 
Namely: 
- [x] Removes outdated test
- [x] Adds test to compare accuracy of a single minimizer from each software using simple_tests against a previous version
- [x] Adds test to compare accuracy of a single scipy minimizer using a problem set with a single problem for each parser against a previous version

Note, also Fixes #308 and some other small things:
- [x] Update paths in output files to be relative. This is needed as the test file has links in it, which won't work if it expects absolute paths that were generated from another machine.
- [x] Add a writer for the options. This means dynamically generating an options file is much easier. 

#### Testing Instructions

1. Did the new tests pass?!

Function: Does the change do what it's supposed to?
Tests: Does it pass? Is there adequate coverage for new code?
Style: Is the coding style consistent? Is anything overly confusing?
Documentation: Is there a suitable change to documentation for this change?

PR #394, opened 2020-01-14
135 make all plots

### LINKED ISSUE #135: A way to view plot for each individual fit
Feedback from Tyrone. It would be useful to see plots of all fits. 

Two ideas for doing this

* Add those plots to existing results page at the bottom
* Add link to separate page containing these plots to the result page

This could be a separate issue: have each number in the table link to a plot

AndrewLister-STFC:
#### Description of Work

Fixes #135 

Restructures outputting so that resulting files are created in the following order:
 - All directories
 - All plots
 - All support pages
 - All tables
 - Index page

Adds links from table cells to a connected support page.

Refactors plotting to remove repeated calculations.

#### Testing Instructions

1. Test table looks as it should, links work, and plots are correct on the supporting page


AndrewLister-STFC:
Leaving this as draft until #352 is merged

fitbenchmarking/results_processing/visual_pages.py
```diff
@@ -171,11 +107,16 @@ def get_figure_paths(support_pages_dir, problem_name, count):
     :rtype : tuple(str, str)
     """
 
-    figures_dir = os.path.join(support_pages_dir, "figures")
-    figure_fits = os.path.join(figures_dir, "Fit_for_" + problem_name +
-                               "_" + str(count) + ".png")
-    figure_strt = os.path.join(figures_dir, "start_for_" + problem_name +
-                               "_" + str(count) + ".png")
+    if result.figure_link == '' and result.start_figure_link == '':
+        message = 'Re-run with make_plots set to yes in the ini file to' \
```
wathen: With make_plots off the output looks like this:
plotting
Is it possible to remove the directory name before the message?

fitbenchmarking/core/results_output.py
```diff
@@ -1,151 +1,233 @@
...
+            plot.plot_best(best.minimizer, best.params)
+
+        # For each result, if it succeeded, create a plot and add plot links to
+        # the resuts object
+        for result in prob_result:
+            if result.params is not None:
+                plot_path = plot.plot_fit(result.minimizer, result.params)
+                result.figure_link = plot_path
+            else:
+                result.figure_link = 'Failed to return params'
```
wathen: Similarly to the make_plots off, is it possible to remove the directory path and produce the full message?
Screenshot at 2020-01-15 14-08-09

fitbenchmarking/core/results_output.py
```diff
@@ -1,151 +1,233 @@
...
+    generate_tables(results, best_results, table_names, options.table_type)
+    return table_names
 
 
-def generate_tables(results_per_test, minimizers,
-                    linked_problems, table_names,
-                    table_suffix):
+def generate_tables(results_per_test, best_results, table_names, table_suffix):
     """
     Generates accuracy and runtime tables, with both normalised and absolute
```
wathen: Can we say combined tables as well?

fitbenchmarking/utils/create_dirs.py
```diff
@@ -28,60 +28,51 @@ def results(results_dir):
...
-
-
-def restables_dir(results_dir, group_name):
-    """
-    Creates the results directory where the tables are located.
-    e.g. fitbenchmarking/results/neutron/
+    Creates the results directory for a specific group.
+    e.g. fitbenchmarking/results/Neutron/
 
     @param results_dir :: directory that holds all the results
```
wathen: Shall we update the docstrings here?

fitbenchmarking/utils/fitbm_result.py
```diff
@@ -45,9 +45,17 @@ def __init__(self, options=None, problem=None, fit_status=None,
...
+        self.support_page_link = ''
+        self.start_figure_link = ''
+        self.figure_link = ''
+        self.html_print = False
+
     def __str__(self):
         if self.table_type is not None:
             output = self.table_output
+            if self.html_print:
+                output = '<a target="_blank" href="{0}">{1}</a>'.format(
```
wathen:
```suggestion
                output = '<a href="{0}">{1}</a>'.format(
```

wathen:
In `fitting_benchmarking` we create the results directories 
https://github.com/fitbenchmarking/fitbenchmarking/blob/63d43061108c8405e500ecb3eff94d6117aa2a41/fitbenchmarking/core/fitting_benchmarking.py#L38-L39

which are then re-created in
https://github.com/fitbenchmarking/fitbenchmarking/blob/63d43061108c8405e500ecb3eff94d6117aa2a41/fitbenchmarking/core/results_output.py#L60-L61

It would be good to remove the first occurrence of this. In addition, the `directory` variable in `fitbenchmark_one_problem` 
https://github.com/fitbenchmarking/fitbenchmarking/blob/63d43061108c8405e500ecb3eff94d6117aa2a41/fitbenchmarking/core/fitbenchmark_one_problem.py#L15
is no longer used

AndrewLister-STFC:
Good spot

fitbenchmarking/core/results_output.py
AndrewLister-STFC: Done

fitbenchmarking/core/results_output.py
```diff
@@ -165,69 +253,99 @@ def select_table(value, table_suffix):
...
+        :rtype: fitbenchmarking.utils.fitbm_result.FittingResult
+        '''
+        result.html_print = True
+        return result
+
     for name, title, table in zip(table_names.values(), table_title,
                                   table_dict.values()):
-        table.index = html_links
+        # Update table indexes to link to the best support page
+        index = ['<a target="_blank" '
```
wathen:
```suggestion
        index = ['<a  '
```

AndrewLister-STFC:
> When column labels you are linked to the best minimiser support page, is it possible to only plot one in line for the figure? Currently, you see two lines both from the same minimiser:
> 
> If this is easy, then we should put and if statement in the support page template saying that this is best minimiser and thus the plot only has the data and the fitted result

Which line should we keep? Just the green?

wathen:
> > When column labels you are linked to the best minimiser support page, is it possible to only plot one in line for the figure? Currently, you see two lines both from the same minimiser:
> > If this is easy, then we should put and if statement in the support page template saying that this is best minimiser and thus the plot only has the data and the fitted result
> 
> Which line should we keep? Just the green?

Yea just the green line

PR #398, opened 2020-01-16
Change defaults to pip installables only

### LINKED ISSUE #343: Defaults should only include pip installable packages
**Description of the error**
With a new install, users are required to install mantid and ralfit before running.

Defaults should not include software that have complex installs

**Suggested fix**
Make the defaults simpler and provide an `all_softwares.ini` options file for those who do want to install everything

AndrewLister-STFC:
#### Description of Work

Fixes #343 

Remove gsl from setup.py
Update defaults to pip installable modules only
Edit docs to state that some modules are not installed by default

#### Testing Instructions

1. Check installs work


AndrewLister-STFC:
Docs can be seen here: https://fitbenchmarking.readthedocs.io/en/343-change_defaults_to_only_pip/

Anders-Markvardsen:
pip install . works and fitbenchmarking examples\benchmark_problems\NIST\average_difficulty runs:-)

@AndrewLister-STFC could you check that the removal of './' does not unintentionally also remove another desired effect?

AndrewLister-STFC:
> pip install . works and fitbenchmarking examples\benchmark_problems\NIST\average_difficulty runs:-)
> @AndrewLister-STFC could you check that the removal of './' does not unintentionally also remove another desired effect?

This still works on my machine, but I'll grab a blank vm and test it there

AndrewLister-STFC:
This works on a blank vm. I did notice 1 issue which I'll fix while we're looking at installs.

Anders-Markvardsen:
@AndrewLister-STFC, last Friday, was it this issue you mentioned something about unknown or known reason why the build is failing?

fitbenchmarking/systests/test_regression.py
```diff
@@ -23,6 +23,7 @@ def setUpClass(cls):
 
         opts = Options()
         opts.minimizers = {k: [v[0]] for k, v in opts.minimizers.items()}
+        opts.software = sorted(opts.minimizers.keys())
```
Anders-Markvardsen: could you add comment to why you sort keys here please?

AndrewLister-STFC: done

AndrewLister-STFC:
Currently getting a segfault after controllers pass.
More specifically, after the controllers tests run, travis shows a pass for all tests and the returns a segfault. This only occurs on the first run of the tests for the vm which is making debugging very difficult

PR #402, opened 2020-01-21
Local minimizer table

### LINKED ISSUE #401: Local minimizer table
Currently there is not a table that checks to see if the results are local minimums.

A new table where each cell is a true or false which corresponds to whether a local minimum has been found.

wathen:
#### Description of Work

Fixes #401 

Adds an additional table which indicates whether the software converges to a local minimum

#### Testing Instructions

1. Check table is consistent.

### DIFF
--- a/fitbenchmarking/core/results_output.py
+++ b/fitbenchmarking/core/results_output.py
@@ -20,6 +20,14 @@ error_options = {0: "Successfully converged",
                  2: "Software run but didn't converge to solution",
                  3: "Software raised an exception"}
 
+ACC_DESCRIPTION = "The accuracy results are calculated from the final " \
+                  " chi squared value."
+RUNTIME_DESCRIPTION = "The runtime results are calculated using the timeit " \
+                      " module in python"
+COMPARE_DESCRIPTION = "The combined results show the accuracy in the first " \
+                      " line of the cell and the runtime on the second line " \
+                      "of the cell."
+
 
 def save_results(options, results, group_name):
     """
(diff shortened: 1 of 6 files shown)


fitbenchmarking/utils/fitbm_result.py
```diff
@@ -4,6 +4,8 @@
 
 import numpy as np
 
+LOCAL_MIN_TOL = 1e-2
```
tyronerees: I wonder if even this is too tight?  For example, the local minima for the gaussian tests are coming up as non local minima.  I think we should raise this to `1e-1`, but we also need to make sure it's clear to the user what we're telling them here (although the same is true for all the tables, so I don't think we need to do that here).

wathen:
> The code looks neater without the if statements, but the UI seems to have taken a step backwards in going from this:
> old_layout
> to this:
> new_layout
> 
> Can we go back to the same formatting we had before? Ideally with the combined tables at the top again, as I'm sure that gets used most.

I have reverted the ordering back to what it was before and removed the indentation of the buttons.

fitbenchmarking/utils/fitbm_result.py
wathen: I've set the tolerance to be `1e-1` but not reported this to the users yet. As you said, more information about the tables should be added but I don't think this is the place to do that

fitbenchmarking/core/results_output.py
```diff
@@ -15,11 +15,24 @@
...
+RUNTIME_DESCRIPTION = \
+    "The runtime results are calculated using the timeit module in python."
+COMPARE_DESCRIPTION = \
+    "The combined results show the accuracy in the first line of the cell " \
+    "and the runtime on the second line of the cell."
+LOCAL_MIN_DESCRIPTION = \
+    "The local min results show whether the software has converged to a " \
+    " local minimum."
+
+SORTED_TABLE_LINKS = ["compare", "acc", "runtime", "local_min"]
```
wathen:
```suggestion
SORTED_TABLE_NAMES = ["compare", "acc", "runtime", "local_min"]
```

fitbenchmarking/utils/fitbm_result.py
```diff
@@ -68,6 +70,33 @@ def __init__(self, options=None, problem=None, fit_status=None,
...
+
+    @property
+    def local_min(self):
+        if self.params is not None:
+            r = self.problem.eval_r(self.params)
+            min_test = np.matmul(self.problem.eval_j(self.params).T, r)
+            norm_r = np.linalg.norm(r)
+            norm_min_test = np.linalg.norm(min_test)
+            self.norm_rel = norm_min_test / norm_r
+            if norm_r <= LOCAL_MIN_TOL or norm_min_test <= LOCAL_MIN_TOL \
```
tyronerees: Following on from the discussion friday, can we set two tolerances, `GRAD_TOL` and `RES_TOL`, and test `norm_r` with `RES_TOL` and the other two with `GRAD_TOL`.  We can keep the tolerance to `GRAD_TOL` as is, but tighten `RES_TOL` to, say, `1e-8`.

wathen:
> I still prefer the original titles on the index page sections -- "Combined accuracy and runtime results" is much more informative than "compare table". Was your intention to to link this with label the options the user sends in? If so, I think we should highlight that somewhere (in the docs?)
> 
> The description seemed more natural to me when it came before the buttons -- there were clear sections then, which are a little disjointed now.

I was going to do the docs as part of the refactoring but not here since we have changed the layout completely

fitbenchmarking/utils/fitbm_result.py
wathen: Done

PR #405, opened 2020-01-22
Issue #385: Move data corrections from controller to fitting problem

### LINKED ISSUE #385: Move data preparation from controllers to fitting_problem
Currently the data preparation (truncation at start/end x and estimate of error) is done in the controller and copied back to the problem.

This should be moved to the fitting problem to avoid the copying.

AndrewLister-STFC:
#### Description of Work

Fixes #385 

Move data corrections out of controller and into fitting problem

#### Testing Instructions

1. This should not change anything

### DIFF
--- a/fitbenchmarking/results_processing/plots.py
+++ b/fitbenchmarking/results_processing/plots.py
@@ -156,9 +156,10 @@ class Plot(object):
         plot_options_dict['label'] = label
         plot_options_dict['color'] = self.best_fit_plot_options['color']
 
+        x = self.problem.data_x[self.problem.sorted_index]
         self.plot_data(errors=False,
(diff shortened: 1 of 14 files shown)


fitbenchmarking/results_processing/plots.py
```diff
@@ -156,9 +156,10 @@ def plot_best(self, minimizer, params):
         plot_options_dict['label'] = label
         plot_options_dict['color'] = self.best_fit_plot_options['color']
 
+        x = self.problem.data_x[self.problem.sorted_index]
```
wathen: This should be moved to the `__init__` to avoid duplicate code.

PR #409, opened 2020-01-24
265 simple tests fb

### LINKED ISSUE #265: Simple tests for each type
The simple tests work very well. It would be useful if this was also available in the other formats to check consistency.

We should add datasets mimicing the simple_tests but for each of our problem types (e.g. Sasview, and Fitbenchmarking).

tyronerees:
#### Description of Work

Addresses part of #265 by duplicating the simple tests in NIST format for Mantid functions.

Also removed some simple tests that didn't add anything to the range of test cases.

#### Testing Instructions

1. Run `simple_tests` for some minimizers
2. Check results are as expected


examples/benchmark_problems/simple_tests/cubic-fb.dat
```diff
@@ -0,0 +1,8 @@
+# FitBenchmark Problem
```
wathen: To be consistent with the Neutron and Muon problem sets shouldn't we use a txt file extension for the new simple tests?

examples/benchmark_problems/simple_tests/cubic-fb.dat
```diff
@@ -0,0 +1,8 @@
...
+# This is a simple test problem where we interpolate points that lie exactly on a cubic polynomial.
+software = 'mantid'
```
wathen: The issue talks about sasview problem sets as well, do you think its worth adding them in the simple tests?

examples/benchmark_problems/simple_tests/gauss_exact-fba.dat
```diff
@@ -0,0 +1,8 @@
+# FitBenchmark Problem
```
wathen: Running this problem with `mantid` `Levenberg-Marquardt` gives very different results compared to the other format:
results

wathen:
Do you think we should put one of these new problems into the system tests?

examples/benchmark_problems/simple_tests/gauss_exact-fba.dat
tyronerees: I don't think there's anything we can do here -- must be a numerical issue due to the way the functions are handled.

examples/benchmark_problems/simple_tests/cubic-fb.dat
tyronerees: fixed

examples/benchmark_problems/simple_tests/cubic-fb.dat
tyronerees: I don't know if that's possible at the moment -- I don't know how user functions are defined in SasView (the tutorial just says how to do them in the GUI, as far as I can tell).  Since the SasView/native parsers have now been merged, I'd say this is less of an issue than it was.

PR #432, opened 2020-02-04
391 exceptions

### LINKED ISSUE #391: FitBenchmarking exceptions
Observed by @AndrewLister-STFC  in #379.

Create a specific exceptions for FitBenchmarking, e.g., have an `OptionsError` for when incorrect values are set.

Create FitBenchmarking exception errors to be wrapped around existing functions

AndrewLister-STFC:
#### Description of Work

Fixes #391 

Implements a collection of custom exceptions and exception handler

#### Testing Instructions

1. Check exceptions are printed correctly


fitbenchmarking/utils/exceptions.py
```diff
@@ -0,0 +1,127 @@
+"""
+This file holds all FitBenchmarking exceptions, organised by exception id
+"""
+
+
+class FitBenchmarkException(Exception):
```
wathen: All looks good but is it possible to test this file too?

AndrewLister-STFC: I'm not sure how to test this file... Ideally I think this would be tested as a by-product of testing the files that we expect to raise them, although that would be a separate task.

Issue #441, opened 2020-02-12
Activate ties/constraints on non Mantid minimizers


AndrewLister-STFC:
**Is your feature request related to a problem? Please describe.**
Currently, there are Mantid problems that cannot be benchmarked against other softwares as they include complex features which are not available elsewhere (ties/constraints/…).

**Describe the solution you'd like**
It would be good if some (or even 1) other fitting software was available to benchmark against Mantid for these.

**Additional context**
As an example anything that uses ties could provide vastly differing results between mantid and say scipy as scipy will not use the ties.

Relates to closed issue #121 (closed as line in description was removed)
Relates to issue #303, although this has been moved to a feature request rather than bug as it is now documented clearly, and is a missing feature rather than a broken one.

AndrewLister-STFC:
Looks like this was fixed in some sense in #750. Not confident that it covers ties in the general sense, but it allows for fixing values. E.g. I think in mantid, one can use ties to describe $h=frac{1}{d}$ which I'm not sure we allow for.

PR #442, opened 2020-02-12
Issue #303: Document native fitbenchmarking file format and available…

### LINKED ISSUE #303: Mantids ties are ignored by software that is not mantid
**Description of the error**
Using ties on a mantid function with non-mantid fitting software seems to have no impact. This could explain why GEM is problematic.

Setting a tie should fix a parameter.

Running FitBenchmarking with ties and without gives the same result.

**Suggested fix**
We should consider whether this is something that should be supported if we have to jump through hoops to make it work for non-mantid fitting software.

**Additional details**
This currently works for Mantid fitting.

AndrewLister-STFC:
… functionality for mantid wrt ties

#### Description of Work

Fixes #303 

Documents Mantid problem format and available features for mantid.

Note: I have raised #441 to track that ties and other advanced features are not available with non mantid minimizers.

#### Testing Instructions

1. Try tie/constraint/fix with mantid (e.g. `name=Gaussian,PeakCentre=0,Height=1,Sigma=0.1,constraints=(0<Sigma<1);name=Gaussian,PeakCentre=1,Height=1,Sigma=0.1;ties=(f1.Sigma=f0.Sigma)`)
2. Proof read documentation


fitbenchmarking/parsing/fitbenchmark_parser.py
```diff
@@ -245,7 +245,7 @@ def _get_starting_values(self):
         :return: Starting values for the function
         :rtype: list of OrderedDict
         """
-        ignore = ['name', 'BinWidth', 'ties', 'Formula']
+        ignore = ['name', 'BinWidth', 'ties', 'Formula', 'constraints']
```
wathen: Can you add a simple comment to explain this?

PR #470, opened 2020-02-26
Switch to python3

### LINKED ISSUE #213: Move from Python 2 to Python 3
Python 2 is being sunsetted on January 1st 2020:
https://www.python.org/doc/sunset-python-2/
and advice from Python is that "You should upgrade to Python 3 as soon as you can".

The last version of NumPy to support Python 2.7 was numpy 1.16.x (it's now 1.17.2). The last SciPy version to do so was 1.2.x (it's now at 1.3.1). FitBenchmarking should be upgraded to ensure it continues to use the latest software, and that it remains easy to install.

AndrewLister-STFC:
#### Description of Work

Fixes #213 

Enables code to work on python 3 including installation of all softwares.

#### Testing Instructions

1. Load up python3 and try it out


AndrewLister-STFC:
Still need to check over documentation.

Anders-Markvardsen:
On Windows and using Python 3.7.3.

* `pip install .` is successful with final message "Successfully installed FitBenchmarking-0.1.dev2 h5py-2.10.0"
* however unfortunately running `fitbenchmarking examples\benchmark_problems\NIST\average_difficulty` gives the error

```
C:\Midlertiddig\fitbenchmarking>fitbenchmarking examples\benchmark_problems\NIST\average_difficulty

Running the benchmarking on the NIST_average_difficulty problem set

  File "c:\python37\lib\site-packages\fitbenchmarking\cli\exception_handler.py", line 22, in wrapped
    return f(*args, **kwargs)
  File "c:\python37\lib\site-packages\fitbenchmarking\cli\main.py", line 102, in run
    data_dir=data_dir)
  File "c:\python37\lib\site-packages\fitbenchmarking\core\fitting_benchmarking.py", line 47, in fitbenchmark_group
    parsed_problem.correct_data()
  File "c:\python37\lib\site-packages\fitbenchmarking\utils\output_grabber.py", line 22, in __exit__
    self.stop()
  File "c:\python37\lib\site-packages\fitbenchmarking\utils\output_grabber.py", line 44, in stop
    self.origstream.flush()
OSError: [WinError 1] Incorrect function

During handling of the above exception, another exception occurred:

  File "C:\Python37\Scripts\fitbenchmarking-script.py", line 11, in <module>
    load_entry_point('FitBenchmarking==0.1.dev2', 'console_scripts', 'fitbenchmarking')()
  File "c:\python37\lib\site-packages\fitbenchmarking\cli\main.py", line 147, in main
    run(problem_sets=args.problem_sets, options_file=args.options_file)
  File "c:\python37\lib\site-packages\fitbenchmarking\cli\exception_handler.py", line 29, in wrapped
    print('Unknown exception. Exiting.')
OSError: [WinError 1] Incorrect function
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
OSError: [WinError 1] Incorrect function
```

Any suggestions?

Anders-Markvardsen:
Note on the same laptop but within a ubuntu terminal both installation and running the above command works. So yes a Windows issue. Before trying to dig into this @AndrewLister-STFC or @wathen  from the error message anything obvious?

AndrewLister-STFC:
> 
> 
> Note on the same laptop but within a ubuntu terminal both installation and running the above command works. So yes a Windows issue. Before trying to dig into this @AndrewLister-STFC or @wathen from the error message anything obvious?

Nothing from me that isn't obvious in the traceback:
- The issue seems to be with windows and the OutputGrabber
- More specifically windows doesn't like the call to flush

I don't think the outputGrabber has changed in this PR though, so this is potentially only an issue on python3?

fitbenchmarking/core/results_output.py
```diff
@@ -129,8 +129,8 @@ def find_between(s, start, end):
     with open(filename) as f:
         output_str = f.read()
     output_str = output_str.replace(':ref:', '')
-    reload(sys)
-    sys.setdefaultencoding('utf-8')
+    #reload(sys)
```
Anders-Markvardsen: Are these two lines meant to deleted or un-commented?

Anders-Markvardsen:
following on from discussion earlier in weekly friday catch-up. Found this https://eli.thegreenplace.net/2015/redirecting-all-kinds-of-stdout-in-python/ which explains by redirect_stdout will not work for capturing e.g. mantid output. Note managed to get this to work on windows yet. An alternative, but more heavyweight approach would be to create subprocesses and capture from these but I think this should probably only be considered as a last resort

Anders-Markvardsen:
Offending line to Windows appear to be `os.dup2(self.pipe_in, self.origstreamfd)` which is meant to point stdout (self.origstreamfd) to be pipe_in here but keeping its fd (file descriptor number). This works on linux but appears not working on windows as code written. There are two work arounds 1) ugly and using ctypes 2) pretty and using subprocesses. However, 2) may have undesirable timing side effects. Perhaps we can talk about this in next weekly fridag catchup

Anders-Markvardsen:
@wathen could you provide an example where the existing output_grabber grabs stdout which is not using Mantid. This is for debugging. Example meaning a minimizer + problem definition directory, ideally something which is quick to run. Thanks

Anders-Markvardsen:
@wathen separately to the above. When I run this PR using Ubuntu with scipy + NIST/average_difficulty I observe

```
###############################
 Running data from: MGH17 7/9
###############################

    Starting value: 1/2
        Software: SCIPY
            Minimizer: lm-scipy-no-jac
<string>:1: RuntimeWarning: overflow encountered in exp
<string>:1: RuntimeWarning: invalid value encountered in add
<string>:1: RuntimeWarning: overflow encountered in multiply
            Minimizer: lm-scipy
            Minimizer: trf
            Minimizer: dogbox
    Starting value: 2/2
        Software: SCIPY
            Minimizer: lm-scipy-no-jac
            Minimizer: lm-scipy
            Minimizer: trf
            Minimizer: dogbox
```

e.g. "RuntimeWarning: overflow encountered in exp" above looks to me to come from Scipy. Were the existing output_grabber meant to grab these?

wathen:
> @wathen could you provide an example where the existing output_grabber grabs stdout which is not using Mantid. This is for debugging. Example meaning a minimizer + problem definition directory, ideally something which is quick to run. Thanks

I think it is actually just mantid that you can't suppress the output. I will do a few test to see if I can get an example though

wathen:
> @wathen separately to the above. When I run this PR using Ubuntu with scipy + NIST/average_difficulty I observe
> 
> ```
> ###############################
>  Running data from: MGH17 7/9
> ###############################
> 
>     Starting value: 1/2
>         Software: SCIPY
>             Minimizer: lm-scipy-no-jac
> <string>:1: RuntimeWarning: overflow encountered in exp
> <string>:1: RuntimeWarning: invalid value encountered in add
> <string>:1: RuntimeWarning: overflow encountered in multiply
>             Minimizer: lm-scipy
>             Minimizer: trf
>             Minimizer: dogbox
>     Starting value: 2/2
>         Software: SCIPY
>             Minimizer: lm-scipy-no-jac
>             Minimizer: lm-scipy
>             Minimizer: trf
>             Minimizer: dogbox
> ```
> 
> e.g. "RuntimeWarning: overflow encountered in exp" above looks to me to come from Scipy. Were the existing output_grabber meant to grab these?

These warnings must happen outside the `output_grabber`,  my guess is that they occur when processing the results. For example, calculating the final `chisq` value

Anders-Markvardsen:
@wathen would you be OK to fix this conflict while keeping the new travis structure introduced in this PR and without reverting the GSL change you have introduced in #497?

PR #526, opened 2020-05-13
Enable Azure to run on most unit tests

### LINKED ISSUE #511: Enable Azure to run on most unit tests
Following first step #510.

This will require modification to output_grappers.py to enable run this as 'pass through' on windows.

Anders-Markvardsen:
#### Description of Work

Enable Azure to run on most unit tests.

Enabled on all tests except

fitbenchmarking\controllers\test\test_externals_controllers.py
fitbenchmarking\cli\tests\test_main.py (scheduled to be removed or heavily modified in other issue)

Fixes #511

#### Testing Instructions

1. check code
2. check tests parse


azure-pipelines.yml
```diff
@@ -26,6 +26,10 @@ steps:
 
 - script: |
     pip install pytest pytest-azurepipelines
-    pytest fitbenchmarking\utils\tests\test_logger.py
-#    pytest fitbenchmarking\utils\tests\test_fitbm_result.py (breaks in first azure pipeline install)
+    pytest fitbenchmarking\utils
```
wathen, line 29: I think it would be nice to make an `azure` folder similar to the `travis` folder and have two scripts in it, one to install (this would be super short) and the other to run the tests

fitbenchmarking/controllers/tests/test_basic_controllers.py
```diff
@@ -1,3 +1,4 @@
+# Tests for controllers available from basic fitbenchmarking install
```
wathen, line 1: We should probably use `default` instead of `basic` here and for the file name too

fitbenchmarking/controllers/tests/test_externals_controllers.py
```diff
@@ -0,0 +1,246 @@
+# Tests for controllers for externals, i.e. those available in addition
+# to those from a basic fitbenchmarking install
```
wathen, line 2: same as before `basic` to `default`

azure-pipelines.yml
Anders-Markvardsen: will address this in separately created issue

wathen: Is there a particular reason for this? It is definitely in scope for this issue and will make sure that any future tests are put in correctly

Anders-Markvardsen: Yes. I could hack something together, but in my opinion it deserve some thinking, is not urgent, and it would be recommended to check a few other repos and advice on this before splitting

PR #531, opened 2020-05-15
Software links in tables

### LINKED ISSUE #196: Add way to get info about each minimizer from table
An example table output is:

Capture

In meeting today we agreed that the user should have the option to click on any of the minimizer names in the top row and then be presented with information about the minimzers which the user clicked on. Ideally this would just be an URL for each minimizer.

Important this option must be optional.

wathen:
#### Description of Work

Fixes #196 

Adds links to the read the docs how page minimizer list.

#### Testing Instructions

1. Check that all the links work for all minimizers


fitbenchmarking/results_processing/base_table.py
```diff
@@ -217,11 +217,19 @@ def to_html(self, table, colour, links, error):
         :return: HTLM table output
         :rtype: str
         """
+        link_template = '<a href="https://fitbenchmarking.readthedocs.io/en/'\
+            'latest/concept/how/Minimizers.html#{0}">{0}</a>'
```
tyronerees, line 221:
```suggestion
            'latest/concept/how/Minimizers.html#{0}" target="_blank">{0}</a>'
```

tyronerees: I think this should open the docs in a new tab, which might be better than navigating away from the results page completely.

wathen: Yea I like that better. I have created a PR to fix the `sasview`/`bumps` naming issue so once that goes in then the naming conversion should work with the links

PR #569, opened 2020-06-12
Combining benchmarking functions in core

### LINKED ISSUE #567: Combine `fitting_benchmarking.py` and `fitbenchmark_one_problem.py`
By combing `fitting_benchmarking.py` and `fitbenchmark_one_problem.py` there would be two entry ports to FitBenchmarking.

wathen:
#### Description of Work

Fixes #567 

#### Testing Instructions

1. Check the new structure is clearer


fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -36,13 +50,62 @@ def fitbenchmark_group(group_name, options, data_dir):
...
+    # Loops over benchmark problems #
+    #################################
+    results, failed_problems, unselected_minimzers = \
+        loop_over_benchmark_problems(problem_group, options)
+
+    for keys, minimzers in unselected_minimzers.items():
+        minimizers_all = options.minimizers[keys]
+        options.minimizers[keys] = list(set(minimizers_all) - set(minimzers))
+
+    # options.minimizers = converge_minimizers
```
Anders-Markvardsen, line 67: Did you mean to leave `# options.minimizers = converge_minimizers` here?

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -36,13 +50,62 @@ def fitbenchmark_group(group_name, options, data_dir):
...
+    # If the results are and empty list then this means that all minimizers
```
Anders-Markvardsen, line 68:
```suggestion
    # If the results are an empty list then this means that all minimizers
```

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -36,13 +50,62 @@ def fitbenchmark_group(group_name, options, data_dir):
...
+    # raise an exception and the tables will produce errors if they run.
+    if results == []:
+        message = "The current options set up meant that all minimizers set " \
+                  "raised an exception. This is likely due to the " \
+                  "`algorithm_type` set in the options. Please review " \
+                  "current options setup and re-run FitBenmarking."
```
Anders-Markvardsen, line 74:
```suggestion
        message = "The user chosen options setup resulted in that all minimizers " \
                  "raised an exception. This is likely due to the way " \
                  "`algorithm_type` was set in the options. Please review " \
                  "your options setup and re-run FitBenmarking."
```

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -65,37 +128,204 @@ def fitbenchmark_group(group_name, options, data_dir):
...
+        #########################
+        problem_result, minimizer_failed = \
+            loop_over_minimizers(controller=controller,
+                                 minimizers=minimizers,
+                                 options=options)
+        unselected_minimzers[s] = minimizer_failed
+        software_results.extend(problem_result)
+
+    # Checks to see if all of the minimizers raise and exception and
+    # records the problems name for that case
```
Anders-Markvardsen, line 227:
```suggestion
    # Checks to see if all of the minimizers raised and exception and
    # record the problem name if that is the case
```

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -65,37 +128,204 @@ def fitbenchmark_group(group_name, options, data_dir):
...
+        # Catching all exceptions as this means runtime cannot be calculated
+        # pylint: disable=broad-except
+        except Exception as excp:
+            if isinstance(excp, UnknownMinimizerError):
+                minimizer_failed.append(minimizer)
+                minimizer_check = False
+            LOGGER.warn(str(excp))
+
+            runtime = np.inf
+            controller.flag = 3
```
Anders-Markvardsen, line 286: Unrelated to this issue I think. How well do we currently describe the different flag values? Do we document this beyond `def check_attributes(self):` in base_controller.py?

TOFarmer: I agree with Anders, this doesn't appear to be documented anywhere - I think you have to look at the code in fitbenchmark_one_problem.py to determine what the flags mean.

wathen: The documentation is [here](https://fitbenchmarking.readthedocs.io/en/latest/extending/controllers.html) for when you add a new controller. Do you think we should add more somewhere else?

fitbenchmarking/core/fitting_benchmarking.py
wathen: Updated

fitbenchmarking/core/fitting_benchmarking.py
wathen: Was a few grammer things so updated accordingly

fitbenchmarking/core/fitting_benchmarking.py
wathen: Updated

fitbenchmarking/core/fitting_benchmarking.py
TOFarmer: I thought it would be almost as quick to solve the problem as explain my suggested solution, so I've created an issue and a PR - no problems if you want to go down a different route for this.

wathen: I have just looked over this PR and I agree that this is a much nicer way of doing it

fitbenchmarking/core/fitting_benchmarking.py
wathen: Good spot, I have removed this

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -65,37 +127,204 @@ def fitbenchmark_group(group_name, options, data_dir):
...
+    :type minimizers: list
+    :param options: FitBenchmarking options for current run
+    :type options: fitbenchmarking.utils.options.Options
+
+    :return: list of all results and dictionary of unselected minimizers
+             based on algorithm_type
+    :rtype: tuple(list of fibenchmarking.utils.fitbm_result.FittingResult,
+                  list of failed minimizers)
+    """
+    grabbed_output = output_grabber.OutputGrabber(options)
```
TOFarmer, line 252: I know you mentioned the other day that ideally this should be a class (which I agree with), but in the meantime should we maybe just pass the `grabbed_output` to this point (from `loop_over_benchmark_problem`) rather than having to calculate it again here?

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -65,37 +127,204 @@ def fitbenchmark_group(group_name, options, data_dir):
...
+    unselected_minimzers = {}
+    software_results = []
+    for s in software:
+        LOGGER.info("        Software: %s", s.upper())
         try:
-            results_dict[name].append(problem_result)
+            minimizers = options.minimizers[s]
         except KeyError:
-            results_dict[name] = [problem_result]
+            raise UnknownMinimizerError(
```
TOFarmer, line 206: Could I suggest `UnsupportedMinimizerError` or something similar?  At first read I thought this meant it was being raised because an unknown error had propagated from the minimizer, rather than because there wasn't a minimizer with that name.

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -6,19 +6,33 @@
...
+from fitbenchmarking.utils import fitbm_result, misc, output_grabber
+from fitbenchmarking.utils.exceptions import NoResultsError, \
+    UnknownMinimizerError
 from fitbenchmarking.utils.log import get_logger
 
 LOGGER = get_logger()
 
 
-def fitbenchmark_group(group_name, options, data_dir):
+def benchmark(group_name, options, data_dir):
```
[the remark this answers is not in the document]
wathen: Corrected

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -36,13 +50,61 @@ def fitbenchmark_group(group_name, options, data_dir):
...
+                  "`algorithm_type` was set in the options. Please review " \
+                  "your options setup and re-run FitBenmarking."
+        raise NoResultsError(message)
+
+    # Used to group elements in list by name
+    results_dict = {}
+    for problem_result in results:
+        name = problem_result.name
+        # group by name
+        try:
```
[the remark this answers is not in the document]
wathen: This looks much neater, I have updated the code

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -65,37 +127,204 @@ def fitbenchmark_group(group_name, options, data_dir):
...
+
+def loop_over_fitting_software(problem, options, start_values_index):
+    """
+    Loops over fitting software selected in the options
+
+    :param problem: a problem object containing information used in fitting
+    :type problem: FittingProblem
+    :param options: FitBenchmarking options for current run
+    :type options: fitbenchmarking.utils.options.Options
+
```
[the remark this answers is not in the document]
wathen: corrected

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -36,13 +50,61 @@ def fitbenchmark_group(group_name, options, data_dir):
...
+    results_dict = {}
+    for problem_result in results:
+        name = problem_result.name
+        # group by name
+        try:
+            results_dict[name].append(problem_result)
+        except KeyError:
+            results_dict[name] = [problem_result]
+
+    results = [results_dict[r] for r in
```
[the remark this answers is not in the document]
wathen: Did not run correctly with the code snippet but with `sorted(results_dict.keys(), key=lambda x: x.lower()` it worked fine

fitbenchmarking/core/fitting_benchmarking.py
wathen: Agreed, I have added a new exception and updated the `UnknownMinimizerError` exception as well

fitbenchmarking/core/fitting_benchmarking.py
wathen: Agreed and corrected

fitbenchmarking/core/fitting_benchmarking.py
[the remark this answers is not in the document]
TOFarmer: Ah yes, I didn't spot that the grouping wasn't just to enable your sorting.  It feels like there should be an even more straightforward way of doing this without resorting to using `defaultdict`, but I can't think of it.  Unfortunately `itertools.groupby` isn't very helpful in this instance, as you don't want to keep the keys.

PR #573, opened 2020-06-17
Refactor and document Controller.flag

### LINKED ISSUE #572: Controller.flags should become a property and should be documented within
Two related minor problems:
- The only documentation of what the `int`s for `Controller.flags` mean is in controllers.rst (i.e. not in the code)
- Whether a `flag` is valid is determined in `Controller.check_attributes`, which therefore has two different functions.

TOFarmer:
#### Description of Work

Fixes #572 

This results in a marginally less nicely formatted display of the flags in controllers.html, so I guess it needs to be determined if this is acceptable.  Arguably the correct solution would be to use sphinx autodoc to have a selection of the code documentation (which I don't think is currently the case, although I could be wrong) included and then simply link to the location of `Controllers.flag` in controllers.rst, rather than including it directly.

Also ideally you would have both the `int` `flag` values and the associated `str` meanings in a class level `OrderedDict` within `Controller`, and then dynamically generate the `Controller.flag` docstring from this: this would mean that the the `flag` values and their meanings are always together.  However I decided that implementing this was more hassle than it was worth, particularly as `VALID_FLAGS` is unlikely to change that often.

NB: I haven't checked if a test for the `flags` validation already exists - if not, this should probably be created.

#### Testing Instructions

1. Check that Adding Fitting Software documentation is still acceptable
2. Check that invalid flags can't be passed to a `Controller` object


TOFarmer:
Ah yes, well spotted - I've moved it now.

TOFarmer:
> NB: I haven't checked if a test for the flags validation already exists - if not, this should probably be created.

So this has come back to bite me!  Travis failed because there is a test which checks this but it assumes the checking is being done in `check_attributes`.  I'll change these tests to account for this.

However one further question which is raised: It seems like `check_attributes` is only called in `benchmark`.  I guess this is to test that `Controller.flag` has actually been set at some point, and so isn't still `None`.  So do we want to add an `assert Controller.flag` in here?

TOFarmer:
> However one further question which is raised: It seems like check_attributes is only called in benchmark. I guess this is to test that Controller.flag has actually been set at some point, and so isn't still None. So do we want to add an assert Controller.flag in here?

Never mind, I think that is redundant actually.

It should all be good to go now.

Issue #200, opened 2019-08-23
Display numpy and matplotlib messages using the logger format


Anders-Markvardsen:
When running `python example_runScripts_expert.py` a number of 3rd party library warnings are not displayed (and logged?) in the logger format.

Examples of logger formatted messages are:
```
[08:09:09]  ERROR main.py: Warning, fit failed. Going on. Error: Optimal parameters not found: Number of calls to function has reached maxfev = 500.
[08:09:09]  ERROR main.py: Warning, fit failed. Going on. Error: Optimal parameters not found: The maximum number of function evaluations is exceeded.
```

Examples of 3rd party library warning messages:
```
C:\MantidInstall\bin\lib\site-packages\matplotlib\axes\_axes.py:519: UserWarning: No labelled objects found. Use label='...' kwarg on individual plots.
 warnings.warn("No labelled objects found. "
C:\Midlertiddig\fitbenchmarking\fitbenchmarking\parsing\nist_data_functions.py:1: RuntimeWarning: invalid value encountered in power
C:\MantidInstall\bin\lib\site-packages\matplotlib\pyplot.py:516: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (matplotlib.pyplot.figure) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam figure.max_open_warning).
 max_open_warning, RuntimeWarning)
C:\Midlertiddig\fitbenchmarking\fitbenchmarking\resproc\numpy_restables.py:70: RuntimeWarning: All-NaN slice encountered
 min_chi_sq = np.nanmin(accuracy_tbl, 1)
C:\Midlertiddig\fitbenchmarking\fitbenchmarking\resproc\numpy_restables.py:71: RuntimeWarning: All-NaN slice encountered
 min_runtime = np.nanmin(time_tbl, 1)
```
These are from numpy and matplotlib and of type RuntimeWarning and UserWarning.

This issue is to catch these warning with the FitBenchmarking logger, thereby 
* making the displayed warning all look the same and easier to read
* these warning messages correctly does not specify from where they originate in the FitBenchmarking code
* have one record of all warnings 

Observed while testing #199

Anders-Markvardsen:
Most if not all matplotlib errors has gone

Anders-Markvardsen:
Out of date issue and new with examples provided with `fitbenchmarking --help`

PR #598, opened 2020-07-28
Adding NIST analytic Jacobians

### LINKED ISSUE #482: Analytic jacobians in NIST format
Extend the NIST format to include analytic Jacobians, and extend the NIST parser.

Possibly, have companion files <nist_name>.jac which hold the Jacobian information.

wathen:
#### Description of Work

Fixes #482 

Adds analytic Jacobians to the three NIST example

#### Testing Instructions

1.
2.
3.


tyronerees:
All looks good.  There is some printing of errors to the screen in the table generation, which we hadn't seen before.  I think we should investigate these as part of #200 (which we should fix before the release)

PR #599, opened 2020-08-04
Setting Jacobian in Mantid controller

### LINKED ISSUE #485: Set the Jacobian in Mantid
If we're using a mantid minimizer, get it to use the Jacobian we give it, rather than using it's own approximation.

wathen:
#### Description of Work

Fixes #485 

For none Mantid data the Mantid controller will now use the user defined Jacobian from the options.

#### Testing Instructions

1. Check the `chi_sq` values are consistent between this branch and master
2. Check that the order of the two loops in `functionDeriv1D` does not matter with respect to time and `chi_sq`


tyronerees (commented):
The changes do what they say on the tin.

Coveralls is complaining that the new lines in functionDeriv1D aren't being hit.  I don't see how this can be the case -- the tests cover it to my eyes -- but could you have a look?

wathen:
> The changes do what they say on the tin.
> 
> Coveralls is complaining that the new lines in functionDeriv1D aren't being hit. I don't see how this can be the case -- the tests cover it to my eyes -- but could you have a look?


It seems that the normal function evaluation is not covered either. Maybe this is to do with how `Mantid` does the evaluation?
https://github.com/fitbenchmarking/fitbenchmarking/blob/03d0325aaa8e40df9f47ef876324833ad9882fe6/fitbenchmarking/controllers/mantid_controller.py#L141-L146

tyronerees (approved):
Coverage drop seems to be because this class is called internally by mantid, which isn't picked up by the coverage tools.  All seems good.

PR #604, opened 2020-08-13
Updating first example docs

### LINKED ISSUE #589: Use a better first example
The suggested example is to use the `simple_tests` problem set.

This includes Mantid data, and so will fail if Mantid isn't installed.  Change the default to be, e.g., a set of NIST problems, so that it can be run independently of third party software.

wathen:
#### Description of Work

Fixes #589 

Updates the first example in the docs to be from the NIST problem set. Also, allows problem sets with different problem types (`mantid`, `NIST`, etc) to be run without errors if an external package is not installed. 

#### Testing Instructions

1. Check new first example runs without any warnings for default installation.
2. Check `simple_tests` run with just a warning for the mantid problems
3. Check a Mantid data set (`Neutron` or `Muon`) raise a `FitBenchmarking ran with no results` error.

### DIFF
--- a/docs/source/users/running.rst
+++ b/docs/source/users/running.rst
@@ -10,7 +10,7 @@ directory along with a template for passing in options.
 To run with any of these problem sets, you should pass the path to the
 `fitbenchmarking` command.
 
-  e.g ``fitbenchmarking examples/benchmark_problems/NIST/high_difficulty``
+  e.g ``fitbenchmarking examples/benchmark_problems/NIST/low_difficulty``
 
 An options file can also be passed with the ``-o`` argument.
 The template in `examples` is heavily documented and serves as a good
--- a/docs/source/users/running.rst
+++ b/docs/source/users/running.rst
@@ -10,7 +10,7 @@ directory along with a template for passing in options.
 To run with any of these problem sets, you should pass the path to the
 `fitbenchmarking` command.
 
-  e.g ``fitbenchmarking examples/benchmarking_problems/simple_tests``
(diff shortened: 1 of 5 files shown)


wathen:
> Ok here is my review:
> 
>     1. Run fine but take a while to do so, maybe use a smaller/easier test set? Also some strange warnings (hard to tell where these are coming from):
> 
> 
> ```
>  <string>:1: RuntimeWarning: invalid value encountered in power 
>  'NoneType' object has no attribute 'T'
> ```
> 
>     1. Run fine with the expected warning for mantid problems:
> 
> 
> ```
> Missing dependencies for fit.
> Details: Requirements are missing for mantid parser: No module named 'mantid'
> ```
> 
>     1. Raises the expected error along with an informative message (but with weird spaces between words):
> 
> 
> ```
> Error while running FitBenchmarking. Exiting. See below for more information.
> FitBenchmarking ran with no results
> Details: The user chosen options and/or problem setup resulted in  all minimizers and/or parsers raising an exception.  This is likely due to the way `algorithm_type` was set  in the options or the selected problem set requires  additional software to be installed. Please review your  options setup and/or problem set then re-run FitBenmarking.
> ```
1) I changed the default to NIST low diff data set however this still produces the warning you have seen. Maybe it would be better to run with a smaller number of minimizers as a default. I will create an issue for this.

3) Fixed spacing issues.

jfowkes (approved):
Yeah I'd say either a) run with a smaller number of minimisers as default or b) use a subset of NIST low difficulty containing just a few really easy problems.

Otherwise looks great to me! Approving as the above will go into another issue.

PR #605, opened 2020-08-14
103 setup docker

### LINKED ISSUE #103: Run all tests inside docker image
Currently, [Travis](https://travis-ci.com/mantidproject/fitbenchmarking) is installing Mantid on a virtual machine before running any tests. This is currently required since Mantid is needed to run the current suite of tests. As one might expect, downloading and installing Mantid every time a suite is run takes a considerable amount of time. To make the testing more efficient and less convoluted, run all the tests inside a docker image. More information on docker is available [here](https://www.docker.com/). Travis is also compatible with docker.

tyronerees:
#### Description of Work

Fixes #103 

#### Testing Instructions

1. Check the tests run
2. Check you can run the docker container locally
3.


docs/source/contributors/structure.rst
```diff
@@ -19,6 +20,27 @@ Build (``build``)
 This directory contains scripts to allow for installing packages such as Mantid
 through setuptools.
 
+*******************
+Docker (``Docker``)
```
wathen, line 24: We should probably put the docker stuff into the [install instructions](https://fitbenchmarking.readthedocs.io/en/latest/users/install_instructions/index.html) as well.

travis/unit_tests_default.sh
```diff
@@ -1,10 +1,16 @@
 #!/bin/bash
 # Test default installation
-pytest fitbenchmarking/cli --cov=fitbenchmarking/cli --cov-report term-missing
+if [ -z "$1" ]
```
wathen, line 3: From our discussion this should be reverted back to what it was before as optional arguments are no longer needed

wathen (changes requested):

docs/source/contributors/structure.rst
tyronerees: I was just typing this up, but I don't think the Docker stuff is useful to a user yet.  In particular:
- fitbenchmarking isn't included, so we'd have to explain some gymnastics where the user pulls locally or shares the filesystem, and then does pip install
- the docker containers don't include a browser, so nothing useful is output

I think we should have an issue to make further containers which contain fitbenchmarking + dependencies, but I don't think we should complicate the install instructions quite yet.

tyronerees: Created #614 to address this

wathen (approved):

PR #611, opened 2020-09-09
use pycutest's cache clearing functions

### LINKED ISSUE #610: Handling of PYCUTEST_CACHE
**Description of the error**
The `$PYCUTEST_CACHE` environment variable needs to be set in order for the cutest parser to work.  One of the first things that fitbenchmarking does is empty this folder:

https://github.com/fitbenchmarking/fitbenchmarking/blob/cc5e89c2a63efa7befcd08ef9436706749ae52fe/fitbenchmarking/parsing/cutest_parser.py#L24-L25

This can be dangerous if `$PYCUTEST_CACHE` points to a directory that has shared usage.  

In fact, only the subdirectory `$PYCUTEST_CACHE/pycutest_cache_holder` is accessed by pycutest, and so it is enough (and safer) to empty that directory.

tyronerees:
#### Description of Work

Fixes #610 

#### Testing Instructions

1. Check that SIF problems run correctly

### DIFF
--- a/fitbenchmarking/parsing/cutest_parser.py
+++ b/fitbenchmarking/parsing/cutest_parser.py
@@ -20,10 +20,8 @@ try:
 except ImportError:
     from backports.tempfile import TemporaryDirectory
 
-
-# Empty the cache
-del_contents_of_dir(os.environ["PYCUTEST_CACHE"])
-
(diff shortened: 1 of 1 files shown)


jfowkes (approved):
I can confirm that the SIF problem set runs correctly.

PR #617, opened 2020-09-17
Fixing chi_sq and runtime NaN issue

### LINKED ISSUE #616: Performance profiler error
As noticed by @jfowkes default settings with the `SIF` problem set produces an error in the performance profiler, this is caused by the calculated `chi_sq` being `nan`.

wathen:
#### Description of Work

Fixes #616 

Adds checks that the computed values for `chi_sq` and `runtime` are not NaN. 

#### Testing Instructions

1. Check default options with SIF and NIST problem sets


tyronerees (approved):
All looks good, and fixes the issue

PR #623, opened 2020-09-22
Restructuring external build scripts

### LINKED ISSUE #621: move install examples
The installation scripts should be moved to /build/<software>/, in line with the updated docs

wathen:
#### Description of Work

Fixes #621 

Removes old `mantid` build system and makes it consistent with other external packages

#### Testing Instructions

1. Test the build scripts work


tyronerees (changes requested):
Most of these scripts require setting environment variables.  We should write somewhere (ideally in a comment in the scripts and in the docs) that they should either be `source`d, typed into the terminal, or the environment variables must be set elsewhere.

*GSL*

The GSL build should have `python -m pip install pygsl` added to it, so that it's self contained, like the others.

*CUTEST*

Installing CUTEST gives the instruction to add the CUTEST environment variable to your `.bashrc` file, but doesn't give any instructions about PYCUTEST_CACHE or updating PYTHONPATH.  We should probably add these to the install script, as we have in the Mantid script.

Change line 16 to 
`export MASTSIF=$SCRIPTPATH/../examples/benchmark_problems/SIF/`

*RALFIT*
It's probably worth adding cmake to the apt-get installs at the start -- it's bundled with travis, but needed on the system to build ralfit

Add `export LD_LIBRARY_PATH=${SCRIPTPATH}/RALFit/libRALFit/build/src:$LD_LIBRARY_PATH` (and print a message to add this to .bashrc for consistency with the others.)

tyronerees (changes requested):
One change suggested in code, and it just needs an echo mentioning PYCUTEST_CACHE and PYTHONPATH (and the other CUTEST environment variables, if you think it needs it) at the end of `build/cutest.sh`

build/ralfit.sh
```diff
@@ -19,4 +19,6 @@ cd $SCRIPTPATH
 
 export LD_LIBRARY_PATH=${SCRIPTPATH}/RALFit/libRALFit/build/src:$LD_LIBRARY_PATH
 
-echo "To enable RALFit in future sessions we recommend adding 'export LD_LIBRARY_PATH=${SCRIPTPATH}/RALFit/libRALFit/build/src:$LD_LIBRARY_PATH'"
+printf "To enable RALFit in future sessions we recommend adding \n
+'export LD_LIBRARY_PATH=${SCRIPTPATH}/RALFit/libRALFit/build/src:$LD_LIBRARY_PATH' \n
```
tyronerees, line 23:
```suggestion
'export LD_LIBRARY_PATH=${SCRIPTPATH}/RALFit/libRALFit/build/src:\$LD_LIBRARY_PATH' \n
```

tyronerees (changes requested):
Added a backslash, but then it's good to go

tyronerees (approved):

PR #638, opened 2020-10-02
Make the links relative

### LINKED ISSUE #615: Make the css links relative
At the moment the css files are linked to using an absolute file path.  We should change this to a relative path, which will allow easy deployment of the result on the web, as well as viewing the results generated in a Docker container on a native filesystem.

Probably the cleanest way to do this would be to copy the css files into a `css` subdirectory of the `results` directory, and change paths to be to that relative to root, e.g., `../../css/main_style.css`, not `/home/fitbenchmarking/fitbenchmarking/templates/main_style.css`.  This would allow the `results` directory to be moved around without breaking the output.

tyronerees:
#### Description of Work

Fixes #615 

I've also made performance profile links relative, and added a sanitiser for the minimizer names (":" and " " were breaking things for me on Windows)

#### Testing Instructions

1. run fitbenchmarking and check all outputs render correctly
2.
3.


fitbenchmarking/cli/main.py
```diff
@@ -138,8 +138,10 @@ def run(problem_sets, options_file='', debug=False):
     root = os.path.dirname(inspect.getfile(fitbenchmarking))
     template_dir = os.path.join(root, 'templates')
     env = Environment(loader=FileSystemLoader(template_dir))
-    style_css = os.path.join(template_dir, 'main_style.css')
-    custom_style = os.path.join(template_dir, 'custom_style.css')
+    local_css_dir = os.path.join(options.results_dir,'css')
```
wathen, line 141: These four lines of code look almost identical each time we create a html page, I suggest this is put into `utils` or something like that to avoid duplication

fitbenchmarking/core/results_output.py
```diff
@@ -36,7 +37,16 @@ def save_results(options, results, group_name,
     :return: Path to directory of group results
     :rtype: str
     """
-    _, group_dir, supp_dir, fig_dir = create_directories(options, group_name)
+    _, group_dir, supp_dir, fig_dir, local_css_dir = \
```
wathen, line 40: Seems that the first argument is not used anywhere, I suggest removing it if this is the case

fitbenchmarking/core/results_output.py
```diff
@@ -36,7 +37,16 @@ def save_results(options, results, group_name,
...
+                                        create_directories(options, group_name)
+
+    # copy the template css files into a subfolder of results
+    root = os.path.dirname(inspect.getfile(fitbenchmarking))
+    template_dir = os.path.join(root, 'templates')
+    local_css_dir = os.path.join(options.results_dir,'css')
+    for css_file in ["main_style","custom_style","math_style","table_style"]:
+        copy2(os.path.join(template_dir,css_file+".css"),local_css_dir)
+    rel_css = os.path.relpath(local_css_dir,supp_dir)
```
wathen, line 49: `rel_css` is not used in this function, please remove

fitbenchmarking/core/results_output.py
```diff
@@ -180,9 +192,11 @@ def create_problem_level_index(options, table_names, group_name,
     root = os.path.dirname(inspect.getfile(fitbenchmarking))
     template_dir = os.path.join(root, 'templates')
     env = Environment(loader=FileSystemLoader(template_dir))
-    style_css = os.path.join(template_dir, 'main_style.css')
-    custom_style = os.path.join(template_dir, 'custom_style.css')
-    maths_style = os.path.join(template_dir, 'math_style.css')
+    local_css_dir = os.path.join(options.results_dir,'css')
```
wathen, line 195: See above comment about repeated code

fitbenchmarking/results_processing/support_page.py
```diff
@@ -79,9 +80,11 @@ def create_prob_group(prob_results, group_name, support_pages_dir,
         root = os.path.dirname(inspect.getfile(fitbenchmarking))
         template_dir = os.path.join(root, "templates")
         env = Environment(loader=FileSystemLoader(template_dir))
-        style_css = os.path.join(template_dir, 'main_style.css')
-        table_css = os.path.join(template_dir, 'table_style.css')
-        custom_style = os.path.join(template_dir, 'custom_style.css')
+        local_css_dir = os.path.join(options.results_dir,'css')
+        rel_css = os.path.relpath(local_css_dir,options.results_dir)
+        style_css = os.path.join(rel_css, 'main_style.css')
```
wathen, line 85: It seems that when I run an example the style sheet are not correctly used in the support pages
Screenshot from 2020-10-05 11-12-15

fitbenchmarking/results_processing/tests/test_support_page.py
```diff
@@ -103,10 +103,15 @@ def test_file_name(self):
                                        support_pages_dir=self.dir.name,
                                        options=self.options)
         file_names = [r.support_page_link for r in self.results]
-        expected = [os.path.join(self.dir.name, f)
+        print("*** file_names ***")
```
wathen, line 106: Unless the printing is needed I suggest this is removed

wathen (changes requested):
Looks good, a few points which I have outlined here

fitbenchmarking/results_processing/tests/test_support_page.py
tyronerees: Good catch -- fixed

fitbenchmarking/core/results_output.py
tyronerees: It looks like you're right.  I was wary of taking it out because I didn't know why it was there in the first place, but I don't see any reason for it (it's still built, but it's available as a member of `options`  anyway)

fitbenchmarking/cli/main.py
tyronerees: I can only pull out two -- the  particular css files linked to depend on the page being generated.  Have created a function to do this, which I think makes the code a bit more understandable.

wathen (approved):

PR #661, opened 2020-10-13
Updating default minimizers in options

### LINKED ISSUE #607: Default minimizer selection
As spotted in #589 the default runtime for the default problem set with the default minimizers takes a while to run. Might be worth changing the default minimizers to just `scipy` and `scipy_ls`. This would be inline with #602.

### LINKED ISSUE #656: Tests run by default
If you install via
`pip install fitbenchmarking`
then only scipy minimizers are available.  The default options should reflect this.

wathen:
#### Description of Work

Fixes #656 
Fixes #607

Sets the default minimizers to be `scipy` and `scipy_ls`.

#### Testing Instructions

1. Check tests pass 
2. Check that if you `pip` install FitBenchmarking without any external packages (`pip` installable or otherwise) you can run FitBenchmarking.


tyronerees (approved):
Looks good -- works as expected

PR #712, opened 2021-01-21
Issue #708: Use min arg rather than best

### LINKED ISSUE #708: Plots for jacobian options
When cycling through different jacobian options, the plots produced are not correct.  

I'll update with instructions to reproduce later, but I noticed when working on issue #386 that the final(?) params calculated are used to produce a plot for all minimizers, even if multiple jacobians are used

AndrewLister-STFC:
#### Description of Work
Revrts to using the minimizer arg for generating plots rather than the stored argument.

Fixes #708

#### Testing Instructions

1. Run Fitbenchmarking with plots
2. Check that plots display the correct minimizer
3. Check that plots folder contains a full set of files

### DIFF
--- a/fitbenchmarking/results_processing/plots.py
+++ b/fitbenchmarking/results_processing/plots.py
@@ -180,7 +180,7 @@ class Plot(object):
                        plot_options=plot_options_dict,
                        y=y)
         self.format_plot()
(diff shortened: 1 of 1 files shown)


tyronerees (approved):
Runs as expected.  Linting failure is not caused by this change, so pushing this through.

PR #719, opened 2021-02-19
Fix badges in branch

AndrewLister-STFC:
Touches up badges before submitting build to pypi, so that versions are fixed.

I'm not sure we'll be able to do the same for the badges that show up on the docs though as RTD pulls the tagged commit so I can't intercept it before it goes.


tyronerees (approved):
Looks good

PR #720, opened 2021-02-23
Add script and docs for preparing releases

AndrewLister-STFC:
I've made a script to ensure tagging is done consistently and added docs around the release process.

Please check the docs make sense and feel free to check the script works (although you may have to undo the 2 commits it will make!)


docs/source/contributors/workflow.rst
```diff
@@ -94,3 +94,42 @@ the PR that this being backported separately.
 
 If you encounter problems with cherry picking into release branch please
 don't hesitate to speak to an experienced member of the FitBenchmarking team.
+
+
+================
+Create a release
```
tyronerees, line 100:
```suggestion
Creating a release
```

docs/source/contributors/workflow.rst
```diff
@@ -94,3 +94,42 @@ the PR that this being backported separately.
...
+================
+In order to create a new release for FitBenchmarking, there are a few manual steps.
+These have been streamlined as much as possible.
+
+First checkout the branch to create the release from.
```
tyronerees, line 105:
```suggestion
First checkout the branch to create the release from.  Releases should only be made from a `release-x-x` branch, not a development branch or master.
```

tyronerees: I can't think of a reason why we'd not do this -- can you?  We should probably enforce this in the script.

docs/source/contributors/workflow.rst
```diff
@@ -94,3 +94,42 @@ the PR that this being backported separately.
...
+
+From the root of the repo run the "ci/prep_and_tag_release.sh" script with the new version number.
+The version number will be rejected if it is not of the expected form.
+We expect a "v" followed by the major, minor, and patch numbers,
+and an optional numbered label to mark the type of release.
+
+Possible lables are:
```
tyronerees, line 112:
```suggestion
Possible labels are:
```

docs/source/contributors/workflow.rst
```diff
@@ -94,3 +94,42 @@ the PR that this being backported separately.
...
+ - -beta (release for testing)
+ - -rc (release candidate)
+ - -dev (release of a development version)
+
```
tyronerees, line 116: I know we have these three already, but do we need them?  What's the difference between `dev` and `beta`, for example?

docs/source/contributors/workflow.rst
```diff
@@ -94,3 +94,42 @@ the PR that this being backported separately.
...
+and revert the change in a second commit so that the links point back to the latest versions.
+
+These commits will need to be pushed to github.
+
+Finally, you will need to create a release on github.
+This can be done by navigating to the releases page, selecting new release
+and typing in the tag that was given to the release
+(it should tell you the tag exists at this point!).
+
+E.g For a first beta version of release 0.1.0. One would run:
```
tyronerees, line 127:
```suggestion
For example, For a first beta version of release 0.1.0, one would run:
```

tyronerees (changes requested):
I can't think of a reason why a release wouldn't be made from the branch `release-??`, so I think it's probably worth checking for that in the script.  It's very easy to run stuff from a feature branch by accident, which may cause a few headaches.

Otherwise looks good to me -- I've just made a few minor suggestions on the wordings in the docs.

docs/source/contributors/workflow.rst
AndrewLister-STFC: I agree, might as well check.

docs/source/contributors/workflow.rst
AndrewLister-STFC: I was just listing what we already had... Maybe we'd want a `dev` if someone requested a bleeding edge feature and we weren't near a `beta`? We can probably remove this until we need it.

docs/source/contributors/workflow.rst
```diff
@@ -94,3 +94,42 @@ the PR that this being backported separately.
...
+
+From the root of the repo run the "ci/prep_and_tag_release.sh" script with the new version number.
+The version number will be rejected if it is not of the expected form.
+We expect a "v" followed by the major, minor, and patch numbers,
+and an optional numbered label to mark the type of release.
+
+Possible labels are:
+ - -beta (release for testing)
+ - -rc (release candidate)
+ - -dev (release of a development version)
```
AndrewLister-STFC, line 115:
```suggestion
```

tyronerees (approved):
all looks good

Issue #690, opened 2020-12-08
Support iminuit 2.0


AndrewLister-STFC:
**Is your feature request related to a problem? Please describe.**
iminuit recently released version 2.0.
This introduced breaking changes when moving from 1.x to 2.x, documented [here](https://iminuit.readthedocs.io/en/stable/changelog.html).

**Describe the solution you'd like**
We should move to iminuit 2.0, and raise an exception when the wrong iminuit is installed.

**Describe alternatives you've considered**
We could support both, although this seems like it will be extra overhead for little gain.

PR #724, opened 2021-03-02
690 support iminuit 2.0

jess-farmer:
#### Description of Work

Modify minuit_controller to add support for the iminuit 2.x series, see [here](https://iminuit.readthedocs.io/en/stable/changelog.html) for more details on changes between 1.x and 2.x.

Raise an exception when wrong version of iminuit (<2) is installed

Fixes [#690](https://github.com/fitbenchmarking/fitbenchmarking/issues/690)

#### Testing Instructions

1. Try running benchmark problems with minuit minimizer with both iminuit 1.x and iminuit 2.x


fitbenchmarking/controllers/minuit_controller.py
```diff
@@ -5,6 +5,7 @@
 http://iminuit.readthedocs.org
 """
 from iminuit import Minuit
+import iminuit
```
AndrewLister-STFC, line 8: I think we should only use 1 import style, we should either import just `Minuit` and `__version__` as iminuit_version or something sensible, or import iminuit and use `iminuit.Minuit`

fitbenchmarking/controllers/minuit_controller.py
```diff
@@ -24,6 +25,11 @@ def __init__(self, cost_func):
                 :class:`~fitbenchmarking.cost_func.base_cost_func.CostFunc`
 
         """
+
+        if int(iminuit.__version__[:1]) < 2:
+            raise Exception("iminuit version " + iminuit.__version__ +
```
AndrewLister-STFC, line 30: Where possible I'd advise against using the raw `Exception`, we have some more specific ones in `utils/exceptions.py` and if none of those fit we should make a new one.

AndrewLister-STFC (changes requested):
Looks good to me, just 2 minor points, tests should be running (bar the regression tests) now too!

AndrewLister-STFC (approved):
Looks good to me.

PR #725, opened 2021-03-02
Add cmake to docker for iminuit2.x

AndrewLister-STFC:
#### Description of Work

For the new iminuit, we need cmake in the basic docker


AndrewLister-STFC:
I've pushed the change to dockerhub already

Issue #680, opened 2020-11-05
Behaviour when incompatible options are selected


tyronerees:
When the user has chosen an option that isn't available or possible, the default behaviour should be to ignore that option, and carry on.   

At the moment, running with an ini file that uses analytic jacobians on, say: 
`fitbenchmarking -o master_chi_squared.ini -p /NIST/average_difficulty /Neutron /SIF`
 will run on NIST, then stop with an error (and not try SIF).  I think we should ignore if that option is not available, with a warning, not an error, and continue with any available options.

This would also be a problem if we, e.g., choose 'general' minimizers are selected in 'algorithm_type', but the minimizer is chosen from 'scipy_ls'  (as 'scipy_ls' has no 'general' minimizers).

To fix this we should catch any unsupported combinations and print a warning saying that this will not be run due to conflicting requirements in the options file.

**Conflicting options**
- 'algorithm_type' can conflict with 'minimizers'
- 'jac_method' can conflict with the problem type (e.g., Mantid/bumps minimizers don't have analytic jacobians yet)
Future cases where this will be a problem are:
- 'cost_function' can conflict with 'minimizers' (e.g., hellinger conflicts with bumps, all ls minimizers conflict with Poisson)
- 'cost_function' can conflict with tables (Local min tables do not work with Poisson)

PR #735, opened 2021-03-11
680 Behaviour when incompatible options selected

jess-farmer:
#### Description of Work

Moves check for empty results outside of benchmark function, so that if all minimizers for a problem set fail then it can be ignored rather than causing an error.

If minimizer conflicts with analytic jacobian option then instead of an error, a warning is produced and option is ignored.

A function has been added to the base cost function classes to check that the selected minimizer is compatible with the selected cost function e.g if algorithm_type all has been selected with poisson cost function then the function will identify any ls minimizers that will then be ignored.

Finally when results are processed, if both the poisson cost function and local_min table have been selected, then the local min table option is ignored to avoid an error

Fixes [#680](https://github.com/fitbenchmarking/fitbenchmarking/issues/680)

#### Testing Instructions

1. Try running various options/problems to see if above changes work as intended
2. Do warning messages make sense and are they helpful to the user?


fitbenchmarking/results_processing/tables.py
```diff
@@ -64,62 +67,68 @@ def create_results_tables(options, results, best_results, group_name,
...
-                                    has_pp=has_pp,
-                                    pp_filenames=pp_filenames,
-                                    table=html_table,
-                                    error_message=ERROR_OPTIONS,
-                                    failed_problems=failed_problems,
-                                    unselected_minimzers=unselected_minimzers,
-                                    algorithm_type=options.algorithm_type,
-                                    report_failed_min=report_failed_min))
+
+            if suffix == 'local_min' and options.cost_func_type == 'poisson':
```
AndrewLister-STFC, line 71: This is a bit too specific. If we add a new cost function which is not a least squares one in the future, then we'll have to extend this with each new cost function.

I think this should be handled in the class that creates the table, either by exceptions or a validate function.

E.g. the local_min table class raises an exception if it's called with results from a cost function which is not a subclass of BaseNLLSCostFunc (`not isinstance(..., BaseNLLSCostFunc)`). We could then catch the exception in a try except around `generate_table` with a slightly more generic warning and `continue` to skip to the next table type.

AndrewLister-STFC (changes requested):
I like the method for checking minimizers are acceptable! It should make it much more user friendly, and opens the door for allowing multiple types to be passed in the future (the types feel more like a tag than a premade set now). I'm not sure about the extensibility of the method for handling table/cost_function clashes though.

I haven't had chance to run it yet, so I'll do that tomorrow afternoon. I think one of my comments about "self.invalid_algorithm_types" being repeated in subclasses may be relevent for things that aren't part of this pr... so I want to see if @tyronerees knows whether this was included for reasons I don't know (e.g. the docs need it)?

jess-farmer:
@AndrewLister-STFC thanks for looking through this and for your comments - I have now made the suggested changes and also changed the method for handling table/cost_function clashes

AndrewLister-STFC (approved):
Looks good!

Issue #736, opened 2021-03-16
Add support for bounds for SciPy


jess-farmer:
**Is your feature request related to a problem? Please describe.**
See [#433](https://github.com/fitbenchmarking/fitbenchmarking/issues/433)

**Describe the solution you'd like**
Add support for SciPy fitting software and add tests to check that parameter bounds are handled correctly.

PR #739, opened 2021-03-24
736 scipy support for bounds

jess-farmer:
#### Description of Work

Fixes [#736]

Adds support for bounds to SciPy and SciPy LS controllers. Not all SciPy minimizers accept bounds, so in this case when a problem with parameter_ranges is used, the problem will not be run and results will appear as inf in tables (with an error flag explaining that the solver doesn't support bounded problems).

Tests have also been added to check that for those minimizers that accept bounds, they are being respected for bounded problems.

#### Testing Instructions

1. Run Fitbenchmarking with sasview examples (these problems have parameter bounds) with scipy and scipy_ls fitting software selected - check that minimizers that don't support bounds give inf results


tyronerees (changes requested):
Looks good -- as far as I can tell this doesn't affect regular usage, and adds the required functionality for SASView problems.  Many thanks!

`Powell` doesn't seem to respect the bounds -- I asked for `radius` to be between 40 and 50, and got this:
image
The docs https://docs.scipy.org/doc/scipy/reference/optimize.minimize-powell.html#optimize-minimize-powell explicitly say that the solution is not guaranteed to be in the bounds.  We might want to think about a post-check in FitBenchmarking to flag these cases (but that's something that shouldn't be addressed in this PR)

This also doesn't seem to work for Mantid problems (at least, I tried the two obvious incantations on the dictionary and neither worked) -- it's probably worth mentioning that in the docs and/or explicitly bypassing the logic for Mantid problems so that all minimzers are still run.  I think this needs a bit of fiddling, as we pass in two functions that may have the same parameter name, but should be straightforward.  Again, work for another PR.  

Could you also add unit tests which hit the exception lines: 
https://coveralls.io/builds/38246934/source?filename=fitbenchmarking%2Fcore%2Ffitting_benchmarking.py#L287
and the dummy results lines:
https://coveralls.io/builds/38246934/source?filename=fitbenchmarking%2Fcli%2Fmain.py#L149
(or give a reason why these can't be hit)

docs/source/users/problem_definition_files/native.rst
```diff
@@ -11,12 +11,13 @@ In this format, data is separated from the function. This allows running the
...
 
-An example of a native problem is:
+Examples of native problems are:
 
 .. literalinclude:: ../../../../examples/benchmark_problems/Muon/Muon_HIFI_113856.txt
 
+.. literalinclude:: ../../../../examples/benchamrk_problems/SAS_modelling/1D/prob_def_1.txt
 
-This example shows the basic structure in which the file starts with a comment
+These examples shows the basic structure in which the file starts with a comment
```
tyronerees, line 20:
```suggestion
These examples show the basic structure in which the file starts with a comment
```

tyronerees (changes requested):
This looks great -- many thanks!  Just one tiny doc change needed.

tyronerees (approved):

Issue #740, opened 2021-03-24
Add support for bounds for more software


jess-farmer:
**Is your feature request related to a problem? Please describe.**
See [#433](https://github.com/fitbenchmarking/fitbenchmarking/issues/433)

**Describe the solution you'd like**
Add support for minuit, dfo and ralfit fitting software and add tests to check that parameter bounds are handled correctly.

PR #743, opened 2021-03-29
Add support for bounds for minuit, ralfit and dfo

jess-farmer:
#### Description of Work

Fixes [#740](https://github.com/fitbenchmarking/fitbenchmarking/issues/740)

Add upper and lower bounds to each controller and add tests to check bounds are respected. 

In the dfo controller, the argument `rhobeg` is also checked. This is because when specifying bounds, the gap between a lower and upper bound must be at least `2*rhobeg` -if this condition is not satisfied then the value of `rhobeg` is changed to the maximum possible value given the bounds for the problem.

#### Testing Instructions

1. Run Fitbenchmarking with minuit, ralfit and dfo fitting software
2. For problems with no parameter bounds, check that results are consistent
3. For SASView problems, check that bounds are respected

### DIFF
--- a/docs/source/users/problem_definition_files/native.rst
+++ b/docs/source/users/problem_definition_files/native.rst
@@ -95,5 +95,5 @@ parameter_ranges
   Similarly to `fit_ranges`, it takes the form where the first number
   is the minimum in the range and the second is the maximum.
 
-  Currently in Fitbenchmarking, problems with `parameter_ranges` can only
-  be handled by SciPy fitting software.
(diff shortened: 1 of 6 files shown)


tyronerees (approved):
Looks really good -- happy to approve!

I'm going to create an issue to pull more of the `value_ranges` legwork into the problem class. I think we're getting a clearer picture of what is common to all solvers now.  I'll put some thoughts down, but please edit/amend -- you know this stuff best!

Issue #741, opened 2021-03-26
Extend support for bounds to Mantid Problems


jess-farmer:
**Is your feature request related to a problem? Please describe.**
Currently only SASView bounded problems are supported, this support should be extended to Mantid problems. 

**Describe the solution you'd like**
Allow for the parameter_range dictionary to be read in and handled correctly for Mantid problems.

PR #747, opened 2021-03-30
bounds support for mantid problems

jess-farmer:
#### Description of Work

Fixes [#741](https://github.com/fitbenchmarking/fitbenchmarking/issues/741)

Changes `self._param_names` within controllers to all lower case as `parameter_ranges` names are also set to be lower case. This means that `parameter_ranges` can be handled correctly for problems with upper case parameter names e.g. `'A1'`

#### Testing Instructions

1. Run Fitbenchmarking with a Mantid problem that has `parameter_ranges` set
2. Check in results that bounds are being respected

### DIFF
--- a/docs/source/users/problem_definition_files/native.rst
+++ b/docs/source/users/problem_definition_files/native.rst
@@ -89,8 +89,7 @@ fit_ranges
 
 parameter_ranges
   An optional setting which specifies upper and lower bounds for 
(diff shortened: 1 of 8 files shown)


jess-farmer:
This pull request is no longer needed as it is addressed by changes in [#748](https://github.com/fitbenchmarking/fitbenchmarking/pull/748)

Issue #745, opened 2021-03-30
Pull more of the bounds info into the problem class


tyronerees:
The is currently a fair amount of overlap in the code to support bounds in each minimizer
**DFO**:
```python
        value_ranges_lb = np.array([])
        value_ranges_ub = np.array([])
        bound_range = np.array([])
        for name in self._param_names:
            if self.problem.value_ranges is not None \
                    and name in self.problem.value_ranges:
                value_ranges_lb = np.append(value_ranges_lb,
                                            self.problem.value_ranges[name][0])
                value_ranges_ub = np.append(value_ranges_ub,
                                            self.problem.value_ranges[name][1])
                bound_range = np.append(bound_range,
                                        self.problem.value_ranges[name][1]
                                        - self.problem.value_ranges[name][0])
            else:
                value_ranges_lb = np.append(value_ranges_lb, -10e+20)
                value_ranges_ub = np.append(value_ranges_ub, 10e+20)
        self.value_ranges = (value_ranges_lb, value_ranges_ub)
```
**Minuit**:
```python
        self.value_ranges = []
        for name in self._param_names:
            if self.problem.value_ranges is not None \
                    and name in self.problem.value_ranges:
                self.value_ranges.append(
                    (self.problem.value_ranges[name][0],
                     self.problem.value_ranges[name][1]))
            else:
                self.value_ranges.append((None, None))

        self._minuit_problem.limits = self.value_ranges
```
**RALFit**:
```python
        value_ranges_lb = []
        value_ranges_ub = []
        for name in self._param_names:
            if self.problem.value_ranges is not None \
                    and name in self.problem.value_ranges:
                value_ranges_lb.extend([self.problem.value_ranges[name][0]])
                value_ranges_ub.extend([self.problem.value_ranges[name][1]])
            else:
                value_ranges_lb.extend([-np.inf])
                value_ranges_ub.extend([np.inf])
        self.value_ranges = (value_ranges_lb, value_ranges_ub)
```
**Scipy**:
```python
        self.value_ranges = []
        for name in self._param_names:
            if self.problem.value_ranges is not None \
                    and name in self.problem.value_ranges:
                self.value_ranges.append(
                    (self.problem.value_ranges[name][0],
                     self.problem.value_ranges[name][1]))
            else:
                self.value_ranges.append((-np.inf, np.inf))
```
**Scipy_ls**:
```python
        value_ranges_lb = []
        value_ranges_ub = []
        for name in self._param_names:
[...]

PR #748, opened 2021-03-31
745 Pull more of the bounds info into the problem class 

jess-farmer:
#### Description of Work

Fixes [#745](https://github.com/fitbenchmarking/fitbenchmarking/issues/745)

#### Testing Instructions

1. Run Fitbenchmarking with a variety of options are problems and check that results are consistent


AndrewLister-STFC (changes requested):
I only had a few comments on minor things like casting to a list/tuple when it's not needed.

It's unlikely to matter for performance here as the lists are quite small, but if you're iterating through things, casting o a list is just an added overhead.

AndrewLister-STFC (approved):
This all looks good to me, thanks for updating it.

PR #750, opened 2021-04-07
Add Mantid CrystalField example

### LINKED ISSUE #721: Add Mantid CrystalFieldSpectrum Objective
It would be great to have an interface to Mantid's CrystalFieldSpectrum objective function, the API call to this is as follows:

```python
# Load crystal field data
data_ws1 = mt.Load('NdOs2Al10_5K35meV_Ecut_0to3ang_bp15V1.xye')

# Setup crystal field model
cf = mt.CrystalFieldSpectrum(Ion='Nd', Symmetry='C2v', Temperature=5., FWHM=1., B20=0.19, B22=0.11,
                             B40=-0.0004, B42=-0.002, B44=-0.012, B60=5.e-5, B62=0.00054, B64=-0.0006, 
                             B66=0.0008, PeakShape='Lorentzian', IntensityScaling=2.)

# Fix molecular and external magnetic fields
for magfieldPars in ['BmolX', 'BmolY', 'BmolZ', 'BextX', 'BextY', 'BextZ']:
    cf.function.tie(magfieldPars, '0')

# Fix B66 (possibly a bug)
cf.function.tie('B66','0.0008')
    
# Run fit
fit = mt.Fit(cf, InputWorkspace=data_ws1, Minimizer='Levenberg-Marquardt', MaxIterations=100, Output='fit')
```

A simple working example with dataset can be found here:
https://github.com/ralna/CrystalFieldOpt/blob/main/CrystalFieldExample.ipynb

jfowkes:
#### Description of Work

Fixes #721 by adding a crystal field test set

#### Testing Instructions

1. Run fitbenchmarking on the new test set


jfowkes:
Currently creating the plots fails with the following error (no idea what's going wrong here):
```
[12:47:22]  INFO main.py: Producing output for the CrystalField problem set
[12:47:23]  ERROR exception_handler.py: Unknown exception. Exiting.
[12:47:23]  ERROR exception_handler.py: Python argument types in
    IFunction.setParameter(IFunction, int, str)
did not match C++ signature:
    setParameter(Mantid::API::IFunction {lvalue} self, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > name, double value)
    setParameter(Mantid::API::IFunction {lvalue} self, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > name, double value, bool explicitlySet)
    setParameter(Mantid::API::IFunction {lvalue} self, unsigned long i, double value)
    setParameter(Mantid::API::IFunction {lvalue} self, unsigned long i, double value, bool explicitlySet)
```

AndrewLister-STFC:
When we create the mantid function, we set the ties, I need to look more into what's happening with the Attribute Values before I make ay suggestions there though.

jfowkes:
I think the problem we're going to run into is that there's no way to tell what's an AttributeValue and what's a Parameter, as far as I can tell Mantid doesn't do this, it just passes the argument string along to the name function in that string.

AndrewLister-STFC:
Running some tests on terminal (and using tab completion to see what methods are available) I found the following which might be helpful!

```
from mantid import simpleapi
f = simpleapi.FunctionFactory.createInitialized('name=CrystalFieldSpectrum, Ion=Nd, Symmetry=C2v, Temperature=5., FWHM=1., PeakShape=Lorentzian, IntensityScaling=2., B20=0.19, B22=0.11, B40=-0.0004, B42=-0.002, B44=-0.012, B60=0.00005, B62=0.00054, B64=-0.0006, B66=0.0008, ties = (BmolX=0, BmolY=0, BmolZ=0, BextX=0, BextY=0, BextZ=0, B66=0.0008)')
attrs = [s for s in f.attributeNames()]
params = [(f.getParamName(i), f.getParamValue(i)) for i in range(f.nParams())]
```

jfowkes:
That looks really useful, great detective work there!

AndrewLister-STFC:
We can also pull ties out of this with `f.getTies()` all that remains is checking if the ties are respected by other solvers (I don't think they were last time I checked... but that was a while ago so I may have forgotten what I was trying!)

Note: getTies returns all ties in the none trivial sense... it seems to convert the ties in this case to fixed which is shown via the isFixed method

jfowkes:
I have now added proper attribute support. @AndrewLister-STFC could you take a look at this before I move on to implementing support for ties (aka fixed parameters) in non-mantid solvers? 

Also why do the linting tests fail due to stupid pointless things like not having two spaces before inline comments?

jfowkes:
I have now added support for tied parameters.

I'm wondering whether we actually need to filter out the attributes when creating the mantid function... probably not.

Also should probably add some unit tests for this stuff, but how do you test that Mantid has parsed correctly?

jfowkes:
This is now ready to go in: it turns out that all I had to do is add a dedicated `get_mantid_starting_values` function that filters out reserved keywords, attributes (as above) and tied parameters from the function string and passes them to the starting values  , Mantid takes care of everything else. And of course none of this is documented anywhere in Mantid.

fitbenchmarking/parsing/fitbenchmark_parser.py
```diff
@@ -309,6 +312,48 @@ def _get_starting_values(self):
...
+            for key in ignore:  # filter reserved keywords from params
+                if key in params:
+                    params.pop(key)
+            for attr in attr_names:  # filter attributes from params
+                if attr in params:
+                    params.pop(attr)
+            if 'ties' in f:  # filter tied params from params
+                for tie in f['ties']:
+                    if tie in params:
+                        params.pop(tie)
```
AndrewLister-STFC, line 349: I'd be tempted to invert this and only take the params that are required using something like:
```
params = [(f.getParamName(i), f.getParamValue(i))
          for i in range(f.nParams())
          if f.isFixed(i)]
```

Although maybe that doesn't handle the more complicated ties properly.

fitbenchmarking/parsing/fitbenchmark_parser.py
```diff
@@ -309,6 +312,48 @@ def _get_starting_values(self):
...
+            starting_values += [(name_template.format(i, name), val)
+                                for name, val in params.items()]
```
AndrewLister-STFC, line 351: I'm a bit worried that this creates a badly ordered parameter list in the mantid -> non-mantid case.

i.e. `function=some_func,p0=1.0,p1=2.0,p2=4.0,ties=(p1=2.0)`
would lead to `starting_values == {'p0':1.0, 'p2':4.0}`
which then leads to `initial_params == [1.0, 4.0]` in the controllers...

I think this is covered in the [docs](https://fitbenchmarking.readthedocs.io/en/latest/users/problem_definition_files/native.html) under "Note: Any non-standard arguments (e.g. ties, constraints, fixes, …) will only work with Mantid fitting software. Using other minimizers to fit these problems will result in the non-standard arguments being ignored." which we should change if we're supporting it now.

Looking into this a bit more I'm concerned that if a user passes in:
`function=some_func,p0=1.0,p2=4.0,p1=2.0`
we'll get `initial_params == [1.0, 4.0, 2.0]` in the controller which could come out wrong... so maybe it's already broken 😅

AndrewLister-STFC (commented):
I expect we'll want to change the function creation at some point as well as the starting values, although I think a lot  of the code in the fitbenchmark_parsers could be refactored now that the options are diverging more, so that can be done there!

I've got a couple of comments, but nothing major.

fitbenchmarking/parsing/fitbenchmark_parser.py
```diff
@@ -309,6 +312,48 @@ def _get_starting_values(self):
...
+        name_template = '{1}' if len(self._parsed_func) == 1 else 'f{0}_{1}'
+        starting_values = []
+        for i, f in enumerate(self._parsed_func):
+
+            # use mantid function factory to seperate attributes and parameters
+            fcopy = f.copy()
+            if 'ties' in fcopy:
+                fcopy.pop('ties')
+            fstring = ", ".join(f"{key}={value}"
+                                for key, value in fcopy.items())
```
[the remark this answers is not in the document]
jfowkes: Yes it is, but for each function. Will change this to split `self._entries['function']` on semicolons like in the function parser.

fitbenchmarking/parsing/fitbenchmark_parser.py
jfowkes: Agreed, that should work in exactly the same way with much less code!

fitbenchmarking/parsing/fitbenchmark_parser.py
jfowkes: Yes looks like it's already broken, we should create a separate bug for this.

fitbenchmarking/parsing/fitbenchmark_parser.py
jfowkes: The reason it's done this way is that there are quite a few 'hidden' parameters in a lot of Mantid functions that the user has no interest in, some are even set to internal mantid vectors so best not to expose these.

jfowkes:
Oh dear, looks like I broke the results page display. @AndrewLister-STFC any idea where this is coming from?

Screenshot 2021-04-14 at 2 26 22 pm

fitbenchmarking/parsing/fitbenchmark_parser.py
AndrewLister-STFC: Ahh in that case this is fine!

jfowkes: I take it back, I misunderstood how Mantid does things, your suggestion is the correct approach here as all non-fixed parameters that are optimised over should be exposed, will add this in.

AndrewLister-STFC:
@jfowkes Does this do what you need now? Testing with scipy_ls and mantid gives parameters and graphs that look like they coud be correct although mantid is struggling to solve it, do you expect this?

If it is better we should also double check all of the other combinations like:
Non-mantid problem -> mantid solver (I'd expect this still to work)
Mantid multisolve (This could easily be broken in the controller by my commit...)

jfowkes:
Thank you very much, this does everything I hoped for and more!

Mantid multifit breakage aside, this works perfectly :smiley:

AndrewLister-STFC:
```
{'Cost function value': 480.7293657843396,
 'f0.f0.A0': 0.0,
 'f0.f0.A1': 0.0,
 'f0.f1.A': 0.2,
 'f0.f1.Frequency': 1.0,
 'f0.f1.Phi': 0.0,
 'f0.f1.Sigma': 0.2,
 'f1.f0.A0': 0.0,
 'f1.f0.A1': 0.0,
 'f1.f1.A': 0.2,
 'f1.f1.Frequency': 1.0,
 'f1.f1.Phi': 0.0,
 'f1.f1.Sigma': 0.2}
['f0.A0', 'f0.A1', 'f1.A', 'f1.Sigma', 'f1.Frequency', 'f1.Phi']
'f0.A0'
```
It looks like it just adds an extra f{idx} to the start of the name for multifit, so we should be able to just iterate through them

fitbenchmarking/parsing/fitbenchmark_parser.py
```diff
@@ -299,9 +302,8 @@ def _get_starting_values(self):
         :return: Starting values for the function
         :rtype: list of OrderedDict
         """
-        # Mantid functions can have reserved keywords so ignore these as they
-        # are not parameters.
-        ignore = ['name', 'BinWidth', 'ties', 'Formula', 'constraints']
+        # SasView functions can have reserved keywords so ignore these
+        ignore = ['name']
 
         name_template = '{1}' if len(self._parsed_func) == 1 else 'f{0}_{1}'
```
jess-farmer, line 308: If `_get_starting_values` is now only called for Sasview problems, is `name_template` needed as (from my understanding anyway) the length of `self._parsed_func` should only ever be greater than 1 for multifit problems?

jess-farmer (commented):
When testing the CrystalField example with Mantid, for the Damped GaussNewton minimizer I get the following error (with external output set to debug):

Fit-[Notice] Fit started
********* UNHANDLED EXCEPTION *********
  what(): GSLVector index = 26 is out of range = 26 in GSLVector.get()

Fitbenchmarking then stops running and no output is produced - not sure why this is the case, or if this is an error you've also encountered?

Other than that everything looks good and works as expected!

jfowkes:
Thanks Jess, I can reproduce the Damped GaussNewton minimizer error in Mantid (outside of fitbenchmarking), so yet another example of a broken Mantid minimizer. Will open an issue in Mantid about this, their examples should work.

fitbenchmarking/parsing/fitbenchmark_parser.py
jfowkes: Good catch, looks like we can just use `name_template='{1}'` here?

jess-farmer: Looking at how `starting_values` is set (lines 309-312), I think we can remove `name_template` altogether by replacing lines 309-312 with the following:
```
starting_values = [
            OrderedDict([(name, val)
                         for name, val in self._parsed_func[0].items()
                         if name not in ignore])]
```

fitbenchmarking/controllers/mantid_controller.py
```diff
@@ -2,6 +2,7 @@
 Implements a controller for the Mantid fitting software.
 """
 
+from numpy.core.fromnumeric import sort
```
AndrewLister-STFC, line 5: @jfowkes I think I saw that this is unused. Was it used for something that was changed to function a different way?
```suggestion
```

jfowkes: Yeah I saw that too, that was already there, nothing to do with me 🙂

jess-farmer (approved):

jess-farmer:
Thanks for making these changes, all looks good to me!

jess-farmer (changes requested):
Looking further into the exception from bumps in the regression tests (for EMU 73763), I think the cause of this should be looked into further before this PR is merged. 

Running fitbenchmarking on this branch with the Muon problems and bumps fitting software selected, all minimizers raise an exception - however when running fitbenchmarking with the same options on master, results are produced and output as expected. I'll do a bit more investigating this afternoon to see if I can find what's happening here.

AndrewLister-STFC:
So this is a Mantid problem - bumps solver issue. It's strange that it's not showing for other minimizers too since they should all get the same callable.

One change from the problem side of things here is that we now fit all non fixed parameters not just the ones specified but EMU specifies a value for all parameters and has no ties.

I wonder if the mantid style parameter names are confusing the bumps controller somehow

fitbenchmarking/systests/linux_expected_results/all_parsers.txt
```diff
@@ -4,5 +4,5 @@ BENNETT5            1.608e-05 (1.001)  1.639e-05 (1.021)[1]        0.02183 (1359
 cubic, Start 1  1.119e-14 (2.899e+07)         3.861e-22 (1)     148.9 (3.857e+23)      370.8 (9.604e+23)[2]  6.723e-13 (1.741e+09)  6.267e-05 (1.623e+17)[1]         3.861e-22 (1)
 cubic, Start 2  1.146e-14 (2.969e+07)         3.861e-22 (1)   0.01778 (4.604e+19)  0.0005949 (1.541e+18)[2]  6.926e-18 (1.794e+04)     7.176e-11 (1.859e+11)         3.861e-22 (1)
 cubic-fba       1.146e-14 (2.969e+07)         3.861e-22 (1)   0.01778 (4.604e+19)     3.306e-06 (8.563e+15)   9.768e-18 (2.53e+04)     7.176e-11 (1.859e+11)         3.861e-22 (1)
-EMU 73673                    1046 (1)       1053 (1.006)[1]     2.444e+04 (23.36)           1553 (1.484)[2]               1046 (1)              1055 (1.009)              1046 (1)
-Problem Def 1             0.03175 (1)          inf (inf)[4]  5.26e+05 (1.657e+07)              inf (inf)[4]         0.1119 (3.525)              inf (inf)[4]          inf (inf)[4]
+EMU 73673                inf (inf)[3]       1053 (1.006)[1]     2.444e+04 (23.36)           1553 (1.484)[2]               1046 (1)              1055 (1.009)              1046 (1)
```
AndrewLister-STFC, line 7:
```suggestion
EMU 73673                    1046 (1)       1053 (1.006)[1]     2.444e+04 (23.36)           1553 (1.484)[2]               1046 (1)              1055 (1.009)              1046 (1)
```

jess-farmer (approved):

PR #764, opened 2021-04-19
763 change bounds test

### LINKED ISSUE #763: Change bounds test
Currently within `test_controllers.py`, one of the sasview problems is used to test whether each fitting software respects parameter ranges when they are set. The problem used should be changed to one of the mantid `simple_test` problems, for two reasons: 

- bumps respects bounds for mantid problems but not for sasview problems
- the solutions for `simple_test` problems are known and therefore better to use to check that bounds are respected (can be sure that parameter ranges don't include local minima)

Create a new problem definition file that adds `parameter_ranges` to one of the `simple_test` problems and use this in `test_controllers.py` instead of the sasview problem currently used.

Add in a test for bumps fitting software in the `DefaultControllerBoundsTests` class, as using a mantid problem means it now passes. Add a note to the docs to explain that bumps doesn't necessarily respect parameter bounds for sasview problems.

jess-farmer:
#### Description of Work

Fixes #763

This change adds a note to the documentation that bumps may not respect bounds for sasview problems - I have contacted members of the SANS group for more insight on why this might be the case so more detail can be added to the docs/the problem can be fixed. This will be addressed in a separate PR.

#### Testing Instructions

1. Do all controllers pass tests as expected?


tyronerees:
Should have been fixed in SASView 5.0.4 (Released on April 7th 2021)

AndrewLister-STFC (approved):
The only comment is being fixed in a separate PR.

PR #771, opened 2021-04-21
770 fix bumps parameter ordering

### LINKED ISSUE #770: Bumps changes order of SAS Modelling parameters
**Description of the error**
For the SAS Modelling problems, the results from bumps are being displayed incorrectly in the Fitbenchmarking tables - the values for radius and background are the wrong way round.

e.g. for SAS_modelling prob_def_1, in the bumps controller the order of `self.final_params` causes Fitbenchmarking to output the following fitting results: `radius: -5.50901794e-04, length: 2.86957413e+02, background: 2.01877447e+01, scale:  1.10739293e+00, sld: 4.01660840e+00, sld_solvent:  1.15199600e+00` whereas the parameter values taken from `self._func_wrapper` after fitting has taken place are `radius: 20.187744730999043, length: 286.95741267182336, background: -0.0005509017937177449, scale: 1.1073929329336942, sld: 4.016608403414924, sld_solvent: 1.1519960039685162`. 

The output from `bumpsFit` lists the parameters in a different order to what Fitbenchmarking expects and this is not picked up when processing the results.

**Suggested fix**
Ensure that the parameters values are in the correct order when the the output of the fitting is set as `self.final_params`

jess-farmer:
#### Description of Work

Fixes #770

#### Testing Instructions

1. Run Fitbenchmarking with SASView problems and bumps fitting software to check results are now correct


jfowkes (approved):
Looks great, approved.

AndrewLister-STFC:
@tyronerees Could this fix levmar? (Need issue to link with!)

PR #772, opened 2021-04-22
742 flag minimizers not respecting bounds

### LINKED ISSUE #742: Flag minimizers that do not respect bounds
Currently, `Powell` doesn't seem to respect bounds, e.g. setting `radius` to be between 40 and 50 results in the following:
image
The docs https://docs.scipy.org/doc/scipy/reference/optimize.minimize-powell.html#optimize-minimize-powell explicitly say that the solution is not guaranteed to be in the bounds.

A check should be added to FitBenchmarking to identify cases such as this, where the given solution is outside the parameter bounds, so they can be flagged to the user.

jess-farmer:
#### Description of Work

Fixes #742 

#### Testing Instructions

1. Run fitbenchmarking and check that error flag appears as expected if parameter bounds are not respected (until pull request 771 is merged, using bumps minimizers are a good way to test this)


fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -431,6 +431,12 @@ def loop_over_jacobians(controller, options, grabbed_output):
                         chi_sq = np.inf if not problem.multifit \
                             else [np.inf] * len(controller.data_x)
 
+                    # If bounds have been set, check that they have
+                    # been respected by the minimizer and set error
+                    # flag if not
+                    if controller.problem.value_ranges is not None:
+                        controller.check_bounds_respected()
+
```
tyronerees, line 439: This line is missed by the tests -- can we add a test to cover it?

tyronerees (changes requested):
This looks good.  

Other than the line not hit by the tests, I only have one comment.  The minimizers that don't respect the bounds will usually end up having the smallest residual.  In the table below, I've set '30 < radius < 50' and '325 < length < 375'.  The only minimizer that converges and satisfies these bounds is `levmar:scipy-no-jac`.  Also `dogbox` and `levmar` fail to converge, but both find smaller residuals:
image
I feel the colouring on the table should ignore the results of the minimizers that don't satisfy the bounds -- these should always be darkest red (or potentially another colour to highlight they are different -- darkish blue?).  However, I don't think we should lose the numerical information, as that does tell us something.  

Let me know if this isn't clear and we can chat on zoom

tyronerees:
I should say, I'd also be happy if taking the flagged values out of the table colouring goes in another PR -- this does fix the open issue, and flags it to the user correctly.

jess-farmer:
Thanks for looking through this @tyronerees, I think it definitely makes sense for the colouring on the table to reflect when a minimizer hasn't respected parameter bounds. I'll look at addressing this (as well the line missed by the tests) within this PR

tyronerees (changes requested):
Thanks for this.  I think we still need to iterate slightly on this to make sure it's useful to users. Here's the output I get for a problem that Powell fails for, on which I've highlighted the issues:
image

It's obvious now that Powell shouldn't be the best, which is good, but it's not clear how the others compare, as they are all dark red.  I think it might be best to amend the relative number to 'inf' in this case, (but leave the absolute as is), which will not only push the failing minimizer to red, but make the others yellow too.

I hope this is clear (sorry that I wasn't before) -- let me know if you have any questions.

tyronerees (approved):
This is great -- thanks @jess-farmer!

PR #778, opened 2021-04-27
Add warning message for incorrect parameter names in parameter_ranges

### LINKED ISSUE #777: parameter_ranges with incorrect parameter names are accepted
**Description of the error**
Currently if `parameter_ranges` are set in a problem definition file and incorrect parameter names are used e.g. `parameter_ranges = {'r': (1,50), 'l': (1,500)}` where r and l are used instead of radius and length, fitbenchmarking still runs and produces output as if bounds are set.

As the parameter names in `parameter_ranges` don't correspond to any actual parameter names, I would expect a warning/error message.

Instead, in `fitbenchmark_parser`, the `parameter_ranges` line is parsed and this triggers the `set_value_ranges` function. This function assigns all parameters not listed in `parameter_ranges` lower and upper bounds of `-inf` and `inf` respectively. Because of this, even though every parameter is unbounded, each minimizer is run as if bounds are set. This could be a problem if a user has set bounds and made a typo in a parameter name.

**Suggested fix**
In the function `set_value_ranges` in `fitting_problem.py`, a check should be added to make sure that the parameter names listed in `parameter_ranges` are a sublist of `fitting_problem._param_names`. If this is not the case, then a warning message should be displayed to the user and the problem skipped.

jess-farmer:
#### Description of Work

Fixes #777 

#### Testing Instructions

1. Run problems with incorrect parameter names in `parameter_ranges` to check warning message appears as expected
2. Check that problems will correct `parameter_ranges` still run as expected

### DIFF
--- a/fitbenchmarking/parsing/fitting_problem.py
+++ b/fitbenchmarking/parsing/fitting_problem.py
@@ -12,7 +12,8 @@ except ImportError:
     from itertools import zip_longest as izip_longest
 import numpy as np
 
-from fitbenchmarking.utils.exceptions import FittingProblemError
+from fitbenchmarking.utils.exceptions import FittingProblemError, \
+    IncorrectBoundsError
 
 
 # Using property getters and setters means that the setter does not always use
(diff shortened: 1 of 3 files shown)


AndrewLister-STFC (approved):
Looks good, no comments

PR #795, opened 2021-05-12
Matlab Fitting

### LINKED ISSUE #119: Add matlab fitting
Add Matlab fitting software to FitBenchmarking.

Matlab can be called from Python using the Matlab Engine API for Python.  I find the [official Matlab docs](https://uk.mathworks.com/help/matlab/matlab-engine-for-python.html#:~:text=The%20MATLAB%20Engine%20API%20for%20Python%20provides%20a,of%20Python%20Compatible%20with%20MATLAB%20Products%20by%20Release.) a bit terse, but this seems to be a good tutorial: https://mscipio.github.io/post/matlab-from-python/

Here is a guide to some methods for curve fitting in Matlab:
https://walkingrandomly.com/?p=5196#:~:text=Pure%20MATLAB%20solution%20%28No%20toolboxes%29%20In%20order%20to,not%20be%20suitable%20for%20more%20complex%20fitting%20problems.

I think we should implement, from this article:
- Pure MATLAB solution
- MATLAB with optimization toolbox
- MATLAB with statistics toolbox
- MATLAB with curve fitting toolbox

Each of these should probably be a separate controller (and we should make sure we explore any options within these toolboxes).  Note that these toolboxes need to be installed when matlab is installed -- there'll be an option to pick which toolbox to use (we have a licence for all at STFC).

### LINKED ISSUE #798: Add matlab fitting (pure matlab solution)
Add in fitting software option for pure matlab solution - see #119

jess-farmer:
#### Description of Work

Fixes #798 

Still to do:
- [x] Add in documentation for matlab software 
- [x] Add in tests

#### Testing Instructions

1. Choose Matlab in the options and check results produced
2. Check that documentation makes sense


jess-farmer:
Coverage check currently fails because tests for the Matlab fitting software can currently only be run locally

AndrewLister-STFC (changes requested):
This all looks pretty good, I've made a few small comments and still need to check it out to run the tests, linting and some examples.

AndrewLister-STFC (approved):
All looks good. I tested the regression and controller tests with no failures

PR #799, opened 2021-05-14
fix linting errors and warnings in results

### LINKED ISSUE #788: Fix linting in results
Fix the results subdirectory - related to #702

Bring the errors and warnings down to zero, and address others if appropriate.

jess-farmer:
#### Description of Work

Fixes #788 

#### Testing Instructions

1. Check that there are no pylint or flake8 errors/warnings
2. Check nothing is broken by changes

### DIFF
--- a/fitbenchmarking/results_processing/performance_profiler.py
+++ b/fitbenchmarking/results_processing/performance_profiler.py
@@ -1,14 +1,14 @@
 """
 Set up performance profiles for both accuracy and runtime tables
 """
-from collections import OrderedDict
 import os
+from collections import OrderedDict
 from textwrap import wrap
+import numpy as np
+import matplotlib.pyplot as plt
 
 import matplotlib
 matplotlib.use('Agg')
-import matplotlib.pyplot as plt
-import numpy as np
 
 
 def profile(results, fig_dir):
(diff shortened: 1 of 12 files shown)


fitbenchmarking/results_processing/performance_profiler.py
```diff
@@ -1,14 +1,14 @@
...
-from collections import OrderedDict
 import os
+from collections import OrderedDict
 from textwrap import wrap
+import numpy as np
+import matplotlib.pyplot as plt
 
 import matplotlib
 matplotlib.use('Agg')
-import matplotlib.pyplot as plt
```
AndrewLister-STFC, line 10: I think the matplotlib.pyplot import has to come after the Agg line. I can't remember why -- it should probably have been commented here!

Having looked it up, it appears to be due to needing to save to file as described in [this answer](https://stackoverflow.com/a/20249447). There might also be more info in the git history.

fitbenchmarking/results_processing/plots.py
```diff
@@ -2,13 +2,13 @@
 Higher level functions that are used for plotting the fit plot and a starting
 guess plot.
 """
+import os
+import matplotlib.pyplot as plt
 import matplotlib
 matplotlib.use('Agg')
-import matplotlib.pyplot as plt
```
AndrewLister-STFC, line 7: Same comment as in other file

AndrewLister-STFC (changes requested):
Looks pretty good, I do have 2 comments though - theres a slight issue with the matplot lib import and I get a refactor warning that should be quite quick to fix:
- R1725: Consider using Python 3 style super() without arguments (super-with-arguments)

AndrewLister-STFC (approved):
Looks good! Flake8 has no complaints and pylint has 2 refactors which can be ignored for now if not indefinitely...

PR #801, opened 2021-05-18
Fix linting in parsing

### LINKED ISSUE #787: Fix linting in parsing
Fix the parsing subdirectory - related to #702

Bring the errors and warnings down to zero, and address others if appropriate.

Anders-Markvardsen:
#### Description of Work

Fixes #787

#### Testing Instructions

1. check no W and E when running `flake8` and `pylint` against `fitbenchmarking/parsing`
2. and code changes look reasonable


Anders-Markvardsen:
@AndrewLister-STFC concerning pylint `fitbenchmarking/parsing/fitbenchmark_parser.py:341:8: W0201: Attribute '_mantid_starting_values' defined outside __init__` do you think this variable could be created in init or the one place where it is used in code?

Anders-Markvardsen:
@AndrewLister-STFC or @tyronerees is 

* `from pytest import test_type as TEST_TYPE` needed in `test_parsers.py`?
* also if to 'pip install .' following by 'pytest fitbenchmarking/parsing/tests/test_parsers.py` get error `No module named 'pycutest` is this OK? I.e. are we saying to be able to run test must do xyz extra steps or use docker images? If the latter, do we documnet this?

AndrewLister-STFC:
> 
> 
> @AndrewLister-STFC concerning pylint `fitbenchmarking/parsing/fitbenchmark_parser.py:341:8: W0201: Attribute '_mantid_starting_values' defined outside __init__` do you think this variable could be created in init or the one place where it is used in code?

Good question! The reason it's created where it is is because it's kind of generated as a side effect of making the function and it seems wrong to run this twice. Maybe the best option would be to have a `self._starting_values` defined in the __init__ and changing this to remove the `_mantid` part.

The new IVP PR also uses a similar format of setting a starting vals variable on the parser object.

AndrewLister-STFC:
> 
> 
> @AndrewLister-STFC or @tyronerees is
> 
>     * `from pytest import test_type as TEST_TYPE` needed in `test_parsers.py`?
> 
>     * also if to 'pip install .' following by 'pytest fitbenchmarking/parsing/tests/test_parsers.py`get error`No module named 'pycutest` is this OK? I.e. are we saying to be able to run test must do xyz extra steps or use docker images? If the latter, do we documnet this?

`TEST_TYPE` is used in lines 57-59 to determine which parsers to test. If the issue is along the lines of "Could not find test_type in pytest", I think we have just been putting in a `pylint: disable=...`, e.g. https://github.com/fitbenchmarking/fitbenchmarking/blob/a9e86773105d6f6e16b2ccfe7b8374c1a82de77f/fitbenchmarking/controllers/tests/test_controllers.py#L9

It looks like you don't have cutest installed so can't run the full tests. To run only the default tests I think you need `--test-type default` as an option passed in on your pytest command. Looking at the docs I can't see anywhere that this is written down so we should add it.

Anders-Markvardsen:
> > @AndrewLister-STFC concerning pylint `fitbenchmarking/parsing/fitbenchmark_parser.py:341:8: W0201: Attribute '_mantid_starting_values' defined outside __init__` do you think this variable could be created in init or the one place where it is used in code?
> 
> Good question! The reason it's created where it is is because it's kind of generated as a side effect of making the function and it seems wrong to run this twice. Maybe the best option would be to have a `self._starting_values` defined in the `__init__` and changing this to remove the `_mantid` part.
> 
> The new IVP PR also uses a similar format of setting a starting vals variable on the parser object.
> 
> In the medium term I think we should probably refactor this so that there is a `FitbenchmarkParser` base class and `MantidParser`, `SASViewParser`, and `IVPParser` can all be subclasses which might make this a bit neater.

For now left a pylint disable command. About refactoring then your idea sounds likely good. FitBenchmarkParser, NistParser and CutestParse are already child classes of Parser. I think what you are suggesting is that where this can provide better readability/maintainability and each of these could further also become base classes in their own right and for the different software packages we support. We could bring this up in a future Tuesday + combine with an issue

AndrewLister-STFC (changes requested):
The changes you've made look good, pylint is complaining on my system though. (It might be worth checking that you have version 2.8.2 as Tyrone ran into problems with versioning on pylint earlier).

Pylint complaints:
```
fitbenchmarking/parsing/nist_parser.py:78:12: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
fitbenchmarking/parsing/nist_parser.py:80:12: W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
fitbenchmarking/parsing/nist_parser.py:159:4: R0201: Method could be a function (no-self-use)
fitbenchmarking/parsing/nist_parser.py:188:4: R0201: Method could be a function (no-self-use)
fitbenchmarking/parsing/nist_parser.py:239:4: R0201: Method could be a function (no-self-use)
fitbenchmarking/parsing/nist_parser.py:277:4: R0201: Method could be a function (no-self-use)
fitbenchmarking/parsing/nist_parser.py:344:4: R0201: Method could be a function (no-self-use)
************* Module fitbenchmarking.parsing.fitbenchmark_parser
fitbenchmarking/parsing/fitbenchmark_parser.py:45:8: R1725: Consider using Python 3 style super() without arguments (super-with-arguments)
fitbenchmarking/parsing/fitbenchmark_parser.py:352:0: R1721: Unnecessary use of a comprehension (unnecessary-comprehension)
fitbenchmarking/parsing/fitbenchmark_parser.py:444:12: W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
fitbenchmarking/parsing/fitbenchmark_parser.py:478:12: W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
************* Module fitbenchmarking.parsing.base_parser
fitbenchmarking/parsing/base_parser.py:35:8: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
************* Module fitbenchmarking.parsing.parser_factory
fitbenchmarking/parsing/parser_factory.py:59:16: W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
fitbenchmarking/parsing/parser_factory.py:63:12: W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
************* Module fitbenchmarking.parsing.cutest_parser
fitbenchmarking/parsing/cutest_parser.py:148:15: E1130: bad operand type for unary -: NoneType (invalid-unary-operand-type)
```

Anders-Markvardsen:
@AndrewLister-STFC installed newer version of pylint and fixed new w and e flagged up.

AndrewLister-STFC (approved):
Looks good to me, the only messages I have now are "no-self-use" and "consider-using-with" but neither of these are an issue.

PR #806, opened 2021-05-21
Check pi as a value not a function for nist parser - is_safe method

### LINKED ISSUE #601: fix nist is_safe
**Description of the error**
The function `is_safe`: https://github.com/fitbenchmarking/fitbenchmarking/blob/7c544865b9bd004994a06e8b82357eb28dad6e69/fitbenchmarking/parsing/nist_data_functions.py#L69
strips out `np.<function>` as being safe.  However, `np.pi` is included in this list, and `np.pi` is a constant, not a function.  This can lead to functions like `np.pi * x` being listed as not safe, as they are stripped to the invalid string `*x`, which starts with an operator.  This was the case in a function found in the review of #598.

We should treat `np.pi` differently to functions like `np.cos`, `np.log`, etc -- maybe replacing `np.pi` by `pi`, which can then be treated as a constant?

AndrewLister-STFC:
#### Description of Work

Fixes #601

#### Testing Instructions

1. Run is_safe for "np.pi*x"

### DIFF
--- a/fitbenchmarking/parsing/nist_data_functions.py
+++ b/fitbenchmarking/parsing/nist_data_functions.py
@@ -150,7 +150,7 @@ def is_safe(func_str):
 
     # These are all safe and can be stripped out
     if 'np' in func_str:
-        np_funcs = ['np.exp', 'np.cos', 'np.sin', 'np.tan', 'np.pi', 'np.log']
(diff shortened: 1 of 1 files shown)


jess-farmer (approved):
Approving - works as expected

Issue #776, opened 2021-04-27
Levmar performance


tyronerees:
**Description of the error**
Passing a Jacobian to the levmar minimizer doesn't seem to work robustly.  This needs to be investigated and fixed.

robertapplin:
@tyronerees do you have an options file I could use to reproduce this, and which problem set you used?

tyronerees:
This options file:
```
[MINIMIZERS]
scipy_ls: lm-scipy
bumps: lm-bumps

[FITTING]
num_runs: 1
software: scipy_ls
	  levmar
	  bumps
	  
jac_method: scipy
	    default
	    analytic
```
and running with the NIST/low_difficulty test set (`fitbenchmarking -o levmar-test.ini -p examples/benchmark_problems/NIST/low_difficulty/`) gives:
![image](https://user-images.githubusercontent.com/13645545/136222344-16e4e917-0c2b-43ff-8aa7-ed7121ace641.png)
The results for `levmar: analytic` and `levmar: scipy 2-point` are different from all the others (yet consistent with each other), which suggests something is going wrong with a user-passed Jacobian.  `levmar: default default` is fine, so the function evaluations being sent to levmar seem to be correct.  

The same behaviour is seen for all problem sets.

tyronerees:
Here's a file that calls levmar and scipy for a range of Jacobians for the problem `danwood`, which is in the NIST test set:
[danwood.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/7328910/danwood.zip)

Note that there is no real difference between scipy and levmar's performance

PR #810, opened 2021-05-28
use residual, not function, in levmar

### LINKED ISSUE #804: No reference to cost func choice in Levmar controller
**Description of the error**
Looking at the Levmar Controller, there is no reference to the cost function chosen by the user. I think that the only thing that varies with cost function choice is the evaluation of the jacobian in the `_jeval` function. 

The levmar docs [FAQ page](http://users.ics.forth.gr/~lourakis/levmar/faq.html) also seem to suggest that Levmar only supports NLLS minimization, if this is the case then other cost function choices should result in a `CostFuncError` similar to the one in the Bumps Controller.

tyronerees:
#### Description of Work

Fixes #804 by sending the residual, `eval_r` to levmar, rather than f(x,p) and y.  Since the residual is part of the cost function class, it is correct for the cost function selected.

#### Testing Instructions

1. Run against the three `nlls` cost functions and check default jacobian gives answers consistent with other methods (other jacobians may not, due to #776)
2. Run against the `poisson` cost function, and check that an error is returned


fitbenchmarking/controllers/levmar_controller.py
```diff
@@ -54,9 +55,10 @@ def setup(self):
             lb = [None if x == -np.inf else x for x in lb]
             ub = [None if x == np.inf else x for x in ub]
             self.param_ranges = list(zip(lb, ub))
+        self.lm_y = np.zeros(self.data_y.shape)
 
     # pylint: disable=unused-argument
```
jess-farmer, line 60: This pylint disable warning statement can be removed now as the unused argument, x, has been removed

jess-farmer (changes requested):
Changes look good! The expected results for the regression tests just need updating and my only other comment is that the docstrings for `_feval` and `_jeval` can be updated to reflect the fact that the `x` argument has been removed

jess-farmer (approved):

PR #816, opened 2021-05-28
Adding synthetic SASmodel problem sets

BenPetersRAL:
#### Description of Work

I have compiled four SASview synthetic problem sets and added them to the benchmark_examples:

1. simple shapes
2. moderate shapes
3. complex shapes
4. shape independent

The sets are based on an appropriate selection on sasmoel shapes from the sasview library. Each sasmodel shape has a neutron-style synthetic dataset and an x-ray style synthetic dataset, and each of these has two sets of initial values; so that in total one sasmodel shape has four problems associated with it.

#### Testing Instructions

To run all of the problems: "fitbenchmarking\examples\benchmark_problems\SAS_modelling\*" . To run individual sets, replace the * with the folder of choice.

Checking the runtimes on another machine would be greatly appreciated as they are pretty slow on mine.


tyronerees:
I've run this set, and it looks good to me.  The residuals with an `nlls` cost function are large 6.582e4 -> 5.011e7, but good methods seem to consistently converge to the same local minimum.   `lm-bumps` takes 1.251s to solve the hardest problem on my machine, so any slowness is due to poor algorithms, not inherently due to the problem set.  

There is slightly more variability in terms of the minima with the `weighted_nlls` cost function, but there is at least one method that can solve all problems in less than 0.7s, so again it seems to differentiate between good and poor methods well.

tyronerees:
I've run this set, and it looks good to me.  The residuals with an `nlls` cost function are large 6.582e4 -> 5.011e7, but good methods seem to consistently converge to the same local minimum.   `lm-bumps` takes 1.251s to solve the hardest problem on my machine, so any slowness is due to poor algorithms, not inherently due to the problem set.  

There is slightly more variability in terms of the minima with the `weighted_nlls` cost function, but there is at least one method that can solve all problems in less than 0.7s, so again it seems to differentiate between good and poor methods well.

AndrewLister-STFC:
Was the intention to add the scripts you used to generate these files too?

BenPetersRAL:
> Was the intention to add the scripts you used to generate these files too?

I think I should as it's just one module that is pretty easy to use. I was put off from uploading it because this module has quite a few functions in there that aren't used (they were there to record the different synthesis approaches I'd tried). I can add this module to the scripts directory and then push to this PR if you think it's worth it.

AndrewLister-STFC:
I ran this on friday with bumps, scipy-ls, and dfo using an nlls cost func.

Parallelepiped was the slowest coming in at 1min10s to 11mins and 2mins to 9 mins for neutron and xray respectively.
The rest all completed in reasonable times and I think the 1/2min timeframe is fine with worse ones taking longer not being something we should try to control.

I will say that because the data has such a big range, the cost function is dwarfed by the errors in smaller x's which leads to quite bad fits for the larger x's, minimizers also seem to get the same chi squared but have very different fits. I'll run it again with a weighted cost function which should help with this.

AndrewLister-STFC (approved):

Issue #800, opened 2021-05-17
Failed minimizers report in results GUI typos & bugs


BenPetersRAL:
**Description of the documentation**
In the results GUI there are a few noticeable typos/errors that should be corrected when the "Failed minimizers" report is included. See the screenshot below that was taken when running all of the the default softwares when "algorithm_type" was set to "general". The red annotation highlights the typos, the green highlights the unnecessary repetition of the minimizers that did not run and the blue highlights the occasions where the software is announced despite none of it's minimizers failing.

![image](https://user-images.githubusercontent.com/75368443/118504513-7c2d1600-b723-11eb-8831-e886f7b87017.png)

jess-farmer:
Fixed by #818

PR #818, opened 2021-06-01
Failed minimizers report in GUI correction

BenPetersRAL:
#### Description of Work

Corrected typos, improved the written expression and removed unnecessary text from the failed minimizers report seen in the GUI if the user specifies minimizers that conflict with the algorithm type option. See new report below, refer to #800 to see previous report.

image

#### Testing Instructions
To see this failed minimizer report run an example set of problems with the following specifications in the options file:
1. An "algorithm_type" e.g. "general".
2. A set of minimizers from various software including some that are incompatible with the specified algorithm type.


jess-farmer (approved):
All looks good to me, thanks for fixing this!

Issue #827, opened 2021-06-07
Improve readability of results tables


BenPetersRAL:
The results tables have a colour scheme that make it difficult to read the numbers in the table (black on red is hard to read). Furthermore, to a new user, the meaning of the shading is not immediately obvious (see screenshot below). Lastly, the format of the cell is explained in prose above the table but could be more quickly conveyed by a "key"/"labelled example cell" for the given table.

![image](https://user-images.githubusercontent.com/75368443/120985645-b8c4be00-c773-11eb-9eeb-e3e50610ac6d.png)

**Proposed Solution**
* Update shading in table to improve readability
* Add a table "key" with two elements: 1) colour bar, 2) cell format

PR #831, opened 2021-06-07
Updating results table colours and adding key

BenPetersRAL:
#### Description of Work

The results tables have been updated with a new colour scheme and a key to make it easier for new users to interpret the tables:

image

See issue #827 for the previous colour scheme. To achieve this change, the default colours in the options.py file were changed and a new variable passed to the renderer that stores these colours from which the colourbar and cell key are shaded. Currently, the text is fixed as white so specifying shading colours that are not dark enough will be problematic.

#### Testing Instructions

1. Run fitbenchmarking with any problem set and view the results tables

Could we lock in a colour scheme and remove the colour option?


tyronerees:
Closing this, as superseded by #838

PR #837, opened 2021-06-09
825 matlab controller mixin

### LINKED ISSUE #825: Create base matlab controller
Create a base matlab controller to avoid code repetition in each of the different matlab controllers

jess-farmer:
#### Description of Work

Fixes #825 

#### Testing Instructions

1. Test that matlab controllers still run as expected
2.
3.


fitbenchmarking/controllers/base_matlab_controller.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, cost_func):
+        """
+        Initialise anything that is needed specifically for matlab
+        fitting software
+        """
+        super().__init__(cost_func)
+
+        if not import_success:
+            raise MissingSoftwareError('Requirements are missing for Matlab '
+                                       'fitting, module "dill" is reuqired.')
```
AndrewLister-STFC, line 34:
```suggestion
                                       'fitting, module "dill" is required.')
```

fitbenchmarking/controllers/base_matlab_controller.py
```diff
@@ -0,0 +1,78 @@
...
+        temp_dir = TemporaryDirectory()
+        temp_file = os.path.join(temp_dir.name, 'temp.pickle')
+        with open(temp_file, 'wb') as f:
+            dill.dump(func, f)
+        eng.workspace['temp_file'] = temp_file
+        eng.evalc('py_f = py.open(temp_file,"rb")')
+        eng.evalc('func_mat = py.dill.load(py_f)')
+        eng.evalc('py_f.close()')
+
+        return eng.workspace['func_mat']
```
AndrewLister-STFC, line 64: Since this is now in a helper function it will have a side effect of polluting the namespace, I think we could do with making these variables a bit less likely to cause a clash.

E.g. a new controller could feasibly write:
```
eng.workspace['py_f'] = self.py_to_mat(self._eval_f, eng)
eng.workspace['py_j'] = self.py_to_mat(self._eval_j, eng)
```

AndrewLister-STFC (changes requested):
This all looks very neat, just a few minor questions on the code.

I think it would be good to have a test or two for the `py_to_mat` func.

Can you also add something about this class in the docs, we have instructions for new controllers [here](https://fitbenchmarking.readthedocs.io/en/latest/extending/controllers.html) but that won't cover the py_to_mat function which was the main hurdle you had to overcome to add matlab functions!

fitbenchmarking/controllers/base_matlab_controller.py
```diff
@@ -0,0 +1,78 @@
...
+    def jacobian_information(self):
+        raise NotImplementedError
+
+    def fit(self):
+        raise NotImplementedError
+
+    def cleanup(self):
+        raise NotImplementedError
+
+    def py_to_mat(self, func, eng):
```
AndrewLister-STFC, line 50: Another minor comment is that this will raise the pylint refactor warning about no self use.

e can overcome that by making it a static_method (https://docs.python.org/3/library/functions.html#staticmethod)

fitbenchmarking/controllers/base_matlab_controller.py
```diff
@@ -0,0 +1,78 @@
...
+        raise NotImplementedError
+
+    def jacobian_information(self):
+        raise NotImplementedError
+
+    def fit(self):
+        raise NotImplementedError
+
+    def cleanup(self):
+        raise NotImplementedError
```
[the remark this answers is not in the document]
jess-farmer: I added these functions in because otherwise I got an 'abstract-method' pylint warning saying that the method is abstract in class 'Controller' but is not overridden. Do you know of any ways I could make pylint happy without including these functions?

fitbenchmarking/controllers/base_matlab_controller.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        feval = -self.cost_func.eval_r(p)
+        return feval
+
+    def _jeval(self, p):
+        """
+        Function to call from matlab which evaluates the jacobian
+        """
+        jeval = -self.jacobian.eval(p)
+        return jeval
```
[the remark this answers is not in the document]
jess-farmer: I think I copied across the functions from the `_feval` and `_jeval` functions from the levmar controller which also have the negation, but testing without the negative sign doesn't affect the results so I will remove it to avoid confusion

fitbenchmarking/controllers/base_matlab_controller.py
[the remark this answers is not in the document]
jess-farmer: I couldn't remove pylint warnings using ABCMeta, so have instead made `BaseMatlabController` independent of `Controller`, meaning the matlab controllers now inherit from both the base matlab controller and base controller classes

AndrewLister-STFC (changes requested):
I think the mixin design fit's well here and is probably a better solution than the first idea of a base class (since we don't overwrite any methods, just add new ones). I think we should rename some things since it's not a base class anymore though.

Sorry to keep throwing this back to you.

fitbenchmarking/controllers/tests/test_controllers.py
```diff
@@ -756,6 +758,26 @@ def setUp(self):
         self.jac.method = '2-point'
         self.shared_tests = ControllerSharedTesting()
 
+    def test_py_to_mat(self):
+        """
+        Tests the static method py_to_mat in Base Matlab Controller,
```
AndrewLister-STFC, line 763:
```suggestion
        Tests the static method py_to_mat in MatlabMixin,
```

AndrewLister-STFC (approved):

Issue #832, opened 2021-06-08
Enable standard colormaps for results table shading


BenPetersRAL:
The results table shading is part of the plotting options in the user's option .ini file and currently requires the user to input key values and corresponding colours to define the shading. There is not an option to specify the text colour either.

The plotting options should be updated to: 

1. accept specification of recognised/standard colormaps such as "jet" and "viridis" in the options file in addition to custom key values and colours,

2. Include a text colour specification

BenPetersRAL:
Pr #838 merged, consider closed.

PR #838, opened 2021-06-10
Introduce standard colourmaps for table shading

BenPetersRAL:
#### Description of Work

Draft to address #832. The  `colour_scale` plotting option has been replaced with three new parameters: `colour_map`, `colour_ulim` and `cmap_range`. The text colour in the table has been set to be white. The `colour_map` string now matches the name of a selected standard colourmap from matplotlib, the choice has been narrowed down to sequential colour maps that are visible with **white text**, valid options are: 

`'colour_map': ['viridis', 'plasma', 'inferno', 'magma', 'cividis',
                        'Greys', 'Purples', 'Blues', 'Greens', 'Oranges', 'Reds',
                        'YlOrBr', 'YlOrRd', 'OrRd', 'PuRd', 'RdPu', 'BuPu',
                        'GnBu', 'PuBu', 'YlGnBu', 'PuBuGn', 'BuGn', 'YlGn']`
image

image

Colours are then sampled from the colourmap according to the relative values (normalised in a log scale) and converted to a hex format and stored in a dictionary. The `colour_ulim` defines the value at which the colour will not change, so that an extremely relative high values do not distort the scale; this defaults to a value of 100. The `cmap_range: [low, high]` option allows a custom range  within the colourmap to be sampled from, avoiding the extreme colours of the scales which can improve aesthetics of table and ensure readability.

#### Testing Instructions

1. Run fitbenchmarking and check that table colours are readable ( default `'magma'` )
2. Experiment with different colour maps and values for `colour_ulim` and `cmap_range`

#### Documentation: Is there a suitable change to documentation for this change?

If we are happy with this change, the documentation will need to be updated.


BenPetersRAL:
I have prepared some html results using this branch to give an idea of how the new colourmaps look, please had a look at a few of them to check the readability (and how this changes as a cursor hovers over them. Please and leave some feedback.

Perceptually uniform sequential colourmaps:
[cmap_magma_sample.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/6694728/cmap_magma_sample.zip)
[cmap_viridis_sample.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/6694738/cmap_viridis_sample.zip)
[cmap_plasma_sample.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/6694763/cmap_plasma_sample.zip)

Sequential colourmaps in one colour:
[cmap_Reds_sample.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/6694736/cmap_Reds_sample.zip)
[cmap_Greys_sample.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/6694726/cmap_Greys_sample.zip)

Sequential colourmaps in two colours:
[cmap_PuBu_sample.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/6694732/cmap_PuBu_sample.zip)
[cmap_BuPu_sample.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/6694723/cmap_BuPu_sample.zip)
[cmap_OrRd_sample.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/6694729/cmap_OrRd_sample.zip)
[cmap_YlGn_sample.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/6694740/cmap_YlGn_sample.zip)

jfowkes:
I'm a fan of the two colour sequential colourmaps (Orange-Red being my personal favourite). The perceptually uniform colormaps look weird and the one colour sequential are just too monochrome to easily parse. Others may feel differently.

BenPetersRAL:
> I'm a fan of the two colour sequential colourmaps (Orange-Red being my personal favourite). The perceptually uniform colormaps look weird and the one colour sequential are just too monochrome to easily parse. Others may feel differently.

Thanks for this. I think the two colour sequential are effective too. I like the perceptually uniform maps (magma and plasma) but I think they require a legend because it's not as obvious what the colours mean when they don't just fade to white.

jess-farmer:
I think it's really nice giving users the option to use their preferred colour map. I agree with @jfowkes that the two colour sequential colourmaps are good but I do also like the perpetually uniform ones - I think the contrast between the extreme ends of the scale makes it easy to interpret results at a glance (with a key to explain colour scale).

AnthonyLim23:
I think the two colours are best, followed by single. I think if the user can change it we just need to pick a default

AndrewLister-STFC:
Why do some of the color maps get flipped? Looking at magma I was expecting blue to be low and red high. I don't feel too strongly about this so was just curious?
Maybe we have a reversed option?

AndrewLister-STFC:
In terms of the actual colours, I like magma, reds, and PuBu best, so one from each.

I like magma, so I'm happy with that being the default, if others aren't, some thoughts: The default could be OrRd if we want to keep it looking the same, or Blues/PuBu/GnBu/YlGnBu if we want it to match the css. It looks like Viridis is the default for matplotlib too. Although any of the ones you've zipped up would probably be fine for the default.

I also like the transparency on hover, it's much better than the red!

I also spotted that there are a load more colourmaps in matplotlib; did you just cherry-pick the best/most relevent/least wild ones? I think https://matplotlib.org/stable/api/pyplot_summary.html#matplotlib.pyplot.colormaps will generate a full list of them all if we did want more.

fitbenchmarking/results_processing/base_table.py
```diff
@@ -138,28 +135,51 @@ def get_colour(self, results):
...
+                        (log_ulim - min(log_val))
+                    norm_val[norm_val > 1] = 1  # applying upper cutoff
+                    # trimming colour map according to default/user input
+                    norm_val = cmap_range[0] + \
+                        norm_val * (cmap_range[1] - cmap_range[0])
+                    rgba_list = cmap(norm_val)
+                    hex_list = []
+                    for c in rgba_list:
+                        hex_list.append(mpl.colors.rgb2hex(c))
+                    colour[key].append(hex_list)
```
AndrewLister-STFC, line 182: There's duplicated code here which made me pause for a second. I think it would probably be better to pull this out to a seperate function. e.g.:
```
@staticmethod
def _val_to_colour(vals):
    """
    Converts values to hexadecimal colour strings.

    :param vals: values in the range [0, 1] to convert to colour strings
    :type vals: list[float]

    :return: colours as hex strings for each input value
    :rtype: list[str]
    """
    ...
```

fitbenchmarking/utils/options.py
```diff
@@ -368,3 +367,30 @@ def read_list(s):
...
+    if len(rng) != 2:
+        raise ValueError("cmap_range not specified with 2 elements.")
+    if rng[0] > rng[1]:
+        raise ValueError("Incorrect element order;"
+                         "cmap_range[0] > cmap_range[1] detected."
+                         "The elements must satisfy "
+                         "cmap_range[0] < cmap_range[1].")
+    if rng[0] < 0 or rng[0] > 1 or rng[1] < 0 or rng[1] > 1:
+        raise ValueError("One or more elements in cmap_range are "
+                         "outside of the permitted range 0 <= a <= 1.")
```
AndrewLister-STFC, line 395: These messages should probably be more generic so that the range can be used for other options if we need them in the future. That being said, this can be changed when/if we want to add those options.

AndrewLister-STFC (changes requested):
This is really good, there's a little bit of duplicated code which we can probably move to it's own function. I still need to run it, I'll do that this afternoon.

fitbenchmarking/utils/options.py
BenPetersRAL: I can replace them with one message that just states the correct format for the time being. Something like: `cmap_range only accepts the following format [a, b], where a < b, a <= 1, b <=1`. Perhaps it's still not that generic but it would require less lines and be less to change if you want to repurpose `cmap_range`.

AndrewLister-STFC: I was thinking something more like removing the "cmap" bit, that should be handled by the exception handling in read_value

fitbenchmarking/results_processing/base_table.py
BenPetersRAL: I have introduced a method `_val_to_colour(vals, cmap, cmap_range, log_ulim)` in my latest push. This has reduced the repetition significantly so thanks for the suggestion. I couldn't get the `@staticmethod` to work, it gave the following error `TypeError: 'staticmethod' object is not callable` so it is just a method at the moment.

fitbenchmarking/utils/options.py
BenPetersRAL: I see what you mean now, I have made this change in my latest push.

AndrewLister-STFC (approved):
Just tested and it's much easier to tweak the colours than before! The colours are also a lot nicer to look at

PR #842, opened 2021-06-14
Add Matlab curve fitting toolbox

### LINKED ISSUE #841: Add Matlab Curve Fitting Toolbox
Is your feature request related to a problem? Please describe.
Add in fitting software option for matlab curve fitting toolbox - see #119

jess-farmer:
#### Description of Work

Fixes #841 

To do:

- [x] Fix errors in controller
- [x] Add bounds support
- [x] Add documentation
- [x] Add tests

#### Testing Instructions

1. Run Fitbenchmarking with matlab_curve fitting software selected with various fitting problems.
2.
3.


AndrewLister-STFC:
Another thought for how to get around the x argument issues:

We can wrap the python function in something that takes the index (or a list of indexes) of an x, y data pair and runs that pair(s). It's similar to how the ivp pr supports multivariate problems.

This might work and would give something meaningful as an output.

As a rough example:
```
def f(i, p):
    prob.eval_r(p, x=self.data_x[i], y=self.data_y[i])
```

tyronerees (changes requested):
I'm not sure this works quite right for the `weighted_nlls` cost function.  Here's the output for NIST_average_difficulty with `weighted_nlls`:
image
In all cases, all minimizers converge to (more or less) the same local minimum, except `matlab-curve`. I haven't looked in detail, but I suspect `eval_r` takes care of the weighting and so there's no need to pass this to matlab -- there may be something more subtle going on, though?

Otherwise, all looks good to me (`nlls` and `hellinger_nlls` both seem to work as expected).

tyronerees (approved):
Looks good -- thanks @jess-farmer!

One think I just noticed is that the links at the top of the table don't work for these minimizers: 
image

I think we just need update the docs to use the label generated by the tables -- I'll create a separate issue for this, though.

PR #849, opened 2021-06-23
fix incorrect scipy flags

### LINKED ISSUE #826: SciPy controllers return incorrect cleanup flags
The cleanup flags in the SciPy controllers are set using `OptimizeResult.status` however (as documented [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html#scipy.optimize.OptimizeResult)) its value depends on the underlying solver and this differs wildly from solver to solver. We need to handle the return flag for each SciPy solver separately as implemented in the solver code (as the status flags are rarely if ever documented).

jess-farmer:
#### Description of Work

Fixes #826 

#### Testing Instructions

1. Check that error flags match up to the output in `self.result.message` for each solver
2.
3.


AndrewLister-STFC (approved):
Looks good and tests pass

PR #855, opened 2021-06-30
Split parsing and running so that names can be made unique

### LINKED ISSUE #408: duplicate names error
**Description of the error**
If there are two problems with the same name in the folder being benchmarked, then fitbenchmarking fails with:
```
  File "/home/dgk98065/.local/bin/fitbenchmarking", line 11, in <module>
    load_entry_point('FitBenchmarking', 'console_scripts', 'fitbenchmarking')()
  File "/home/dgk98065/fitbenchmarking/fitbenchmarking/cli/main.py", line 133, in main
    run(problem_sets=args.problem_sets, options_file=args.options_file)
  File "/home/dgk98065/fitbenchmarking/fitbenchmarking/cli/main.py", line 96, in run
    options=options)
  File "/home/dgk98065/fitbenchmarking/fitbenchmarking/core/results_output.py", line 43, in save_results
    create_plots(options, results, best_results, group_name, fig_dir)
  File "/home/dgk98065/fitbenchmarking/fitbenchmarking/core/results_output.py", line 133, in create_plots
    initial_guess_path = plot.plot_initial_guess()
  File "/home/dgk98065/fitbenchmarking/fitbenchmarking/results_processing/plots.py", line 128, in plot_initial_guess
    ini_guess = self.problem.starting_values[self.count - 1].values()
IndexError: list index out of range

```

Ideally the output should still be produced, possibly by appending something to the name.  If this is not possible, then a human readable error message should be raised (ideally right at the start, so that data isn't lost)

Problems that trigger this are here: [duplicate_name_bug_example.zip](https://github.com/fitbenchmarking/fitbenchmarking/files/4107493/duplicate_name_bug_example.zip)
.

AndrewLister-STFC:
#### Description of Work
Run parsing as a seperate step to allow changing the name of the problems to ensure they're unique.

This will mean that all problems are loaded into memory before any are run which I was concerned about until I realised that problems are kept in memory after running anyway. 

Fixes #408 

#### Testing Instructions

1. Rename 2 problems in a problem directory to have the same name.
2. Run fitbenchmarking
3. Check that 2 results are generated


jess-farmer (approved):
Changes look good and work as expected!

PR #858, opened 2021-07-02
270 higher difficulty problems

### LINKED ISSUE #270: Fitbenchmarking should use higher difficulty problems.
There are many problems created for testing global minimisers. These problems generally have many local minima and only one global minima. The efficacy and efficiency of a minimiser to find the global minima is vital in many scattering analyses where the energy landscape is very complex. In such systems the utility of Levenberg-Marquardt (etc) is next to useless.
I suggest Fitbenchmarking should introduce many more global test functions (e.g. https://github.com/scipy/scipy/tree/master/benchmarks/benchmarks/go_benchmark_functions) for characterising minimisers. I suspect the majority of minimizers currently used in fitbenchmarking would fail on such a suite.

As part of such benchmarking it's not only the time taken/function evaluations required that is important. Global minimisers typically don't have a 100% success rate on such difficult problems, so the success rate of a minimiser is a vital statistic. Typically this means one should perform e.g. 50 repeats, and note down the success rate. Time taken and function evaluations are important metrics, but pale in comparison to being able to successfully find the minimum.

jess-farmer:
#### Description of Work

Fixes #270 

#### Testing Instructions

1. Run the new problem set with various options
2.
3.


jess-farmer:
@tyronerees I have added 4 problems from those found [here](https://github.com/scipy/scipy/tree/master/benchmarks/benchmarks/go_benchmark_functions) which work with fitbenchmarking. I have also included `KOWOSBNE.SIF` and `MEYER3NE.SIF` (which were on the list that Nick provided) although these SIF files currently cause errors when fitbenchmarking tries to parse the problem. Looking at these problems, they seem to be the same as `MGH09.dat` and `MGH10.dat` in the `NIST/high_difficulty` problem set - I've also noticed that these problems exist in `cutest/mastif` as `MGH09.SIF` and `MGH10.SIF`, so I wonder if I could use these instead of the two SIF files currently causing issues?

tyronerees:
@jess-farmer good spot!  Yes, I think that makes sense -- copy those files over in place of KOWOSBNE and MEYER3NE and add the bounds in the GO directory.  I still don't see where the 'x' data comes from in KOWOSBNE/MEYER3NE, but I'm sure it's generated somehow since they are nice and straightforward.

tyronerees (changes requested):
These all run and look good and sensible -- many thanks, @jess-farmer!

Could you add a short description of these problems here: https://fitbenchmarking--858.org.readthedocs.build/en/858/users/BenchmarkProblems.html Don't worry about the zip/tar.gz files -- I think it makes sense to add these when we do a new release.  

I'll add an issue to document what we need to do when adding new data sets, because this is not at all clear.

tyronerees (approved):

PR #866, opened 2021-07-05
Linting fixes

### LINKED ISSUE #865: Fix the pythint enumerate complaint
**Description of the error**
Pylint seems to have added a new critical linting error. This should be updated.

AndrewLister-STFC:
#### Description of Work
Pylint was complaining that `for i in range(len(...))` should be changed to `for i, _ in enumerate(...)`.
In this case I disagreed but have changed it so that the meaning of `len(...)` is more explicit to fix the issue.

I also fixed a linting issue with using a list as a default value.

Fixes #865 

#### Testing Instructions

1. Check that pylint passed (using version >= 2.9.3)

### DIFF
--- a/fitbenchmarking/parsing/fitting_problem.py
+++ b/fitbenchmarking/parsing/fitting_problem.py
@@ -192,7 +192,8 @@ class FittingProblem:
         else:
             # Mantid multifit problem
             self.sorted_index = []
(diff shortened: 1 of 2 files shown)


jess-farmer (approved):
Looks good, thanks for these fixes!

PR #877, opened 2021-07-08
Add overflow hide back in

### LINKED ISSUE #875: Links in results table not working correctly
**Description of the error**
In the results table, the links in the first row of results (highlighted in the screenshot below) point to the Minimizer Options docs page instead of the correct support page for the problem and minimizer. 

In all other table rows, the link when hovering over the text in the cell is correct, but, when hovering over the cell background, the link points to the support page for a different problem.

This should be fixed so that all links are correct
Screenshot 2021-07-07 165647

AndrewLister-STFC:
#### Description of Work
Reintroduce the `overflow: hidden;` line in the CSS which looks to have been removed accidentally during a previous change.
Fixes #875

#### Testing Instructions

1. Do the links work?
2. Does table scrolling still work?

### DIFF
--- a/fitbenchmarking/templates/table_style.css
+++ b/fitbenchmarking/templates/table_style.css
@@ -35,6 +35,7 @@ thead tr th.blank.level0 {
 
 th, td {
   border: 1px solid black;
+  overflow: hidden;
 }
 
 th a, td a {


jess-farmer (approved):
Links all work and are correct now, thanks for this fix!

PR #878, opened 2021-07-08
Allow access to hessians

### LINKED ISSUE #853: Allow access to Hessians
Some minimizers (e.g, RALFit, Scipy) can use second derivative information.  We should make this available for appropriate problems classes (e.g., NIST)

jess-farmer:
#### Description of Work

Fixes #853 

This PR adds the option for users to use hessians for NIST problems with fitting software that accept hessian information. The option `use_hessian` should be set to `True` for hessians to be used. If this option is set with problems/minimizers that are not compatible with using hessians then these problems/minimizers will be skipped. 

To do:
- [x] Get hessians working for nlls and weighted_nlls cost functions
- [x] Add documentation for changes
- [x] Add unit tests for new functions

#### Testing Instructions

1. Run fitbenchmarking with option `use_hessian: True` and check that problems run with reasonable results
2. Run with `use_hessian` set to false and check that results are consistent
3.


tyronerees (changes requested):
Tested locally and works as expected -- using Hessians gives a much more robust solver for NewtonCG.  

I've added #896 -- it seems that we've purposefully left out minimizers that can use Hessians, although these should be trivial to add back in.

I'm not sure the options available to the user are quite right -- I don't think we should have this option:
image

The Hessian behaviour should mirror the Jacobian behaviour so as not to make it too confusing.  For now,  I think we should have `default` and `analytic` options:  `default` would not send hessians to the minimizer, and `analytic` would.  As is the case for Jacobians, if you've asked to use a Hessian with software that can't support it, I think it should just run with defaults and move on.

The comments at the top of `core/fitting_benchmarking.py` should be updated to include `loop_over_hessians()`

Let me know if anything's not clear, or if I've missed something that means it needs to be designed this way.

fitbenchmarking/controllers/bumps_controller.py
```diff
@@ -61,6 +61,14 @@ def jacobian_information(self):
...
         return has_jacobian, jacobian_free_solvers
 
+    def hessian_information(self):
+        """
+        Bumps cannot use Hessian information
+        """
+        has_hessian = False
+        hessian_free_solvers = []
+        return has_hessian, hessian_free_solvers
+
```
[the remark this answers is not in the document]
jess-farmer: Good point - I have created a separate issue for this

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -493,4 +433,164 @@ def loop_over_jacobians(controller, options, grabbed_output):
...
+            else:
+                new_chi_sq.append(chi_sq)
+
+            # For minimizers that do not accept hessians we raise an
+            # StopIteration exception to exit the loop through the
+            # Hessians
+            if not minimizer_check:
+                raise StopIteration
+    except StopIteration:
+        pass
```
AndrewLister-STFC, line 594: This is equivalent to using `break` but more complicated. The Jacobian implementation is slightly different as it has to escape 2 levels of for loops

fitbenchmarking/core/tests/test_fitting_benchmarking_hessians.py
```diff
@@ -0,0 +1,221 @@
...
+        self.controller.invalid_jacobians = ["deriv_free_algorithm"]
+        self.controller.minimizer = "deriv_free_algorithm"
+        self.controller.has_hessian = [False]
+        self.controller.valid_hessians = ["general"]
+        jacobian = False
+        minimizer_name = "deriv_free_algorithm"
+
+        # Cleanup has been mocked out with a no-op, so set the outputs now.
+        self.controller.flag = 3
+        self.controller.final_params = [1, 2, 3, 4]
```
AndrewLister-STFC, line 209: I'm a bit confused at why this is needed, it looks like we can just set "final_params_expected" and "flag_expected" like you do with has_hessian

AndrewLister-STFC: Ahh, I've just spotted this is how it's done in jacobians... we can save this for when #901 is looked at

fitbenchmarking/hessian/base_hessian.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        Base class for the Hessians
+
+        :param cost_func: Cost function object selected from options.
+        :type cost_func: subclass of
+                :class:`~fitbenchmarking.cost_func.base_cost_func.CostFunc`
+        """
+        self.cost_func = cost_func
+        self.problem = self.cost_func.problem
+        self.jacobian = None
```
AndrewLister-STFC, line 24: If the hessian is now after the jacobian, can this be an argument?

fitbenchmarking/hessian/hessian_factory.py
```diff
@@ -0,0 +1,37 @@
...
+from fitbenchmarking.hessian.base_hessian import Hessian
+from fitbenchmarking.utils.exceptions import NoHessianError
+
+
+def create_hessian(hes_method):
+    """
+    Create a Hessian class.
+
+    :param jac_method: Type of Hessian selected from options
+    :type jac_method: str
```
AndrewLister-STFC, line 18:
```suggestion
    :param hes_method: Type of Hessian selected from options
    :type hes_method: str
```

fitbenchmarking/parsing/nist_parser.py
```diff
@@ -89,6 +100,30 @@ def _parse_jacobian(self, name):
...
+        """
+        Parses the Hessian for the NIST file format
+
+        :param name: name of the NIST file
+        :type name: str
+        """
+        file_dir = os.path.abspath(os.path.join(self._filename, os.pardir))
+        hes_file = os.path.join(file_dir, "data_files", "{}.hes".format(name))
+        try:
+            hes_data = open(hes_file, "r")
```
AndrewLister-STFC, line 113: This file needs to be closed. Generally this is done using the with syntax, where exiting the indented block cleans up the files for you:
```
with open(...) as f:
    lines = f.readlines()

...
```

AndrewLister-STFC (changes requested):
Looks good, I spotted afew minor things that could do with tweaking. Let me know if you think any should be put in #901 (as I haven't checked if these comments are how the jacobians are done!)

fitbenchmarking/parsing/nist_parser.py
jess-farmer: This is the same for the `_parse_jacobian` function - I can either fix both in this PR or the issue can be addressed in #901, what do you think?

AndrewLister-STFC: I think we should change this one as it's not just a style/implementation change that's required. Although I don't have a strong preference over using the context manager (with) vs just closing explicitly (`f.flush();f.close()`) if it's temporary -- after #901 it should be a with statement though.

AndrewLister-STFC (approved):
Looks great now, thanks for the edits!

PR #880, opened 2021-07-08
Fix incorrect minimizer links

### LINKED ISSUE #864: Matlab minimizers links
The links on the table for the Matlab minimizers:
image
don't go to the correct place.  The labels in the docs should be updated accordingly.

jess-farmer:
#### Description of Work

Fixes #864 

#### Testing Instructions

1. Check that the minimizer links in the results table now point to the correct fitting software from the Minimizer Options docs page

### DIFF
--- a/docs/source/users/options/minimizer_option.rst
+++ b/docs/source/users/options/minimizer_option.rst
@@ -81,8 +81,8 @@ The DFO minimizers are set as follows:
    See :ref:`extra_dependencies`.
 
 
-Gradient-Free-Optimizers (``gradient_free``)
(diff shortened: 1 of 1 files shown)


tyronerees (approved):
Thanks, @jess-farmer -- looks good.  I think, more generally, this part needs a refactor, as I don't see why the minimizer name needs to be in this format in the title (the docs looked more natural as you originally wrote them).  I'll create the issue for that, but at least the links work for now.

tyronerees:
Issue #881 created to resolve the underlying issue

PR #885, opened 2021-07-14
Updating sasmodel problem error column headers

### LINKED ISSUE #876: Sasview problems not parsing due to datafile headers
**Description of the error**
Sasview problem sets are returning parsing errors because the header of the error column is not recognised. Currently, the header of the error column is `<devY>` which is unrecognised. This should be changed to `<E>` for all sasview problem datafiles.

image

BenPetersRAL:
#### Description of Work

Header of error column in datafiles changed for all sasview problem. Fixes #876.


tyronerees (approved):
Thanks, @BenPetersRAL!

PR #887, opened 2021-07-16
fix analytic jacobian for hellinger_nlls

### LINKED ISSUE #886: Hellinger nlls cost function -- Jacobian incorrect?
It seems like the Jacobian for the Hellinger nlls cost function is incorrect.  If we have 
image
with Jacobian 
image
then for the Hellinger residual:
image
we should have 
image
However, the code actually calculates:
https://github.com/fitbenchmarking/fitbenchmarking/blob/c35a839af9dc75807c452ebcf2346c18fb4ce558/fitbenchmarking/jacobian/analytic_jacobian.py#L39-L41

This presumably explains the difference between accuracy in `scipy:2-point` and `analytic` Jacobian seen here:
image

jess-farmer:
#### Description of Work

Fixes #886 

#### Testing Instructions

1. Run fitbenchmarking with analytic and scipy jacobian options selected and check that results are consistent
2.
3.

### DIFF
--- a/fitbenchmarking/jacobian/analytic_jacobian.py
+++ b/fitbenchmarking/jacobian/analytic_jacobian.py
@@ -2,6 +2,7 @@
 Module which acts as a analytic Jacobian calculator
 """
 from numpy import matmul
+from numpy import sqrt
 
 from fitbenchmarking.jacobian.base_jacobian import Jacobian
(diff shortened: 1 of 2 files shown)


tyronerees (approved):
Perfect -- thanks, @jess-farmer!

PR #893, opened 2021-07-22
Maximum Time Option

### LINKED ISSUE #846: New option: maximum time
Some softwares that have been added lately have a long runtime. We should have an option to limit the amount of time any minimizer can take.

This could be an option that just takes the maximum number of seconds, or it could take a string like "100s", "1m,40s","2h,15m",etc.

jess-farmer:
#### Description of Work
This PR adds a `max_runtime` option which allows you to set a maximum time that you want your benchmark of each fitting combination to run for. The default has been set to 10 minutes (600 seconds) - other suggestions are welcome.

The flag 6 indicates when the `max_runtime` is reached.

Fixes #846 

#### Testing Instructions

1.  Set `max_runtime` option and check that slow minimizers are not run and instead show error flag 6.


fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -410,12 +413,48 @@ def loop_over_jacobians(controller, options, grabbed_output):
...
+                        # Set up thread to run Timeit function
+                        t = threading.Thread(target=Timeit)
+                        t.start()
+                        t.join(options.max_runtime)
+
+                        # if thread is still running after the
+                        # max runtime has elapsed, then stop
+                        # the thread and raise MaxRuntimeError
+                        if t.is_alive():
+                            t.join()
```
AndrewLister-STFC, line 444: Does this stop the thread? I thought join just waited patiently?

jess-farmer: Good point - there doesn't appear to be any way to stop threads, so I have switched to using processes, using the `multiprocess` package which also works on windows

Anders-Markvardsen:
Here are some first test results using WSL2 ubuntu 18.04. Running `fitbenchmarking -o examples/options_template_anders.ini -p examples/benchmark_problems/NIST/average_difficulty/` with max_runtime=600 and num_runs=1 gives

image

with max_runtime=0.1 and num_runs=1 gives

image

with max_runtime=0.1 and num_runs=5 gives

image

Hence looks good so far

Anders-Markvardsen:
Testing on Windows. Something is not quite working with the timings on Windows. Running again `fitbenchmarking -o examples/options_template_anders.ini -p examples/benchmark_problems/NIST/average_difficulty/ with num_runs=1. Setting max_runtime=600 or max_runtime=2.5 both have all runs completing and tables displaying as expected, e.g. 



the table above suggest setting max_runtime=0.1 should have some passing fine, however none are with this value - hence this is a bug.

Secondly, picking instead e.g. max_runtime=1.1 now some are passing through



As can be by comparing the numbers in the first row in the above two table, with max_runtime=1.1 all though be passing through but only some are. Maybe `timeit` time thinks by default different on windows?

jess-farmer:
Currently when running Mantid or SASView problems, I am finding that every minimizer (other than `mantid`) times out, no matter what `max_runtime` is set to. It seems that the call to `controller.fit` within this line:
```
runtime_list.extend(timeit.Timer(
                                                setup=controller.prepare,
                                                stmt=controller.fit
                                                ).repeat(num_runs, 1))
```
is not running properly and hangs until the max runtime is exceeded.

@Anders-Markvardsen - I think the tables you showed at the meeting yesterday were for SASView problems and run from this branch, did you experience any similar problems to the above, or if you run the same command again following my recent changes, are you still able to produce the same results? I have also removed the serialization of the controller, which we suspected was causing the timing problems on Windows, please could you test if this change results in an improvement in the timings for Windows?

Anders-Markvardsen:
@jess-farmer will repeat. About running against mantid not done this yet but in the process. Will aim to join the stand-up today

jess-farmer:
@Anders-Markvardsen thanks - I'm unfortunately not able to make today's stand up so if you could also let me know your findings on here/send me a message on slack that would be much appreciated.

Anders-Markvardsen:
@jess-farmer no worries. I have repeating the Windows. With max_runtime = 0.1 no run completed and with max_runtime = 1.1 observe

image

hence appears the timing less accurate. Thanks

Anders-Markvardsen:
Here are more observation using Windows with commits as of this point in time. max_runtime = 600

image

max_runtime = 2 then all passes and

image

max_runtime = 1.5 then

image

so timings with this tech solution similar-ish in terms of windows interpret run_maxtime to above

Anders-Markvardsen:
On ubuntu 18.04 Sasview still appear to run fine for me. With max_runtime = 3:

image

jess-farmer:
@Anders-Markvardsen many thanks for testing these scenarios and sharing your observations

AndrewLister-STFC:
This works for me on WSL and the code all looks very neat. I'll hold off on a review until the tests are up and running again though.

robertapplin:
Thanks for testing @AndrewLister-STFC , the tests are running ok now

tyronerees:
@Anders-Markvardsen  -- would you be able to check this does what you need?  Andrew's been through the code, and is happy with it, so I think we just need a quick functional approval from you.

Anders-Markvardsen:
Here are some observation running this on Windows, and with new implementation it should be agnostic to platform. Running `examples\benchmark_problems\NIST\average_difficulty` with a high max_runtime gives



Selecting max_runtime to 2.1 and 1.1 make no difference to above, which means, as we know the value of max_runtime may not match exactly runtime in tables. Picking value of 0.1 gives



which is a sufficiently acceptable result in that the effect of max_runtime generally stopped those jobs which where running the longest

Anders-Markvardsen (approved):
As per comments. 

Thanks

PR #899, opened 2021-08-02
Update description of caching warning

### LINKED ISSUE #898: Better description of caching warning
Concerning warnings currently outputted as

```
/home/ajm64/venv_fitbenchmarking/lib/python3.6/site-packages/fitbenchmarking/core/fitting_benchmarking.py:474: UserWarning: The ratio of the max time to the min is 4.399321326763531 which is  larger than the tolerance of 4, which may indicate that caching has occurred in the timing results
```

Clarify what is mean by caching above and also output absolute value, which will help indicate if the problem is caching or just because hardware gets used for something else in parallel

Anders-Markvardsen:
#### Description of Work

Fixes #898

#### Testing Instructions

1.  Read description of updated warning of potential caching issue and change that this description is valid


Anders-Markvardsen:
Windows build is not right and with not pinned versions in requirements.txt including flake8 and as can be seen from

ddd

In commit above pinned versions of all dependenties in requirements.txt, which may need to adjustments since initial guess at what these should be

Anders-Markvardsen:
@AndrewLister-STFC and/or @tyronerees do we still require coverage<5.0 in requirements.txt?

AndrewLister-STFC:
> 
> 
> @AndrewLister-STFC and/or @tyronerees do we still require coverage<5.0 in requirements.txt?

Looking at the history that was added 2 years ago to fix something like this... so we can probably remove it and assume they have fixed whatever our problem was.

In general I'd also recommend pinning versions with a < or > so that pip can have some slack in case the dependencies don't line up.

Anders-Markvardsen:
> > @AndrewLister-STFC and/or @tyronerees do we still require coverage<5.0 in requirements.txt?
> 
> Looking at the history that was added 2 years ago to fix something like this... so we can probably remove it and assume they have fixed whatever our problem was.
> 
> In general I'd also recommend pinning versions with a < or > so that pip can have some slack in case the dependencies don't line up.



> > @AndrewLister-STFC and/or @tyronerees do we still require coverage<5.0 in requirements.txt?
> 
> Looking at the history that was added 2 years ago to fix something like this... so we can probably remove it and assume they have fixed whatever our problem was.
> 
> In general I'd also recommend pinning versions with a < or > so that pip can have some slack in case the dependencies don't line up.

Maybe, but is it true that in theory with Dependabot it should ensure that OK versions are used for the OS configurations we support? And if this is true we should be able to just bin versions?

Anders-Markvardsen:
@AndrewLister-STFC and @tyronerees all test are now passing and builds order of magnitude faster on Windows as a bonus. I have pinned most of the packages but if you prefer change these to >= before you merge this PR. Note changes in this issue may help testing of #846

jess-farmer (approved):
Approved - thanks for these changes @Anders-Markvardsen

PR #900, opened 2021-08-04
Refactor of table inheritance

AndrewLister-STFC:
#### Description of Work

Refactors the tables to be neater and more maintainable.

Most notably, this allows the tables to be sorted in a more flexible way.
This will be useful as more critereon are brought in and tables have more columns that need to be sorted.

#### Testing Instructions

1. The regressions tests will fail but it should only be due to a reordering of the elements.
   We will have to do them by hand.
2. Make some results with various jacobians, minimizers, softwares, etc. Do the tables look correct and match master?


AndrewLister-STFC:
Still to do:

- [ ] Write some tests
- [ ] Edit the docs to match the changes

fitbenchmarking/results_processing/base_table.py
```diff
@@ -61,296 +66,356 @@ def __init__(self, results, best_results, options, group_dir,
...
 
+        # Format the row labels
         index = []
         for b, i in zip(self.best_results, table.index):
             rel_path = os.path.relpath(path=b.support_page_link,
                                        start=self.group_dir)
             index.append('<a href="{0}">{1}</a>'.format(rel_path, i))
         table.index = index
+        print(table.columns)
+        print(table.index)
```
jess-farmer, line 401: Have these print statements been intentionally left in?

fitbenchmarking/results_processing/base_table.py
```diff
@@ -61,296 +66,356 @@ def __init__(self, results, best_results, options, group_dir,
...
+        with rows and columns as the key and list elements respectively.
+        This is used to create HTML and txt tables.
+        This is stored in self.sorted_results
+        """
+
+        # Might be worth breaking out into an option in future.
+        # sort_order[0] is the order of sorting for rows
+        # sort_order[1] is the order of sorting for columns
+        sort_order = (['problem'],
+                      ['software', 'minimizer', 'jacobian'])
```
jess-farmer, line 170: Should `sort_order` also include hessian names? Although I suppose at the moment with only one hessian method it won't make a difference

fitbenchmarking/results_processing/tests/test_base_table.py
```diff
@@ -0,0 +1,230 @@
...
+from fitbenchmarking.results_processing.base_table import Table
+from fitbenchmarking.utils.fitbm_result import FittingResult
+from fitbenchmarking.utils.options import Options
+
+
+def generate_results():
+    """
+    Create a predictable set of results.
+
+    :return: Set of manuall generated results
```
jess-farmer, line 23:
```suggestion
    :return: Set of manually generated results
```

jess-farmer (changes requested):
Changes look good and work well - just a few very minor comments

fitbenchmarking/results_processing/base_table.py
AndrewLister-STFC: Good spot, should be fixed now!

fitbenchmarking/results_processing/base_table.py
AndrewLister-STFC: Yes it should! I think I've hooked it up properly now

jess-farmer (approved):
Looks great, approved

PR #904, opened 2021-08-10
Fix poisson cost func/analytic jacobian combination

### LINKED ISSUE #891: Poisson cost function not compatible with analytic jacobian
**Description of the error**
When running fitbenchmarking with poisson cost function and analytic jacobian selected, for minimizers that accept jacobian information the following error message is output for each minimizer:
`'PoissonCostFunc' object has no attribute 'cache_rx'`

Because the poisson cost function is not a least squares problem it has no `cache_rx` attribute or `eval_r` function, so the following line in `analytic_jacobian` causes the issue:
```
rx = self.cached_func_values(self.cost_func.cache_rx,
                                     self.cost_func.eval_r,
                                     params,
                                     **kwargs)
```
An alternative to this line should be used for cost functions that are not least squares problems.

jess-farmer:
#### Description of Work

Fixes #891

#### Testing Instructions

1. Run fitbenchmarking with poisson cost function and analytic jacobian selected
2.
3.


fitbenchmarking/jacobian/analytic_jacobian.py
```diff
@@ -39,9 +38,11 @@ def eval(self, params, **kwargs):
             jac = jac / e[:, None]
         elif self.problem.options.cost_func_type == "hellinger_nlls":
             # calculates the Jacobian of the hellinger(root) NLLS cost function
-            jac = jac/(2*sqrt(self.problem.eval_model(params, x=x)[:, None]))
+            jac = jac/(2*np.sqrt(self.problem.eval_model(params, x=x)
+                       [:, None]))
         elif self.problem.options.cost_func_type == "poisson":
             jac = jac * (1 - y / self.problem.eval_model(params, x=x))[:, None]
```
AndrewLister-STFC, line 44: This could be having issues as it's only accurate when `y[i] != 0`...

For the `y[i] == 0` case, it will be:

```
cf=self.problem.eval_model(x=x, params=params)
jac=jac
```

So we could try using a mask like in the `_safe_a_log_b` function in `poisson_cost_func.py`.


I also see what you were saying with the penalise negative values discontinuity... I guess we could try setting the jacobian to 0 if something does take a step that ends up with a negative value (although as far as I know, none of the methods take steps towards a worse result!)

AndrewLister-STFC (approved):
Looks good! Tested with NIST/low_dificulty and it makes some of them perform better when compared with the scipy jac.

PR #906, opened 2021-08-13
Add Hessians for hellinger_nlls and poisson cost functions

### LINKED ISSUE #889: Add `hellinger_nlls` and `poisson` cost functions to analytic hessian class
The analytic hessian class currently only works for the `nlls` and `weighted_nlls` cost functions, the `hellinger_nlls` and `poisson` cost functions should also be added.

jess-farmer:
#### Description of Work

Fixes #889 

#### Testing Instructions

1. Run fitbenchmarking with analytic hessians selected for both hellinger_nlls and poisson cost functions for scipy Newton-CG minimizer. Check results are comparable for this minimizer when hessians are not used.
2.
3.


fitbenchmarking/hessian/tests/test_hessian.py
```diff
@@ -115,7 +123,60 @@ def test_analytic_cutest_weighted(self):
...
+                * self.actual[:, :, i] - \
+                1/2*(f(self.fitting_problem.data_x, *self.params)**(-3/2))[i]\
+                * np.matmul(jac_actual_h.T, jac_actual_h)
+        scaled_actual = np.matmul(scaled_actual, np.sqrt(
+            self.fitting_problem.data_y)-np.sqrt(f(self.fitting_problem.data_x,
+                                                   self.params[0],
+                                                   self.params[1])))
+        self.assertTrue(np.isclose(scaled_actual, eval_result).all())
+
+    def test_analytic_cutest_poisson(self):
```
AndrewLister-STFC, line 157: Similarly here, I think taking:
```
f(x, p1, p2) = e**(x*(p1 + p2))
```
will lead to
```
H[i](p1, p2) = x**2*f(x, p1, p2)*np.ones((2,2))
```

AndrewLister-STFC (changes requested):
I have a few notes, the main one is about the tests. Where I think we should calculate a hessian by hand and verify against that rather than repeating the route that the function it's testing takes.

fitbenchmarking/hessian/analytic_hessian.py
```diff
@@ -32,20 +32,36 @@ def eval(self, params, **kwargs):
...
+            for i, (f, j) in enumerate(zip(model_eval, J)):
+                j = np.array([j])
+                grad2_r[:, :, i] = 1/2 * f**(-1/2) * grad2_r[:, :, i] \
+                    + f**(-1/2) * np.matmul(j.T, j)
+        if self.problem.options.cost_func_type == "poisson":
+            model_eval = self.problem.eval_model(params, x=x)
+            for i, (f, j) in enumerate(zip(model_eval, J)):
+                j = np.array([j])
+                grad2_r[:, :, i] = (-grad2_r[:, :, i] * (1-y[i]/f)
+                                    + (y[i]/f**2)*(1/(1-y[i]/f)**2)
```
AndrewLister-STFC, line 54: I think this is right but the f's cancel? I doubt it will change performance but it will reduce the number of operations

fitbenchmarking/hessian/tests/test_hessian.py
```diff
@@ -15,54 +21,92 @@
...
+    """
+    return np.column_stack((np.exp(p[1]*x),
+                            p[0]*x*np.exp(p[1]*x)))
+
+
+def H_poisson(x, p):
+    """
+    Analyic Hessian evaluation of f_poisson
+    """
+    return np.array([[0*(np.ones(x.shape[0])),
```
AndrewLister-STFC, line 94: I think this is the same as np.zeros/np.zeros_like

AndrewLister-STFC (changes requested):
This looks much better, I'm glad we managed to weed out the tricky bugs in the hessians!!

I've only got 2 very minor comments.

AndrewLister-STFC (approved):

PR #910, opened 2021-08-20
Refactor jacobian and hessian information

### LINKED ISSUE #901: Refactor Jacobian and Hessian information functions
The functions `jacobian_information` and `hessian_information` currently have to be implemented in each controller. These should instead be instantiated in the base controller class, with the default being that jacobians/hessians cannot be used. This way a user is only required to update the function in a controller if jacobians/hessians can be used.

jess-farmer:
#### Description of Work

Fixes #901 

#### Testing Instructions

1. Check that jacobians/hessians are still only used for minimizers that accept them
2.
3.


fitbenchmarking/controllers/base_controller.py
```diff
@@ -293,12 +293,10 @@ def check_attributes(self):
                             'list or numpy ndarray of floats, got '
                             '{}'.format(attr_name, attr))
 
-    @abstractmethod
     def jacobian_information(self):
```
AndrewLister-STFC, line 296: I think we should just make this a list in the class (like `algorithm_check`). The bool can probably also be removed if we make the list show the enabled ones, as in the hessian func, as that is the same as checking if the list is empty (`if controller.hessian_enabled_solvers`)

AndrewLister-STFC (changes requested):
I'm not sure this goes far enough. Since the function returns the same thing every time it might as well just be a class attribute.

fitbenchmarking/controllers/base_controller.py
```diff
@@ -58,6 +58,26 @@ class Controller:
                        'steepest_descent': [],
                        'global_optimization': []}
 
+    #: Within the controller class, you must define the list
+    #: ``jacobian_enabled_solvers`` if any of the minimizers
+    #: for the specific software are able to use jacobian
+    #: information.
+
```
AndrewLister-STFC, line 65: The docs only show the second paragraph here. Is that intentional? If we want the full thing we need to have a single block where each line starts with `#:` (including any blank lines).

AndrewLister-STFC (commented):
Looks good to me, just 1 question about the docs

fitbenchmarking/controllers/base_controller.py
jess-farmer: Ah, thanks for spotting this - have now updated it so that both paragraphs are shown in the docs

AndrewLister-STFC (approved):

PR #918, opened 2021-09-13
Use Relative Path for colourbar image

### LINKED ISSUE #911: Table key should use relative path
The table key should use a relative (not absolute) path otherwise it will not show up in certain cases, e.g. see below:

Screenshot 2021-08-23 at 9 53 55 am

(Here I am serving the results pages to another machine via a ssh tunnel).

robertapplin:
#### Description of Work
This PR fixes a bug where the colourbars that are saved were not being returned and so were said to be 'None' in the comparison page. It also ensures that the path to the colourbars is a relative path so the bug in the issue doesn't occur when ssh tunnelling to a different machine.

Fixes #911

#### Testing Instructions

1. Run a fitbenchmark of any problem. Open the results and you should see the colourbar is present instead of saying 'None'
2. Right click the colour bar and Inspect it. The path to the image should be relative.

image

### DIFF
--- a/fitbenchmarking/results_processing/base_table.py
+++ b/fitbenchmarking/results_processing/base_table.py
@@ -558,4 +558,4 @@ class Table:
 
         plt.savefig(fig_path, dpi=150)
 
-        return fig_path
+        return os.path.relpath(fig_path, self.group_dir)
--- a/fitbenchmarking/results_processing/base_table.py
+++ b/fitbenchmarking/results_processing/base_table.py
@@ -518,7 +518,7 @@ class Table:
 
         return hex_strs
 
-    def save_colourbar(self, fig_dir, n_divs=100, sz_in=(3, 0.8)):
+    def save_colourbar(self, fig_dir, n_divs=100, sz_in=(3, 0.8)) -> str:
         """
         Generates a png of a labelled colourbar using matplotlib.
 
@@ -528,6 +528,9 @@ class Table:
         :type n_divs: int
         :param sz_in: dimensions of png in inches [width, height]
         :type sz_in: list[float] - 2 elements
+
+        :return: The relative path to the colourbar image.
+        :rtype: str
         """
         fig_path = os.path.join(fig_dir, "{0}_cbar.png".format(self.name))
 
@@ -554,3 +557,5 @@ class Table:
         fig.set_size_inches(sz_in[0], sz_in[1])
 
         plt.savefig(fig_path, dpi=150)
+
+        return fig_path
--- a/fitbenchmarking/results_processing/tests/test_base_table.py
+++ b/fitbenchmarking/results_processing/tests/test_base_table.py
@@ -202,11 +202,9 @@ class DisplayStrTests(TestCase):
 
     def setUp(self):
         results = generate_results()
-        self.root_directory = os.path.join(os.path.dirname(__file__),
-                                           os.pardir, os.pardir, os.pardir)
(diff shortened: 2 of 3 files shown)


fitbenchmarking/results_processing/base_table.py
```diff
@@ -518,7 +518,7 @@ def vals_to_colour(vals, cmap, cmap_range, log_ulim):
 
         return hex_strs
 
-    def save_colourbar(self, fig_dir, n_divs=100, sz_in=(3, 0.8)):
+    def save_colourbar(self, fig_dir, n_divs=100, sz_in=(3, 0.8)) -> str:
```
AndrewLister-STFC, line 521: Have you used type annotations much? We haven't used them on this project yet but there's no reason not to.

fitbenchmarking/results_processing/tests/test_base_table.py
```diff
@@ -230,3 +233,21 @@ def test_display_str_both(self):
         self.table.options.comparison_mode = 'both'
         s = self.table.display_str([7, 9])
         self.assertEqual(s, '9 (7)')
+
+    @mock.patch("fitbenchmarking.results_processing.base_table.plt.savefig")
```
AndrewLister-STFC, line 237: We generally have 1 test file per source file and within, 1 class per function. Can you split this off into a new class.

fitbenchmarking/results_processing/tests/test_base_table.py
```diff
@@ -230,3 +233,21 @@ def test_display_str_both(self):
...
+        Test the relative path to the colourbar figure is returned after saving
+        """
+        self.table.name = "fake_name"
+        colourbar_file = f"{self.table.name}_cbar.png"
+
+        figure_sub_directory = "fitbenchmarking_results"
+        figure_directory = os.path.join(self.root_directory,
+                                        figure_sub_directory)
+
+        relative_path = f"{figure_sub_directory}/{colourbar_file}"
```
AndrewLister-STFC, line 249: I think os.path.join here will remove the need for the replace in the check at the end?

fitbenchmarking/results_processing/base_table.py
robertapplin: Yes I tend to use them all the time, when I remember to. They've been quite useful to me in the past, especially when coming across a new piece of code. Other than being good for documentation, there are other benefits too https://stavshamir.github.io/python/the-other-benefit-of-python-type-annotations/

I noticed that you specify the `type` of parameters in the docstring so maybe its not necessary to use type annotations in this case?

AndrewLister-STFC (changes requested):
Good job, only 1 small point about the structure of tests. I'm also interested to hear any thoughts on type hints, I've started trying to use them in other projects but they haven't been used here yet.

fitbenchmarking/results_processing/base_table.py
AndrewLister-STFC: I don't see a problem with them being here, we should probably use them more... I heard about "mypy" as well recently so that may be something to aim for if we start adding them everywhere

fitbenchmarking/results_processing/tests/test_base_table.py
robertapplin: Thanks, I think that works much better!

AndrewLister-STFC (approved):
This looks good, I'll update the branch now and merge tomorrow (when the tests have run on the updated verson)

PR #921, opened 2021-09-16
Add summary page for problems

### LINKED ISSUE #491: Compare different cost functions
Find and implement a metric which allows users to compare the success of different cost functions to fit their data in one table.

AndrewLister-STFC:
#### Description of Work

Adds a summary page so that cost functions can be more easily compared.

Fixes #491

Summary of main changes:
- **Allow multiple cost functions**
  This is the main reason for the pr. Basically allow users to run fitbenchmarking with multiple cost functions and fix all associated problems...
- **Flatten the list of results returned from core/fitting_benchmarking**
  This is so that the results can be sorted as the display sees fit and further decouples the 2 parts of the software. One possible use for this is that it will be easier to allow users to select what the columns and rows in the table represent.
- **All results functions use a preprocessed dataset**
  This formats the data into a a dictionary of the form `{<row_name>: {<category_name>: [<column_result>, ...], ...}, ...}`. The terminology might be a bit confusing in places so feel free to make me fix them if it is, but the idea is that categories may be something other than cost_functions in the future. Categories introduce a new element to the tables where a category is displayed as new columns but comparisons are restricted to within the category.
- **Summary Page**
  This adds plots to compare category results and can be accessed by clicking the name of the problem on an index page.
- **Remove cost function descriptions**
  The existing method for this was not easily extended and seemed like it would be too much if there were multiple. I suggest these are added back in in a new PR (via the cost_function class in results_processing without passing it back from fitting_benchmarking) with some extra thought on the best way of showing 2 or 3 in the UI.
- *Fixed `minimzers` typo in fitting_benchmarking...*
  This was annoying me when I was editing the lines with it on...

I'm aware that this PR is already huge (apologies to anyone reviewing...) so will likely create new issues for the following when this is merged:
- Add links to support pages for best result from summary page (#?)
- Parse descriptions from problems to show in summary pages (#948)
- Reintroduce cost function descriptions as accordian/tabbed/new page(?) on index page. (#?)
- Add support for log plots in the problem summary page (#?)

#### Testing Instructions

1. Run fitbenchmarking with a single cost function and check the results including the new summary page
2. Run again with multiple cost functions


fitbenchmarking/core/tests/test_fitting_benchmarking_benchmark.py
```diff
@@ -63,36 +63,35 @@ def setUp(self):
         self.options = Options()
         self.options.software = ["scipy"]
         self.scipy_len = len(self.options.minimizers["scipy"])
+        print(self.options.minimizers)
+        print(self.scipy_len)
```
AndrewLister-STFC, line 67:
```suggestion
```

fitbenchmarking/results_processing/tables.py
```diff
@@ -65,12 +67,12 @@ def create_results_tables(options, results, group_name, group_dir, fig_dir,
     for suffix in SORTED_TABLE_NAMES:
         if suffix in options.table_type:
 
-            table_names[suffix] = \
-                f'{group_name}_{suffix}_{options.cost_func_type}_table.'
+            table_names[suffix] = f'{group_name}_{suffix}_table.'
```
AndrewLister-STFC, line 70: @tyronerees I seem to remember you saying something about filenames needing the cost function in. Obviously this doesn't work with multiple cost functions, so if it's needed for something can you let me know as we'll need to figure out a new solution...

AndrewLister-STFC: Ahh that makes sense. Does the problem directory not get wiped when it starts fresh anyway? Thanks for the clarification :)

fitbenchmarking/results_processing/support_page.py
```diff
@@ -13,86 +13,84 @@
...
+    fit_success = init_success = options.make_plots
+
+    if options.make_plots:
+        fig_fit, fig_start = get_figure_paths(result)
+        if fig_fit == '':
+            fig_fit = result.figure_error
+            fit_success = False
+        if fig_start == '':
+            fig_start = result.figure_error
+            init_success = False
```
robertapplin, line 66:
```suggestion
        fit_success, init_success = fig_fit != "", fig_start != ""
        fig_fit = result.figure_error if not fit_success else fig_fit
        fig_start = result.figure_error if not init_success else fig_start
```

robertapplin (changes requested):
It looks good to me. I thought of two more code suggestions but looks ready for testing

fitbenchmarking/results_processing/support_page.py
AndrewLister-STFC: I'll make this a bit neater, but I think it's a bit more readable with an if on a seperate line when one of the assignments is a noop.

robertapplin (dismissed):
Thanks, all good from a code point of view now. Ready for some testing @tyronerees

robertapplin (dismissed):

tyronerees:
A couple of cosmetic/functional comments (which may also be separate issues):
- there's nothing in the docs to say that multiple cost functions can be run.  The same is true for, e.g., the Jacobians, so maybe there's wider issue to tidy up this part of the docs.
- The index page still has a **Cost function** section, which now has nothing in it:
image
- What is the performance profile measuring now -- is it just produced for one cost function, or does it concatenate all data together?  I think this needs to be made clear, either in the docs, or on the table page.
- I think there could be a bit more text before the comparison plot to explain what's going on -- the existing description is ambiguous, and we could do with spelling out exactly what the user is looking at.  This is the output for MGH10:
image
I wonder if putting the best fit in a solid colour, and the rest in a muted/transparent colour, might be a bit more clear?
- The best fits could do with rescaling the Y-axis (as is done in the summary plot).  For example, these plots (again from MGH10) are not the most helpful:
image

AndrewLister-STFC:
Some comments:

> A couple of cosmetic/functional comments (which may also be separate issues):
> 
>     * there's nothing in the docs to say that multiple cost functions can be run.  The same is true for, e.g., the Jacobians, so maybe there's wider issue to tidy up this part of the docs.

Good spot, I'll add a line in to each, although we should consider if we need a more heavy handed solution. Let me know if you think I should elaborate further once I've added it.
> 
>     * The index page still has a **Cost function** section, which now has nothing in it:
>       image

👍 

>     * What is the performance profile measuring now -- is it just produced for one cost function, or does it concatenate all data together?  I think this needs to be made clear, either in the docs, or on the table page.

It effectively treats each cost function as a separate problem. So with p problems and c cost functions, the plot would have c*p inputs per line. Perhaps this is something that should be added to the next ALC -- the ability to specify the columns, rows, categories so that different comparisons can be made?

For now I suggest we pick a sensible way of doing it and create an issue (I'm happy to change it if you think the most sensible way is having it just do 1 cost function or some other solution).

>     * I think there could be a bit more text before the comparison plot to explain what's going on -- the existing description is ambiguous, and we could do with spelling out exactly what the user is looking at.  This is the output for MGH10:
👍 
>       I wonder if putting the best fit in a solid colour, and the rest in a muted/transparent colour, might be a bit more clear?
👍 
>     * The best fits could do with rescaling the Y-axis (as is done in the summary plot).  For example, these plots (again from MGH10) are not the most helpful:

Is this a problem I introduced? The plotting code looks lke it should rescale them. Just to clarify it is correctly scaled in all plots but the best plot (the one with 1 line).

tyronerees:
Thanks for the changes.  

I don't think the plotting issue was introduced here -- it seems to be using the y-scaling from the initial guess (which is often unhelpful)....odd that the other plots all get it right!  Let's create another issue for it, unless it's easy to fix here?

For the performance profile, could you add a line to the performance profile page in the docs that says something like: 
> If 'n' tests are run with 'm' different cost functions, then this defines 'n*m' tests that are used to generate the performance profile.

(or something smoother).  It would be nice to have the behaviour here being an option, as you say, but for now I think it's enough to just document clearly which choice we've made.

Once this is added I'll approve/merge (🍾 🥳)

AndrewLister-STFC:
> I don't think the plotting issue was introduced here -- it seems to be using the y-scaling from the initial guess (which is often unhelpful)....odd that the other plots all get it right! Let's create another issue for it, unless it's easy to fix here?

I'm not sure, I think we might just have to recalculate the bounds and update. Maybe I'll create a new issue to avoind having to fix merge conflicts it ends up taking longer than expected.
 
> For the performance profile, could you add a line to the performance profile page in the docs that says something like:
> 
> > If 'n' tests are run with 'm' different cost functions, then this defines 'n*m' tests that are used to generate the performance profile.
> 
> (or something smoother). It would be nice to have the behaviour here being an option, as you say, but for now I think it's enough to just document clearly which choice we've made.

I couldn't find the docs, I was searching in the users bit rather than the concept. Will update.
 
> Once this is added I'll approve/merge (🍾 🥳)

🎉 🎉

tyronerees (approved):
All looks good **very** many thanks -- I know it was a beast!

PR #926, opened 2021-09-23
Add `results_dir` command line argument

### LINKED ISSUE #792: Results directory location setting
The location of the results directory is set by editing the options file.  Currently this is under `[PLOTTING]`, which is non-intuitive.

I think this should be moved, and potentially the best place for it would be in the command line call, as an optional flag.  At the very least, the instructions output:
image
should be more informative on where this option can be set.

robertapplin:
#### Description of Work
This PR changes adds the `results_dir` option in the options.ini file to be a command line argument. It also creates an `[OUTPUT]` section in the options file to put the existing `results_dir` option.

I'm assuming backwards compatibility won't need too much consideration as fitbenchmarking is still in minor versioning?

Fixes #792

#### Testing Instructions

1. Try variations of the following and make sure the relevant results directories are created and opened in a browser:
```
fitbenchmarking
```
```
fitbenchmarking -r my_results_folder/
```
```
fitbenchmarking -o examples/options_template.ini -r my_results_folder2/
```
```
fitbenchmarking -r ////invalid/folder\to/_cause_error/
```

2. Also try adding the `results_dir` option to the options file in the `[OUTPUT]` section

### DIFF
--- a/fitbenchmarking/utils/options.py
+++ b/fitbenchmarking/utils/options.py
@@ -141,8 +141,7 @@ class Options:
          'colour_ulim': 100,
          'cmap_range': [0.2, 0.8],
          'comparison_mode': 'both',
-         'table_type': ['acc', 'runtime', 'compare', 'local_min'],
-         'results_dir': 'fitbenchmarking_results'}
+         'table_type': ['acc', 'runtime', 'compare', 'local_min']}
     DEFAULT_LOGGING = \
         {'file_name': 'fitbenchmarking.log',
          'append': False,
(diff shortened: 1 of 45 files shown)


fitbenchmarking/utils/options.py
```diff
@@ -290,10 +290,6 @@ def results_dir(self):
         """
         return self._results_dir
 
-    @results_dir.setter
```
AndrewLister-STFC, line 293: I think this is here just to enforce the `os.path.abspath`. Is this unneccessary now?

If we are removing the setter, we could also remove the getter and access the `results_dir` directly rather than through a private variable.

AndrewLister-STFC (commented):
This looks pretty good to me, I do wonder if we want it in both the options (but not under "plotting") and the command line?

I wouldn't worry about backwards compatibility yet -- if someone complains we an always rename a release with a major version bump at this stage.

robertapplin:
@AndrewLister-STFC maybe we want an `OUTPUT` section and we can put the `results_dir` under there?

AndrewLister-STFC:
> 
> 
> @AndrewLister-STFC maybe we want an `OUTPUT` section and we can put the `results_dir` under there? (in addition to the cmd line argument of course)

Sounds like a plan :)

robertapplin:
This should now be ready for more testing @AndrewLister-STFC

AndrewLister-STFC (approved):
All good, I'll update ad merge

PR #928, opened 2021-09-27
Fix MaxRuntimeError wrong flag set

### LINKED ISSUE #927: Correct flag for MaxRuntimeError is not set when using Mantid
**Description of the error**
When the max runtime is reached using a mantid minimizer, the wrong flag is returned (3 is returned instead of 6).

**Suggested fix**
The MaxRuntimeError can sometimes be caught by the fitting software and re-raised as an ordinary Exception. So this error is caught by the 'Exception' clause and flag 3 is set.

To fix, we can check the exception message and if it contains the MaxRuntime error message then we can set the flag as 6.

robertapplin:
#### Description of Work
This PR fixes a problem where the wrong flag was being set for a MaxRuntimeError when using certain software packages (such as mantid) when benchmarking.

This PR also makes it so that nan and inf results are now shown in the comparison table even when it is nan and inf for all minimizers trying to solve a given problem. I think this is a good idea because I found it confusing why certain results were not appearing as expected without a clear cause or warning.

Fixes #927

#### Testing Instructions

1. I fitted the SASView simple shape problems using the following options file
```
[FITTING]

software: mantid

num_runs: 1
max_runtime: 5

[MINIMIZERS]

bumps: lm-bumps
       mp

scipy_ls: lm-scipy

mantid: BFGS
        Levenberg-Marquardt

[PLOTTING]

make_plots: yes

[LOGGING]

external_output: log_only
```
2. The flag next to the two infs should be 6
image

3. If you comment out `Levenberg-Marquardt` in the options file and run again, you should still be able see them in the comparison table when you open the browser.
image

### DIFF
--- a/docs/source/users/options/minimizer_option.rst
+++ b/docs/source/users/options/minimizer_option.rst
@@ -231,7 +231,7 @@ It has support for a number of minimizers, most of which are from GSL.
 
 * `Levenberg-Marquardt MD <https://docs.mantidproject.org/nightly/fitting/fitminimizers/LevenbergMarquardtMD.html>`__ (:code:`Levenberg-MarquardtMD`) - An implementation of Levenberg-Marquardt intended for MD workspaces, where work is divided into chunks to achieve a greater efficiency for a large number of data points.
 
-* `Simplex <https://docs.mantidproject.org/nightly/fitting/fitminimizers/Simplex.html>`__ (:code:`simplex`)
+* `Simplex <https://docs.mantidproject.org/nightly/fitting/fitminimizers/Simplex.html>`__ (:code:`Simplex`)
 
 * `SteepestDescent <https://docs.mantidproject.org/nightly/fitting/fitminimizers/GradientDescent.html>`__ (:code:`SteepestDescent`)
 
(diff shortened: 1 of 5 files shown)


robertapplin:
I think the `nan` makes sense for the second image in the PR description because the relative performance cannot be calculated when there's a single minimizer

AndrewLister-STFC:
> 
> 
> I think the `nan` makes sense for the second image in the PR description because the relative performance cannot be calculated when there's a single minimizer

Using `nan` could be an issue if the black colour clashes with a mapping someone is using, since it will always be black. I'm not sure if it's best to edit the colour mapping or use a safer divide in the result class.

robertapplin:
> > I think the `nan` makes sense for the second image in the PR description because the relative performance cannot be calculated when there's a single minimizer
> 
> Using `nan` could be an issue if the black colour clashes with a mapping someone is using, since it will always be black. I'm not sure if it's best to edit the colour mapping or use a safer divide in the result class.

I see, that could be a problem. It would be good to discuss this with you tomorrow

robertapplin:
After talking with @AndrewLister-STFC , we decided using inf instead of nan would also be ok (this can be changed in `fitbm_result`)

robertapplin:
I ended up changing the colour of `nans` so that they use the same colour as `infs` because it was simpler and is still a good fix. 
image

Its ready for another look @AndrewLister-STFC

AndrewLister-STFC (changes requested):

robertapplin:
Thanks @AndrewLister-STFC , its ready again

AndrewLister-STFC (approved):

PR #931, opened 2021-09-28
Use `eval_model` in bumps controller so that MaxRuntime timer is checked

### LINKED ISSUE #930: MaxRuntime option doesn't work for Bumps fitting
**Description of the error**
The MaxRuntimeError is not raised for bumps until it finishes fitting because it doesn't use `eval_model` when doing the fitting. This means you still have to wait a long time for fitting to finish.

**Suggested fix**
Look at the bumps controller and see if we can inject the checking of the MaxRuntime timer somewhere here 

https://github.com/fitbenchmarking/fitbenchmarking/blob/da40b2fa8cfdf115c06cc4f19eb66f6c264c634e/fitbenchmarking/controllers/bumps_controller.py#L68

**To replicate**
`fitbenchmarking -o examples/options_template.ini -p examples/benchmark_problems/SAS_modelling/SASView_Complex_Shapes_1D/`

```
[FITTING]

software: mantid
       bumps
       scipy_ls

num_runs: 1
max_runtime: 2

[MINIMIZERS]

bumps: lm-bumps
       mp

scipy_ls: lm-scipy

mantid:  Simplex
        Levenberg-Marquardt

[PLOTTING]

make_plots: yes

[LOGGING]

external_output: log_only
# Note: default for external_output is log_only, meaning information is
# printed to log file but not the stdout stream. For more information
# see the fitbenchmarking docs
```

robertapplin:
#### Description of Work
This PR ensures the bumps controller uses the `eval_model` method instead of the `function` method so that the MaxRuntime timer gets checked will fitting is taking place. Previously this was not happening, and so bumps would never time out until the fitting calculation had finished (which could take a long time).

Having skimmed through the other controllers, I think this is a bumps-only issue.

Fixes #930

#### Testing Instructions

1. Follow the instructions in the issue. The max runtime timeout should happen in roughly the expected amount of time.

2. Its worth having a quick check of the other controllers to make sure `eval_model` is called during their fitting.

### DIFF
--- a/docs/source/users/options/index.rst
+++ b/docs/source/users/options/index.rst
@@ -17,6 +17,7 @@ and how to override them, are given in the pages below.
     minimizer_option
     jacobian_option
     plotting_option
+    output_option
     logging_option
 
 
--- a/docs/source/users/options/minimizer_option.rst
+++ b/docs/source/users/options/minimizer_option.rst
@@ -16,6 +16,8 @@ the default list of minimizers unless otherwise stated.
    software is also set in :ref:`fitting_option` (either explicitly, or
    as a default option).
 
+.. _bumps:
+
 Bumps (:code:`bumps`)
 =====================
 
@@ -60,6 +62,7 @@ The Bumps minimizers are set as follows:
    `de` is not included in the default list of minimizers for bumps. To run this solver, you must
    explicitly set the minimizer as seen above.
 	   
+.. _dfo:
 
 DFO (``dfo``)
 =============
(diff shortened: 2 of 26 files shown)


AndrewLister-STFC:
I've just ran this and it works but isn't the full story -- it looks like the error causes bumps to fail "gracefully" and return something that's not flagged as a timeout. I'm happy to merge this and make a new issue for that though, lets discus tomorrow

robertapplin:
> I've just ran this and it works but isn't the full story -- it looks like the error causes bumps to fail "gracefully" and return something that's not flagged as a timeout. I'm happy to merge this and make a new issue for that though, lets discus tomorrow

Might that be fixed by #928 , or do you think its a different problem?

AndrewLister-STFC:
#933 has been opened to cover the new bug this highlights.

AndrewLister-STFC (approved):

PR #934, opened 2021-10-01
Refactor FitBenchmarkParser

### LINKED ISSUE #890: Refactor the native parser
We should refactor the native parser to have a subclass for each software that feeds into it.

robertapplin:
#### Description of Work
This PR refactors the `FitBenchmarkParser` by creating three new parsers that inherit from it called `SASViewParser`, `MantidParser` and `ScipyIVPParser`. This will make it easier to add new parsers in the future without expanding the `FitBenchmarkParser` further.

Fixes #890

#### Testing Instructions

1.
2.
3.

### DIFF
--- a/fitbenchmarking/parsing/ivp_parser.py
+++ b/fitbenchmarking/parsing/ivp_parser.py
@@ -1,0 +1,85 @@
+"""
+This file implements a parser for the Scipy IVP data format.
+"""
+import importlib
+import inspect
+import os
+import sys
+import numpy as np
+from collections import OrderedDict
+
+from fitbenchmarking.parsing.fitbenchmark_parser import FitbenchmarkParser
+from fitbenchmarking.utils.exceptions import ParsingError
+
+import_success = {}
+try:
+    from scipy.integrate import solve_ivp
+    import_success['ivp'] = (True, None)
+except ImportError as ex:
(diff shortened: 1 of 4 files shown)


fitbenchmarking/parsing/ivp_parser.py
```diff
@@ -0,0 +1,86 @@
...
+
+from fitbenchmarking.parsing.fitbenchmark_parser import FitbenchmarkParser
+from fitbenchmarking.utils.exceptions import ParsingError
+
+import_success = {}
+try:
+    from scipy.integrate import solve_ivp
+    import_success['ivp'] = (True, None)
+except ImportError as ex:
+    import_success['ivp'] = (False, ex)
```
AndrewLister-STFC, line 20: We don't need this as a dictionary anymore. In fact this was only here as we couldn't assume the parser would fail based on 1 import exception. With the new implementation, given that the parser is imported only after it is selected we can just let it raise the import error and let the factory handle it (line 61 of the factory).

robertapplin:
Thanks, it should now be up-to-date with your review so far

AndrewLister-STFC (changes requested):
This looks much better than the 1 massive file! I have a couple of nitpicky points but feel free to push back if you disagree.

PR #937, opened 2021-10-07
Add `validate` method to base controller to check compatibility for `cs` jacobian

### LINKED ISSUE #535: `cs` fails on some problems
**Description of the error**
If you run the software with `jac_method=SciPyFD` and `num_method=cs`, then all but the NIST examples will fall over, due to a function that expects a real being given a complex value.

For example, running `simple_tests` with `scipy-lm` gives:

```
        Software: SCIPY
            Minimizer: lm-scipy
Python argument types in
    IFunction.setParameter(IFunction, int, numpy.complex128)
did not match C++ signature:
    setParameter(Mantid::API::IFunction {lvalue} self, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > name, double value)
    setParameter(Mantid::API::IFunction {lvalue} self, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > name, double value, bool explicitlySet)
    setParameter(Mantid::API::IFunction {lvalue} self, unsigned long i, double value)
    setParameter(Mantid::API::IFunction {lvalue} self, unsigned long i, double value, bool explicitlySet)

```

This behaviour should at least be documented and handled gracefully, and ideally fixed so that `cs` can be used for more data sources.

robertapplin:
#### Description of Work
This PR adds a validate method to the base controller so that the compatibility of certain options and problem sets can be checked before running a benchmark combination.

This would be a good place to add future compatibility checks that might be needed in the future.

There is a `validate_minimizer` function in the base controller that could probably be made private and moved to the same place as the `_validate_jacobian` function call. The validation in the `setup` function could also be moved here in the future.

Fixes #535

#### Testing Instructions

1. Use the following options file
```
[MINIMIZERS]
scipy_ls: lm-scipy

[JACOBIAN]
scipy: cs

[FITTING]
num_runs: 1
software: scipy_ls
jac_method: scipy
```
with `fitbenchmarking -o examples/options_template.ini -p examples/benchmark_problems/simple_tests/`
2. Check that the messages output when running the mantid problem sets are clear.


fitbenchmarking/controllers/base_controller.py
```diff
@@ -180,6 +184,13 @@ def software(self):
                 self._software = self.__class__.__name__[:-10].lower()
         return self._software
 
+    def validate(self) -> None:
```
AndrewLister-STFC, line 187: I'm not sure I understand why this has 2 layers (`validate`, `_validate_jacobian`). Is it just pre-empting further validation?

fitbenchmarking/controllers/base_controller.py
```diff
@@ -21,6 +22,9 @@ class Controller:
 
     VALID_FLAGS = [0, 1, 2, 3, 4, 5, 6]
 
+    # Invalid problem formats for specific jacobian methods
+    INVALID_JACOBIAN_FORMATS = {"cs": ["mantid"]}
```
AndrewLister-STFC, line 26: With this being in the base class I'm concerned about name clashes. E.g. if another jacobian software is added with a method named "cs".

AndrewLister-STFC: I like the new implementation with the attributes on the relevent controller but the problem of having the same method on multiple jacobian classes could still be an issue

fitbenchmarking/controllers/tests/test_controllers.py
```diff
@@ -567,6 +568,74 @@ def test_mantid(self):
...
+        controller.minimizer = "L-BFGS-B"
+        controller.jacobian = self.jac
+
+        with self.assertRaises(exceptions.IncompatibleJacobianError):
+            controller.validate()
+
+    def test_controller_will_not_raise_for_compatible_jacobian(self):
+        """
+        ScipyController: Test that the Scipy controller validation
+        will raise for an incompatible Jacobians
```
AndrewLister-STFC, line 627:
```suggestion
        will not raise for a compatible Jacobians
```

PR #941, opened 2021-10-13
Ensure Levmar Jacobians are contiguous in memory

### LINKED ISSUE #776: Levmar performance

robertapplin:
#### Description of Work
This PR fixes a bug where NIST problems supplied to the Levmar fitting package would have bad fits when providing it our own Analytic or Scipy Jacobians. Levmar's default jacobian would work as expected.

I compared the current implementation of the analytic jacobian in fit benchmarking (which didn't work) to a different implementation of the same jacobian (which did work) for the DanWood problem definition file. See the script I used
[here](https://github.com/fitbenchmarking/fitbenchmarking/files/7337564/danwood.zip)

The current implementation `[1]` was:
```
jac = np.array([x ** p[1], p[0] * x ** p[1] * np.log(x)]).T
jac = - jac / e[:, None]
```
The working jacobian `[2]` was:
```
jac = np.empty((x.shape[0], 2))
jac[:, 0] = np.power(x, p[1])
jac[:, 1] = p[0] * np.power( x, p[1] ) * np.log(x)
jac = - jac / e[:, None]
```

I found that these two jacobians would return a np array which had:
- The exact same dimensions and shape
- The exact same data
- The exact same data types inside the array (float64)
- The exact same number of bytes in memory were used for both of them

However, `[1]` gave a poor fit, and `[2]` gave a good fit. Confusing! 

I figured out that the transpose operation in `[1]` was causing the difference and then realised that this operation would make the np array non-contiguous in memory. So I then tried making the jacobian in `[1]` contiguous in memory by wrapping it in `np.ascontiguousarray` and this fixed the issue!

The current theory is that Levmar is doing some sort of memory operation (as it is written in C/C++) which assumes the jacobian provided to it returns an array that is contiguous in memory.

Note that the transpose operation for the analytic jacobian is seen here https://github.com/fitbenchmarking/fitbenchmarking/blob/402ee06e951497e3452e74e9b445f1f1c41413b0/fitbenchmarking/parsing/nist_data_functions.py#L72

Fixes #776

#### Testing Instructions

See issue for testing instructions


fitbenchmarking/controllers/levmar_controller.py
```diff
@@ -90,6 +90,22 @@ def fit(self):
         # self._info[5] = number of `jacf` evaluations
         # self._info[6] = number of linear system solved
 
+    def eval_wrapper(self, params, **kwargs):
```
AndrewLister-STFC, line 93: Can you rename this to be something with "j" or "jac" or "jacobian" in the name to make it a bit clearer what it's evaluating. I think this can also be a protected/private function with an underscore.

PR #966, opened 2021-11-02
Specify minimizers in regression tests

### LINKED ISSUE #946: Change regression tests to test against *working* minimizers
**Description of the error**
Current regression tests test against the first minimizer (alphabetically). While this is easier to extend, it is not robust at spotting breaking changes!

Any change in the regression tests is easy to see if it is trivial or breaking.

The tests can change by seemingly insignificant values when the change breaks better minimizers.

robertapplin:
#### Description of Work
This PR ensures that we choose specific minimizers to test against for each software (i.e. the most-used minimizers rather than just the first alphabetical minimizer).

Fixes #946

#### Testing Instructions

1. Make sure the regression tests pass, and that the new results look sensible.


tyronerees (approved):
This all looks great -- many thanks!

PR #968, opened 2021-11-03
Add `minuit` and `dfo` to the regression tests

### LINKED ISSUE #939: Extend the regression tests
The regression tests should be extended so that they test all deterministic fitting packages supported by fitbenchmarking. The packages currently missing from the `all` test type are:
- [ ] `dfo`
- [ ] `gradient_free` (requires bounds)
- [ ] #968
- [ ] `scipy_go` (requires bounds)
- [ ] The `matlab` regression tests aren't currently being run because of licensing issues

robertapplin:
#### Description of Work
This PR adds the `minuit` software packages back into the regression tests. The `gradient_free` and `scipy_go` packages will not be added in this because they require bounds to be specified for their parameters.

The `dfo` and `minuit` packages have seen unreliable failures in the past, hence why they were removed, so I will try investigate why this is happening.

#### Investigation
After some investigation I found that running a minuit fit on only **one** of the unreliable problems would result in the same unreliable behaviour as previously, so the issue isn't caused by any 'interference' between instances of the `Minuit` python class. There were consistently two different answers appearing for the cost function of the `BENNETS5.SIF` dataset of `A = 2.113e-5` and `B = 2.477e-5`. However I could never replicate this unreliable issue whilst on the same machine during the same session (i.e. repeating the test multiple times did not result in different results either locally or remotely).

When you set the `print_level` to 2 or 3 you get more information about what is going on in the fit. It appears there is an `MnSeedGenerator` object created right at the beginning, and after this the calculation values for `A` and `B` begin to diverge slightly. By the end of the fitting process, the differences are large enough for the stopping condition to be triggered at a different iteration. The stopping condition is the `EDM` or `Estimated Distance to Minimum` which has a default value of 0.0002 for Least Squares (but migrad can violate this by a factor of 10):
https://github.com/scikit-hep/iminuit/blob/23be056e6d25b1c27df6f0529d5d050e5ed8fb2f/src/iminuit/minuit.py#L165

If you reduce the 'tolerance' value then you can decrease the `EDM` max and get more accurate results. Setting this to zero meant the minimum possible tolerance was used, and the results achieved for the `BENNETS5.SIF` dataset became reliable. However, the results for other datasets which had cost function values which were much smaller (~1e-16) were still unreliable.

The best theory I have for what is happening is that the `MnSeedGenerator` is causing very small differences in the calculations early on resulting in different cost function results by the end of the fit. After reading some of the Minuit documentation, it makes me think that we should be choosing our datasets carefully for Migrad fitting. The solution **might** be to have a category of regression test where the problems are simple (e.g. only one local minimum). Maybe the category could be called `Deterministic` or `Simple` or something similar.

#### Other things I tried
I ruled out a few things while investigating:
- [x] There was no interference between instances of the `Minuit` class (which I thought could be possible as it creates a python wrapper around a c++ `MnMigrad` object which I thought could potential hang around in memory and be reused by chance in a subsequent iteration).
- [x] Changing the strategy to 'Careful' didn't work well 
https://github.com/scikit-hep/iminuit/blob/23be056e6d25b1c27df6f0529d5d050e5ed8fb2f/src/iminuit/minuit.py#L180
- [x] I checked there wasn't a precision mis-match when doing the calculations 
https://github.com/scikit-hep/iminuit/blob/23be056e6d25b1c27df6f0529d5d050e5ed8fb2f/src/iminuit/minuit.py#L121
- [x] Tested for thread safety. The function provided to minuit needs to be thread safe when OpenMP is being used for calculations. We are using MPI so should be ok, but I checked anyway and I don't think this is the problem

Fixes #939

#### Testing Instructions

1. Make sure the regression tests pass and the results look sensible.


robertapplin:
I have tracked down the source of the unreliable results to a block of code in iminuit which is parallelized. I did this by first looking at the differing outputs at the start of the fit (using `print_level = 3`):

**A**
image

**B**
image

If you follow the print statements in the output, you will find yourself in the `MnSeedGenerator`

and on line ... it calculates the initial gradient (which matches for both) before calculuting

robertapplin:
These slides talk about some of the parallelization work done in Minuit2 https://indico.cern.ch/event/36763/contributions/871192/attachments/730869/1002804/ParallelMinuit_27Jun.pdf

robertapplin:
It seems that slight rounding differences across machines might be causing the small numerical differences in the calculations. I am going to try to reduce the precision of the calculation to see if the calculations become more reliable. The machine currently uses double precision (16 decimal places), so I will try limit precision to 14 or 15 decimal places.

robertapplin:
Changing the precision didn't prevent all numerical instabilities. I think the solution is to only test minuit with problem sets that give a numerically stable solution. I have started to do a bit of a refactor in `test_regression` to make it easier to do this.

@AndrewLister-STFC I have copied the numerically stable datasets into a separate directory called `minuit_set` so I can use them for testing minuit. But having the same data in two different locations seems bad - any ideas how we could do this better?

fitbenchmarking/systests/test_regression.py
```diff
@@ -240,12 +154,44 @@ def diff_result(actual, expected):
...
+    Compares the expected benchmark results with the actual results,
+    and returns the lines which differ between expected and actual
+    along with a formatted message.
+
+    :param problem_sub_directory: The directory containing problems.
+    :type problem_sub_directory: str
+    :param result_filename: The name of the actual result file.
+    :type result_filename: str
+    :return: The lines which differ and a formatted message
+    :rtype: list of list of strings and str
```
AndrewLister-STFC, line 168: I think this may be clearer as a more concise description
```suggestion
    :rtype: list[list[str]], str
```

fitbenchmarking/systests/test_regression.py
```diff
@@ -240,12 +154,44 @@ def diff_result(actual, expected):
...
+    """
+    expected_file = os.path.join(os.path.dirname(__file__),
+                                 '{}_expected_results'.format(platform),
```
AndrewLister-STFC, line 171:
```suggestion
                                 f'{platform}_expected_results',
```

AndrewLister-STFC (changes requested):
All looks good to me bar some very minor suggestions, I'm surprised that the cubic needed to be swapped out, but I don't think it matters.

AndrewLister-STFC (approved):

PR #977, opened 2021-11-18
Fix Mantid MultiFit results when using more than one Jacobian/Hessian

### LINKED ISSUE #920: Review handling of mantid multifit in core fitting routines
**Description of the error**
Looking at the handling of the chi_sq in `fitbenchmarking.core.fitting_benchmarking.loop_over_[minimizers|jacobians|hessians]()` I think this may have been broken at some point.

These combinations should all be linked to the same datatsets

I expect that if we run this with multiple jacobians and or hessians we will see each result for these combinations is treated as a new dataset.

**Suggested fix**
The block starting:
```
            for result in results:
                if problem.multifit:
```
Should be moved to the `loop_over_hessians()` function.

**Additional details**#
We should probably expand our testing when fixing this to cover this case.

robertapplin:
#### Description of Work
This PR fixes an issue where the results for a Mantid MultiFit would not be created when the multi fit is done using a number of different jacobians or hessians. There used to be an exceptions saying the program was exiting due to unknown exception, but now it opens the browser and displays the results as expected.

Fixes #920

#### Testing Instructions

1. Use the options file below to run the following command 
`fitbenchmarking -p examples/benchmark_problems/MultiFit/ -o examples/test_multi_fit.ini`

```
[FITTING]

software: mantid

num_runs: 1
max_runtime: 600

[MINIMIZERS]

mantid: Levenberg-Marquardt

[JACOBIAN]
scipy: 2-point
       3-point

[PLOTTING]

make_plots: yes

[OUTPUT]

results_dir: new_results/

[LOGGING]

external_output: log_only
```

2. You should get something like this

image

### DIFF
--- a/fitbenchmarking/core/fitting_benchmarking.py
+++ b/fitbenchmarking/core/fitting_benchmarking.py
@@ -291,7 +291,6 @@ def loop_over_fitting_software(cost_func, options, start_values_index,
         unselected_minimizers[s] = minimizer_failed
         minimizer_dict[s] = new_minimizer_list
         results.extend(problem_result)
-
     return results, unselected_minimizers, minimizer_dict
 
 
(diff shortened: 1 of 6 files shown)


fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -375,29 +374,14 @@ def loop_over_minimizers(controller, minimizers, options, grabbed_output):
...
-                        individual_result = \
-                            fitbm_result.FittingResult(**result)
-                        results_problem.append(individual_result)
-                else:
-                    # Normal fitting
-                    individual_result = fitbm_result.FittingResult(
-                        **result)
-                    results_problem.append(individual_result)
-            new_minimizer_list.extend(minimizer_list)
+                results_problem.append(fitbm_result.FittingResult(**result))
```
AndrewLister-STFC, line 382: Is there a reason for keeping the object initialization here rather than moving it with the other code?

AndrewLister-STFC (commented):
This looks good, I have 1 comment

AndrewLister-STFC:
Also let me know if there's an order you'd rather the regression tests are merged in!

fitbenchmarking/core/fitting_benchmarking.py
robertapplin: There was no reason, I'll move it down to where the new code is 👍

robertapplin:
> Also let me know if there's an order you'd rather the regression tests are merged in!

I think it would be better to get #968 merged in first

fitbenchmarking/core/tests/test_fitting_benchmarking_minimizers.py
```diff
@@ -133,8 +133,7 @@ def test_run_minimzers_selected(self, loop_over_hessians):
         Tests that some minimizers are selected
         """
         self.options.algorithm_type = ["general"]
-        self.results = [[self.result_args]]
```
AndrewLister-STFC, line 136: I don't think we need result_args to be a dict anymore? Can you change it to just creating the fitresults object in the setup?

AndrewLister-STFC (changes requested):
As expected there was a conflict with the other PR that touched the regression tests. Once thats fixed I think this will be good to go.

AndrewLister-STFC (approved):

PR #978, opened 2021-11-18
Minor updates to documentation

### LINKED ISSUE #973: typo in error output
If an incorrect option is given in the output file, there's a typo in the warning message: 
image

### LINKED ISSUE #974: Error in hessian options docs
The docs say that `hes_method` can be one of four options:
image
but the code says only `default` and `analytic` are supported:
image

robertapplin:
#### Description of Work
This PR makes some minor updates to the documentation.

Fixes #974
Fixes #973 

#### Testing Instructions

1. Read the docs and make sure they make sense


docs/source/users/options/fitting_option.rst
```diff
@@ -134,8 +134,6 @@ Current Hessian methods are:
 
 * ``default`` - Hessian information is not passed to minimizers
 * ``analytic`` - uses the analytic Hessian extracted from the fitting problem.
-* ``scipy`` -  uses :ref:`SciPy's finite difference approximations <scipy-hes>`.
-* ``numdifftools`` - uses the python package :ref:`numdifftools <numdifftools-hes>`.
```
AndrewLister-STFC, line 138: I can't recreate the issue. I don't think we should remove this until we can confirm it's not just a test on an outdated version. Can you run it?

AndrewLister-STFC (changes requested):
The f-strings are much neater and fix the typo! But the hessians seems to be a missunderstanding...

docs/source/users/options/fitting_option.rst
robertapplin: Looks like you are right, thanks for finding that 👍 I'll add it back in

robertapplin: It runs fine for me too

AndrewLister-STFC (approved):

PR #979, opened 2021-11-18
802: Fail gracefully on mantid/matlab combination

### LINKED ISSUE #802: Error with mantid problems and matlab fitting
**Description of the error**
When running Mantid problems with Matlab fitting software, the serialization of `self.cost_func.eval_model` in the Matlab Controller causes the following error:
`Pickling of "_api.CompositeFunction" instances is not enabled (http://www.boost.org/libs/python/doc/v2/pickle.html)`
As a result, all Mantid problems are skipped over.

Mantid problems should be able to be run with matlab fitting software selected in options.

No results produced for Mantid problems when run with matlab fitting software

**Suggested fix**
There should either be a workaround for having to pickle `self.cost_func.eval_model` to get it to work with matlab, or the function could be somehow modified to make it picklable.

AndrewLister-STFC:
#### Description of Work
Adds a validation step to matlab controllers which checks if the problem came from mantid.
Also refactors the validation check slightly to be more easily extensible.

Docs already seem to be in place (docs/source/users/notes.rst).
Possibly need to add tests

Fixes #802 

#### Testing Instructions

Run any of the matlab controllers with a mantid problem. e.g. running the Muon problem set with the following options

```
[FITTING]

software: matlab_curve

cost_func_type: nlls
num_runs: 1

[PLOTTING]

make_plots: no

[LOGGING]

external_output: debug
```


fitbenchmarking/controllers/matlab_mixin.py
```diff
@@ -53,3 +56,19 @@ def py_to_mat(func, eng):
...
+        self._validate_problem_format()
+        super().validate()
+
+    def _validate_problem_format(self):
+        """
+        Validates that the problem format is compatible with matlab controllers
+        """
+        if self.problem.format == 'mantid':
+            raise IncompatibleProblemError(
+                'Mantid problems cannot be used with MATLAB controllers.')
```
robertapplin, line 74: Would it be possible to move this to the base controller and get it to do something similar to `_validate_jacobian` so that this method wouldn't need reimplementing in another controller if we need similar validation in the future?

So for instance could we have an empty list of `INVALID_PROBLEMS` in the base controller, and then in the matlab controller we could have `INVALID_PROBLEMS = ["mantid"]`. Then we could check if the `self.problem.format` is in the `INVALID_PROBLEMS`? A bit like what is done for the jacobians:
https://github.com/fitbenchmarking/fitbenchmarking/blob/780eeaa485480f2bede7ca5eaf7763ddc4240c8c/fitbenchmarking/jacobian/base_jacobian.py#L15

https://github.com/fitbenchmarking/fitbenchmarking/blob/780eeaa485480f2bede7ca5eaf7763ddc4240c8c/fitbenchmarking/jacobian/scipy_jacobian.py#L17

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -625,14 +624,15 @@ def perform_fit(controller, options, grabbed_output):
...
+        LOGGER.warning(str(ex))
+        controller.flag = 7
     except Exception as ex:  # pylint: disable=broad-except
         LOGGER.warning(str(ex))
 
         # Note: Handle all exceptions as general exception to cover case
         #       where software re-raises our exception as a new type.
-        error_flags = {MaxRuntimeError: 6,
-                       IncompatibleJacobianError: 7,
-                       IncompatibleHessianError: 7}
```
robertapplin, line 635: Just want to check that the validation still works described in here https://github.com/fitbenchmarking/fitbenchmarking/pull/937 

because I had a feeling the validation exception was being caught and re-raised as a different error?

robertapplin (changes requested):

fitbenchmarking/core/fitting_benchmarking.py
AndrewLister-STFC: I think this should raise any validation exception before ~software~ *bumps* gets to run, so it won't be re-raised as something else.

fitbenchmarking/controllers/matlab_mixin.py
AndrewLister-STFC: Absolutely, I'll update

fitbenchmarking/controllers/base_controller.py
```diff
@@ -261,6 +268,15 @@ def _validate_hessian(self) -> None:
                           f"'{self.problem.format}'."
                 raise IncompatibleHessianError(message)
 
+    def _validate_problem_format(self):
+        """
+        Validates that the problem format is compatible with matlab controllers
```
AndrewLister-STFC, line 273:
```suggestion
        Validates that the problem format is compatible with the controller
```

fitbenchmarking/controllers/matlab_mixin.py
```diff
@@ -6,14 +6,17 @@
...
+
 try:
     import dill
     import_success = True
 except ImportError:
     import_success = False
 
-from fitbenchmarking.utils.exceptions import MissingSoftwareError
+from fitbenchmarking.utils.exceptions import (IncompatibleProblemError,
+                                              MissingSoftwareError)
```
AndrewLister-STFC, line 19:
```suggestion
from fitbenchmarking.utils.exceptions import MissingSoftwareError
```

fitbenchmarking/controllers/tests/test_controllers.py
```diff
@@ -907,6 +907,23 @@ def test_py_to_mat(self):
...
+        """
+        MatlabController: Tests for correct error when fitting mantid problem
+        """
+        # No raise for default (NIST) problem
+        controller = MatlabController(self.cost_func)
+        controller.validate()
+        # Raise for Mantid problem
+        cost_func = make_cost_func('cubic-fba-test-go.txt')
+        problem = cost_func.problem
+        jac = Scipy(cost_func.problem)
```
AndrewLister-STFC, line 920:
```suggestion
        jac = Scipy(cost_func.problem)
```

fitbenchmarking/controllers/matlab_mixin.py
```diff
@@ -36,6 +38,7 @@ def __init__(self, cost_func):
                                        'fitting, module "dill" is required.')
 
         self.initial_params_mat = None
+        self.incompatible_minimizers.append('mantid')
```
robertapplin, line 41: Should this be `incompatible_problems`? And also I noticed MatlabMixin doesn't inherit from the base controller, so maybe this would be better off as a class variable in the `MatlabController`, `MatlabStatController`, etc. classes. Unless there is a reason for it to go here?

AndrewLister-STFC: Good spot! I put a test in but forgot the matlab ones aren't running yet. I put it here as it covers all of the matlab based ones, so if we add more it's one less thing to get right. That said I'm happy to move it if you think it would be clearer.

The way the inheritance works, is that the MatlabController will inherit from the BaseController, and then from the MatlabMixin, so provided the mixin is used correctly it should be the same functionality.

robertapplin: I see, so would this only work if the `BaseController` and `MatlabMixin` is inherited in the correct order? I'm thinking it might be safer to have `incompatible_problems` separately in each Matlab Controller class, but having said that I have no great preference so I'm happy with this if if you are

AndrewLister-STFC: Yes, that's true although I think the initialiser might not work if they're out of order, so we do need them to be in the correct order. Either way I'll move it to the individual controllers to make it more obvious

AndrewLister-STFC:
With the example in the description I get:
```
eel23943@al-matlab-testing:~/fitbenchmarking$ python fitbenchmarking/cli/main.py -p examples/benchmark_problems/Muon/ -o opts.ini
Running the benchmarking on the Muon problem set
Parsing problems
FrameworkManager-[Notice] Welcome to Mantid 6.0.0
FrameworkManager-[Notice] Please cite: http://dx.doi.org/10.1016/j.nima.2014.07.029 and this release: http://dx.doi.org/10.5286/Software/Mantid6.0
No protocol specified
Running problems

########################################################
 Running data from: Muon_67068_coadd_67108_long.txt 1/4
########################################################
    Starting value: 1/1
        Software: MATLAB_CURVE
            Minimizer: Levenberg-Marquardt
The selected software can not be used with the given problem.
Details: mantid problems cannot be used with matlab_curve controllers.
            Minimizer: Trust-Region
The selected software can not be used with the given problem.
Details: mantid problems cannot be used with matlab_curve controllers.

###########################################
 Running data from: Muon_EMU_73673.txt 2/4
###########################################
    Starting value: 1/1
        Software: MATLAB_CURVE
            Minimizer: Levenberg-Marquardt
The selected software can not be used with the given problem.
Details: mantid problems cannot be used with matlab_curve controllers.
            Minimizer: Trust-Region
The selected software can not be used with the given problem.
Details: mantid problems cannot be used with matlab_curve controllers.

#############################################
 Running data from: Muon_HIFI_113856.txt 3/4
#############################################
    Starting value: 1/1
        Software: MATLAB_CURVE
            Minimizer: Levenberg-Marquardt
The selected software can not be used with the given problem.
Details: mantid problems cannot be used with matlab_curve controllers.
            Minimizer: Trust-Region
The selected software can not be used with the given problem.
Details: mantid problems cannot be used with matlab_curve controllers.

#############################################
 Running data from: Muon_HIFI_116891.txt 4/4
#############################################
    Starting value: 1/1
        Software: MATLAB_CURVE
            Minimizer: Levenberg-Marquardt
The selected software can not be used with the given problem.
Details: mantid problems cannot be used with matlab_curve controllers.
            Minimizer: Trust-Region
The selected software can not be used with the given problem.
Details: mantid problems cannot be used with matlab_curve controllers.
Producing output for the Muon problem set
invalid value encountered in subtract
invalid value encountered in subtract
invalid value encountered in subtract
Completed benchmarking for examples/benchmark_problems/Muon/ problem set

INFO:
The FitBenchmarking results will be placed into the folder:

   /home/eel23943/fitbenchmarking/fitbenchmarking_results

To change this use the -r or --results-dir optional command line argument. You can also set 'results_dir' in an options file.

WARNING:
The browser failed to open automatically. Copy and paste the following url into your browser:

   /home/eel23943/fitbenchmarking/fitbenchmarking_results/results_index.html
   ```

robertapplin (approved):
Looks good to me from a code standpoint, and the new warning message in your screenshot is good 👍 I'll approve this but I won't merge. I haven't done any functional testing myself as I don't have matlab configured

robertapplin:
If you are happy for this to be merged based on it working for you then let me know

AndrewLister-STFC:
I think I'm the only one of us with a matlab ready machine 😅
Lets chat at the stand up. I can give access to the machine, we can test it with Tryone's test machine if he gets it working, or I think it should be fine to merge.

PR #984, opened 2021-11-22
Fix default jacobians for `gsl`, `mantid`, `scipy - Newton CG` and `RALFit`

### LINKED ISSUE #980: Default Jacobians not working
The `jacobian` option `default` doesn't work for all minimizers.  It seems to fail for:
- `gsl`
- `mantid`
- `ralfit`
- `scipy` (`Newton-CG` only)

Here's an ini file that should highlight the issues:
```
[FITTING]
software: gsl
          mantid
          ralfit
          scipy

jac_method: analytic
	    scipy
	    default
	    numdifftools

hes_method: default
	    analytic	

cost_func_type: weighted_nlls
```
Here is the expected behaviour (from [the docs](https://fitbenchmarking.readthedocs.io/en/latest/users/options/jacobian_option.html#solver-default-jacobian-default)):
image

robertapplin:
#### Description of Work
This PR fixes the default jacobian for `gsl`, `mantid`, `scipy - Newton CG` and `RALFit`.

`RALFit`, `gsl` and `scipy` could be easily fixed by getting the default jacobian to inherit from the `Scipy` jacobian and setting the method to `2-point`. In my mind this makes sense because the documentation says the Scipy 2-point method will be used as the default jacobian if there is no implementation in the relevant solver.

`mantid` was fixed by not implementing `function1DDeriv` when wanting to use the default jacobian in mantid.

- [X] mantid
- [X] Scipy
- [x] GSL
- [x] RALFit
 
Fixes #980

#### Testing Instructions

1. Follow the testing instructions in the issue


tyronerees:
This still doesn't work for me locally -- I get output like: 
image

On the command line, I get an _Unknown method_ error:
image

This does work as expected for minimizers where `default` worked before.  Presumably this works on your end -- any ideas what's up for me?

robertapplin:
I am getting the same, the next commit should fix this

tyronerees (commented):
All works as expected now -- thanks!

These lines of `fitbenchmarking/results_processing/local_min_table.py` are no longer covered by the tests:
image

This is incorrectly implemented (not by this PR, I should say), and the try-except should be replaced by something like
```
if result.cost_func.jacobian.use_default_jac:
            # If using a solver dependent Jacobian, we want
            # to switch to a common jacobian here to generate
            # comparisons
            jacobian_cls = create_jacobian("scipy")
            jacobian_instance = jacobian_cls(result.cost_func.problem)
            jacobian_instance.method = "2-point"
            result.cost_func.jacobian = jacobian_instance
            jac = result.cost_func.jac_res(result.params,
                                           x=result.data_x,
                                           y=result.data_y,
                                           e=result.data_e)
else:
            jac = result.cost_func.jac_res(result.params,
                                           x=result.data_x,
                                           y=result.data_y,
                                           e=result.data_e)

```
Should we amend this here, or open another issue?

robertapplin:
> 

Glad its working now! I'll amend this here

fitbenchmarking/results_processing/local_min_table.py
```diff
@@ -102,23 +101,11 @@ def get_value(self, result):
...
-            result.cost_func.jacobian = jacobian_instance
-            jac = result.cost_func.jac_res(result.params,
-                                           x=result.data_x,
-                                           y=result.data_y,
-                                           e=result.data_e)
+
+        jac = result.cost_func.jac_res(result.params,
+                                       x=result.data_x,
+                                       y=result.data_y,
+                                       e=result.data_e)
```
tyronerees, line 108: Of course -- the default is now scipy-2 point, and so there's no need for an if statement....much neater!

tyronerees (approved):
All looks great -- thanks, Rob!

PR #991, opened 2021-12-01
Interactive toggle of table rows and columns

### LINKED ISSUE #872: Allow users to hide/show columns in the html tables
I found this http://jsfiddle.net/d6JWV/575/ which has an example of how we can toggle the visibility of a column. It would be nice if we use something like this to toggle columns in our results tables. E.g. all of the results for a certain software.

robertapplin:
#### Description of Work
This PR adds the ability to hide/show particular rows (problem sets) and columns (minimizers) in the fitbenchmarking comparison tables.

Fixes #872

#### Testing Instructions

1.
2.
3.


fitbenchmarking/results_processing/tests/expected_results/minimizer_dropdown.html
```diff
@@ -0,0 +1,7 @@
+<div id="minimizer_dropdown" class="dropdown-check-list" tabindex="100">
+    <span class="anchor" onclick="show_dropdown('minimizer_dropdown')">Select Minimizers</span>
+    <ul class="items">
+        <li><label class="noselect"><input type="checkbox" checked=true onclick="toggle_minimizer('scipy-ls', 'dogbox')"/> dogbox</label></li>
+        <li><label class="noselect"><input type="checkbox" checked=true onclick="toggle_minimizer('scipy-ls', 'lm-scipy')"/> lm-scipy</label></li>
+        <li><label class="noselect"><input type="checkbox" checked=true onclick="toggle_minimizer('scipy-ls', 'trf')"/> trf</label></li>    </ul>
```
AndrewLister-STFC, line 6: It looks like a new line gets missed before the `<\ul>` either that or we should remove the spaces.

fitbenchmarking/results_processing/tests/test_tables.py
```diff
@@ -177,36 +177,60 @@ def test_tables_correct(self):
...
+                                                   "minimizer_dropdown.html")
+
+        for expected_file, dropdown_name in zip(
+                [expected_problem_dropdown, expected_minimizer_dropdown],
+                ["problem_dropdown", "minim_dropdown"]):
+            self.compare_files(expected_file, html[dropdown_name])
+
+    def compare_files(self, expected, achieved):
         """
         Compares two tables line by line
```
AndrewLister-STFC, line 211:
```suggestion
        Compares two files line by line
```

AndrewLister-STFC (changes requested):

fitbenchmarking/results_processing/tests/expected_results/minimizer_dropdown.html
robertapplin: Nice spot 👍

AndrewLister-STFC:
This seems to be very browser specific... on firefox I get: 
image

robertapplin:
> This seems to be very browser specific... on firefox I get: 

Is the `dropdown_style.css` in the `fitbenchmarking/fitbenchmarking_results/css/` directory? I might have missed something - does this css sheet need to be copied over manually?

robertapplin:
I would have thought the `css` directory would be copied over here https://github.com/fitbenchmarking/fitbenchmarking/blob/8c92b4757960eb2057491d5cdbf90a760ebcf016/fitbenchmarking/cli/main.py#L141

Is there a reason why we don't do this, or is it done somewhere else @AndrewLister-STFC ?

AndrewLister-STFC:
> I would have thought the `css` directory would be copied over here
> 
> https://github.com/fitbenchmarking/fitbenchmarking/blob/8c92b4757960eb2057491d5cdbf90a760ebcf016/fitbenchmarking/cli/main.py#L141
> 
> Is there a reason why we don't do this, or is it done somewhere else @AndrewLister-STFC ?

Not sure why it's done how it is, but it's in https://github.com/fitbenchmarking/fitbenchmarking/blob/master/fitbenchmarking/core/results_output.py for the css.

AndrewLister-STFC:
There also seems to be a bug when ony 1 minimizer is used for a software as there's no colspan to decrement so we get NaN

robertapplin:
> There also seems to be a bug when ony 1 minimizer is used for a software as there's no colspan to decrement so we get NaN

Thanks for testing, both problems should be fixed now

fitbenchmarking/templates/js/table.js
```diff
@@ -0,0 +1,65 @@
...
+* @param   {jQuery Object} header   A collection of DOM elements representing column headers with a span.
+* @param   {bool} increment         True if the column span should be incremented. Otherwise it is decremented.
+*/
+function _adjust_colspan(header, increment) {
+    var colspan = parseInt(header.attr('colspan'));
+    var new_colspan = increment ? colspan + 1: colspan - 1;
+    header.attr('colspan', new_colspan);
+
+    // Hide the header if the sub-headers are all hidden
+    if (new_colspan == 0 || isNaN(new_colspan)) {
```
AndrewLister-STFC, line 22: This now fails on show as nan+1=nan so it stays hidden.
We may have to do something like this on before `var new_colspan` (syntax may be wrong):
```
if (isNaN(colspan)) {
    colspan = 1;
}
```

AndrewLister-STFC (changes requested):

fitbenchmarking/templates/js/table.js
robertapplin: Thanks again! Should be working this time round

AndrewLister-STFC (approved):
All looks good! I'm going to make a new issue for the bold line between cost functions as it disappears if the column is hidden (but that's very minor)

PR #996, opened 2021-12-06
JUnit Publish github action test results

### LINKED ISSUE #964: Add JUnit Publish action to create test results comments
Add something based on this to improve test feedback: https://github.com/marketplace/actions/publish-unit-test-results

robertapplin:
#### Description of Work
This PR attempts to publish the test results using the JUnit Publish github action. This aims to make the test results more readable and quicker to fix in the future.

Fixes #964

#### Testing Instructions

1.
2.
3.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -122,6 +122,7 @@ jobs:
     - name: 'Installing'
       run: |
         python -m pip install --upgrade pip==20.0.2
+        python -m pip install --upgrade wheel 
         pip install .[bumps,DFO,gradient_free,minuit,SAS,numdifftools]
         pip install -r requirements.txt
         sudo apt-get install libgsl-dev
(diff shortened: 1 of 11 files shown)


robertapplin:
This is what the output looks like when a test fails [this](https://github.com/fitbenchmarking/fitbenchmarking/runs/4432112325)

AndrewLister-STFC:
@robertapplin Did you say windows couldn't be added because of filename clashes?

robertapplin:
> @robertapplin Did you say windows couldn't be added because of filename clashes?

Yes, I think it wouldn't work because both the windows and linux workflow call the same `.sh` file, and this will create a junit xml file with the same name. Maybe it is possible to pass in a unique name to the `.sh` file?

AndrewLister-STFC:
> > @robertapplin Did you say windows couldn't be added because of filename clashes?
> 
> Yes, I think it wouldn't work because both the windows and linux workflow call the same `.sh` file, and this will create a junit xml file with the same name. Maybe it is possible to pass in a unique name to the `.sh` file?

That's fine, usually if windows fails so does linux, and if not it's probably an install issue

AndrewLister-STFC (approved):

PR #1000, opened 2021-12-07
Add Accuracy and Runtime to the Fitting Reports

### LINKED ISSUE #998: Add Chi Squared and Runtime to Fitting Report page
**Description**
The chi squared and runtime of a fit should be added to the Fitting report page. This might be loosely related to #914 .

robertapplin:
#### Description of Work
This PR adds the accuracy and runtime of a fit to its Fitting report:
image

Fixes #998

#### Testing Instructions

1. Run any fit and make sure the accuracy and runtime is displayed


fitbenchmarking/results_processing/fitting_report.py
```diff
@@ -84,6 +84,8 @@ def create_prob_group(result, support_pages_dir, options):
             equation=result.problem.equation,
             initial_guess=result.ini_function_params,
             minimizer=result.minimizer,
+            accuracy=f"{result.chi_sq:.7f}",
+            runtime=f"{result.runtime:.7f}",
```
AndrewLister-STFC, line 88: I think g is generally cleaner as it uses significat figures rather than decimal places.
```suggestion
            accuracy=f"{result.chi_sq:.7g}",
            runtime=f"{result.runtime:.7g}",
```
Was there a reason for 7? It seems quite long but I guess we want to be informative.

AndrewLister-STFC (changes requested):

fitbenchmarking/results_processing/fitting_report.py
robertapplin: I chose 7 because that seemed to be the largest precision in the tables themselves.

Do you remember off the top of your head the number of significant figures used for the tables?

robertapplin: Looks like 4 sig figs so I'll do that 👍

AndrewLister-STFC (approved):

Issue #952, opened 2021-10-21
Change build system to use newer pyproject.toml approach


AndrewLister-STFC:
The setup.py approach to building a package is being(has been?) superceeded by declarative project definitions. In particular, [PEP261](https://www.python.org/dev/peps/pep-0621/) introduces the `pyproject.toml` file and this is starting to become what the "build-backends" are aiming to support.

For our purpose I don't know that the backend matters much? But I found that flit and trampolim which seem to be 2 relatively popular(?) options support it, whereas setuptools and Poetry are still debating how best to alter their metadata definitions.

Some examples of what build backends suggest:
- https://setuptools.pypa.io/en/latest/userguide/quickstart.html
- https://github.com/FFY00/trampolim
- https://flit.readthedocs.io/en/latest/pyproject_toml.html
- https://python-poetry.org/docs/pyproject/

AndrewLister-STFC:
Versioneer - https://github.com/python-versioneer/python-versioneer
This was recommended in the RSE slack and might be a better option than setuptools_scm

PR #1001, opened 2021-12-08
Fix Pygsl installation error

robertapplin:
#### Description of Work
This PR fixes the error causing Pygsl not to install using pip:
image

#### Testing Instructions

1. Make sure the tests pass.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,6 @@ setup(name='FitBenchmarking',
           ]
       },
       packages=find_packages(exclude=('*mock*', '*test*')),
-      setup_requires=['wheel'],
       install_requires=['docutils',
                         'numpy',
                         'matplotlib>=2.0',
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,7 @@ setup(name='FitBenchmarking',
           ]
       },
       packages=find_packages(exclude=('*mock*', '*test*')),
+      setup_requires=['wheel'],
       install_requires=['docutils',
                         'numpy',
                         'matplotlib>=2.0',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -38,6 +38,5 @@
                       'gradient_free': ['gradient-free-optimizers']},
       zip_safe=False,
       use_scm_version={'fallback_version': '1.0.0'},
-      setup_requires=['setuptools_scm'],
```
AndrewLister-STFC, line 41: I think this is used to match the version to the latest tag, does it not work aymore?

robertapplin: I'm not convinced its being run. I tried adding the `wheel` package to this but with no result. The setuptools documentation say that `setup_requires` is now deprecated, so not sure if its still working

robertapplin: I also tried unpinning pip but this just gave a new error https://github.com/fitbenchmarking/fitbenchmarking/runs/4455891900?check_suite_focus=true

robertapplin: Maybe we should do something like this as seen in the release workflow? https://github.com/fitbenchmarking/fitbenchmarking/blob/bfad0eb754e16f32f1d82c24fd196285efb73726/.github/workflows/release.yml#L21

Or the other option might be to switch to using `pyproject.toml`, but this will take longer to implement

AndrewLister-STFC: Yeah in the long run a swithc to pyproject.toml would be best, but I think that can wait until the new year. Fo now lets just get this passing again! I'll merge this after making some notes on #952

AndrewLister-STFC (approved):

PR #1013, opened 2021-12-15
Normalise naming conventions for minimizers

### LINKED ISSUE #1009: Performance profiler fails if 2 minimizers share a name
**Description of the error**
When running scipy and mantid, the performance profiler fails due to both having a minimizer called BFGS.

AndrewLister-STFC:
#### Description of Work
Use fitting result to generate minimizer names
Only pass jac and hess as strings to fitting result (rather than the objects)
Ensure any strings which are required to be unique use both software and minimizer (and jac and hess)
Remove unused return values from fitting_benchmark.py
Indent hessians into loop so running 2 hessian methods will use both hessian methods
Move alogrithm type from options into result

Fixes #1009

#### Testing Instructions

1. Make some varied results - multiple jacs, multiple hessians, etc.
   - [x] Output stream during run looks correct
   - [x] Table headers look correct
   - [x] Fitting reports appear correct (url, text, plots)
   - [x] Summary pages appear correct
2. Try running with mantid - BFGS and scipy - BFGS
   - [ ] Output stream during run looks correct
   - [ ] Table headers look correct
   - [ ] Fitting reports appear correct (url, text, plots)
   - [ ] Summary pages appear correct


tyronerees (changes requested):
The code looks a lot cleaner, and looks like the 'right' way to do this.

Only a couple of minor suggestions for changes in the code.

In terms of features, I'm not sure the way jacobians and hessians are reported in the headings is the clearest.  Here's a screenshot:

image

It seems so be:
- default jac, default hess: no descriptions
- custom jac, default hess: jac description, nothing
- default jac, custom hess: 'default', hess desciption
- custom jac, custom hess: jac description, hess description

It took me a bit to understand what's going on.  I wonder if it would be better to add, e.g., "j:scipy 2-point", "h:analytic" to all the non-default ones, just to make it (a bit more) clear what's being described here without requiring pre-knowledge of the order of jacobians and hessians.  Of course, this wasn't introduced here, and it might be better to create a new issue for this.

tyronerees (approved):

tyronerees (changes requested):
I've noticed two chunks missing from the coverage tests:
from `results_processing/performance_profiler.py`:
image
 
and from `core/fitting_benchmarking.py`:
image

Is there are a reason why these aren't being hit?

tyronerees (approved):

Issue #1017, opened 2022-01-11
Add bumps dream


AndrewLister-STFC:
Add bumps dream minimizer (which is global)

PR #1023, opened 2022-01-26
tag gradient_free optimizers as global

### LINKED ISSUE #1018: List gradient free minimizers as global
Currently these minimizers are not listed as global but claim to be. We should add them to the tag

tyronerees:
#### Description of Work

Fixes #1018 

#### Testing Instructions

1. Check the gradient_free minimizers show up in the list of global optimizers
2.
3.


AndrewLister-STFC:
image

Can we fill in any more gaps here?
Bumps has DE and Dream which they say "takes longer but it does an excellent job of exploring different minima and getting close to the global optimum". so not sure if that means we should classify them here?

tyronerees:
Probably bumps `de` should be included -- I'll add that.  We don't yet support DREAM (#1017), but that would be global too.

The `dfo` ones are definitely local methods.

tyronerees:
Sorry, I just realized I misread your comment.  There seem to be lots of missing `general` tags (which probably means this isn't well defined by us in the docs).  I think `general`=`all` minus `ls` (which sort of begs the questions as to why we have both tags...).  I'll do a sweep through and fix these too.

AndrewLister-STFC (approved):

PR #1028, opened 2022-02-16
Ensure max runtime is checked in matlab controllers

AndrewLister-STFC:
#### Description of Work

Adds Max runtime support for matlab controllers

#### Testing Instructions

1. Try some runs with low max runtime on matlab controllers


tyronerees (approved):
This seems to work for me for small times.  I haven't tried it on the long running case that caused the issue -- will run that as soon as it's merged to master.

PR #1030, opened 2022-02-18
Add a minimum python version

### LINKED ISSUE #1029: Confusing error when installing fitbenchmarking when the python version is too low
If pip installing with too low a python version, you get a failure of the form:
image

The minimal python version we support should be a requirement of a package so that the user gets a helpful error message.

AndrewLister-STFC:
#### Description of Work

Fixes #1029

#### Testing Instructions

1. Does it still install?
2. Does it stop old python versions?


tyronerees (approved):
Tested on 3.6 and 3.7, and does what it says on the tin -- thanks!

PR #1034, opened 2022-03-01
Move start_matlab call so it is only called once

AndrewLister-STFC:
#### Description of Work

Fixes an issue where memory was being accumulated by matlab controllers

#### Testing Instructions

1. Run some Matlab tests on large datasets. Check that memory is not reserved by matlab.


AndrewLister-STFC:
@tyronerees I think this should be ready now. I ran the tests and there is 1 failure but that will be solved by the other PR which should probably be merged after this one.

AndrewLister-STFC:
@tyronerees All updated with horace and ready for review

tyronerees:
I'm not sure this fixes the memory issues.

Here is the status before running fitbenchmarking:
image
I'm testing with `SASVIEW_Complex_Shapes`.  With scipy the memory goes up to 519MB (and one processor is at 100%).

Running matlab_stats, the `htop` looks like this:
image
There are lots of processes with 8.5% or 2.5% of the memory, but only one is using any CPU. The memory has also increased to 925MB (although I don't know how much is the matlab overhead).  This continued to grow until the process crashed.

I'll do some more investigations....

tyronerees:
Running SASView_complex_shapes_1D on fitbenchmarking-medium, I get the following error:
image

This seems to be an issue with the timer not being able to handle the exception raised(?)

AndrewLister-STFC:
> I'm not sure this fixes the memory issues.
> 
> Here is the status before running fitbenchmarking: 
> I'm testing with `SASVIEW_Complex_Shapes`. With scipy the memory goes up to 519MB (and one processor is at 100%).
> 
> Running matlab_stats, the `htop` looks like this: image There are lots of processes with 8.5% or 2.5% of the memory, but only one is using any CPU. The memory has also increased to 925MB (although I don't know how much is the matlab overhead). This continued to grow until the process crashed.
> 
> I'll do some more investigations....

I think that's fine in terms of the htop, if you hit F5 in htop it will show whether these are all subprocesses of the same process or seperate by listing it as a tree - when I have 2 engines open I have around 70 entries per engine in this view....

In the screenshot, I'd expect the 100% cpu one to be the main one. The 2.5% mem processes are all python ones, and the fact that the matlab processes are all 8.5% mem makes me think it's the same process with subprocesses which I think is out of our control.

Skipping ahead 30 mins of inverstigating:
With regard to the timer issue from this morning, I can't recreate it in a run but I think I have a thread to pull:
 - The timer is created in matlab and saved as 'timer' with various pass throughs for the mathods
 - After runs we call clean_matlab() which clears the memory (including 'timer') and points calls to timer back to the python one
 - If we raise an exception, clean_matlab() gets skipped.
 - Calling 'timer.reset' in matlab will give the error message above (timer is a class already in matlab that I'm overriding)

So the third point raises one of why memory would be inflating... It's not clearing it -.-
But I don't know why the timer error is being raised as the timer from step 1 is only deleted if the clean_matlab() function is run, at which point it should be switching back to the python one so reset wouldn't be called on the matlab 'timer' class

tyronerees (approved):
This works for me, and all looks good

PR #1036, opened 2022-03-02
Enable max runtime in curve fitting

AndrewLister-STFC:
#### Description of Work

Fixes matlab curve fitting toolbox with max runtime

Replaces the wrapper with one written in matlab. This is written to a temp file and saved for matlab to open. The cost function and default data are passed via global variables... which might be unwise but I thought it was ok in this very bizare situation.

A side effect of this chang is that the serialisation of all matlab controllers will now be done for the cost function. With individual methods on the cost function being accessed from the same object. This should be an improvement as we currently have multiple instances of the exact (hopefully) same object when both the residual and jacobian are passed into matlab.

TODO:
 - Incorporate changes from #1034 as this will be a clash

#### Testing Instructions

1. Try running with matlab controller and max runtime.
2. Run the matlab tests

### DIFF
--- a/fitbenchmarking/controllers/matlab_mixin.py
+++ b/fitbenchmarking/controllers/matlab_mixin.py
@@ -19,7 +19,7 @@ try:
     eng = matlab.engine.connect_matlab(name='FITBENCHMARKING_MATLAB')
 except matlab.engine.EngineError:
     eng = matlab.engine.start_matlab()
-    eng.shareEngine('FITBENCHMARKING_MATLAB')
+    eng.matlab.engine.shareEngine('FITBENCHMARKING_MATLAB', nargout=0)
 
 
 # If we re-implement caching, make sure the cache is cleared by the
--- a/fitbenchmarking/controllers/matlab_mixin.py
+++ b/fitbenchmarking/controllers/matlab_mixin.py
@@ -15,7 +15,11 @@ try:
 except ImportError:
     import_success = False
 
-eng = matlab.engine.start_matlab()
+try:
+    eng = matlab.engine.connect_matlab(name='FITBENCHMARKING_MATLAB')
+except matlab.engine.EngineError:
+    eng = matlab.engine.start_matlab()
+    eng.shareEngine('FITBENCHMARKING_MATLAB')
 
 
 # If we re-implement caching, make sure the cache is cleared by the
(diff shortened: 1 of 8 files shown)


tyronerees:
As discussed, it might be cleaner to write a static ".m"  file, rather than generate a temporary one each time.

AndrewLister-STFC:
The matlab tests still fail on this, although running the same thing manually works.

AndrewLister-STFC:
> The matlab tests still fail on this, although running the same thing manually works.

Update: The test also works on when run in isolation

fitbenchmarking/controllers/matlab_mixin.py
```diff
@@ -26,63 +30,78 @@ class MatlabMixin:
     Mixin class for matlab fitting software controllers
     """
 
+    incompatible_problems = ['mantid']
```
tyronerees, line 33: since this is now set in `matlab_mixin`, do we still need to set this in the individual controller files?

tyronerees (commented):
Looks good to me -- just one comment (which is probably just my not understanding how this all hangs together...)

fitbenchmarking/controllers/matlab_mixin.py
AndrewLister-STFC: I had to do some git hunting to see why it's in individuals instead of the mixin, but it came up in a previous review and I think Rob thought it was clearer in individual controllers (although neither of us had particularly strong opinions).

You are right though that either place works so happy to remove this or clean out the other controllers.

AndrewLister-STFC: Discussion here: https://github.com/fitbenchmarking/fitbenchmarking/pull/979#discussion_r752378067

fitbenchmarking/controllers/matlab_mixin.py
```diff
@@ -26,63 +30,78 @@ class MatlabMixin:
...
+
     def __init__(self, cost_func):
```
AndrewLister-STFC, line 35:
```suggestion
    Mixin class for matlab fitting software controllers
    """

    def __init__(self, cost_func):
```

fitbenchmarking/controllers/matlab_mixin.py
AndrewLister-STFC: I've removed the line now

tyronerees (approved):
Looks good!

PR #1050, opened 2022-07-12
Fix the poisson cost func test

AndrewLister-STFC:
#### Description of Work
It looks like some dependency has been updated and now one of our tests is failing as the test is too precice?

Fixes the tests on master

#### Testing Instructions

1. Do the tests pass?
2.
3.


tyronerees (approved):
👍

PR #1058, opened 2022-09-12
Add checkpointing

### LINKED ISSUE #1044: Split data generation and data analysis in the backend
When running fitbenchmarking, we should save the data before we generate the tables and plots so that the two phases are decoupled and can be run independently.

AndrewLister-STFC:
#### Description of Work

Adds checkpointing.
In the first draft, the runs are checkpointed by a singleton class (I'm a bit hesitant with this as it feels like a global...).

The checkpointer writes the important variables to temporay files split into the problem and result aspects -- see code for the split. When all the runs have been completed, the checkpointer is finalised which combines these into a single file.

To load a checkpoint, this json file is loaded and the values extracted.

Still to do:

- [x] Populate the duplicated data in the fitting result (the ``_tag fields``). Raises the questions -- do we need them? should they be explicitly copied into the file to prevent future complications?
- [x] Store the additional data for J(p), r(p), some plotting vals which are currently calculated when required in the reporting.
- [x] Add tests for the checkpointer
- [x] Document the checkpointer
- [x] Add results to the checkpointer
- [x] Load the results from the checkpointer
- [x] ~Rewrite results tests to start from a checkpoint file~ This can/should be a separate issue

Fixes #1044 

#### Testing Instructions

1.
2.
3.


AndrewLister-STFC:
In the current state on this branch the tests include a jacobian which is set to 0. This should probably be updated for the sake of the local_min tests before merging

AndrewLister-STFC:
Having written tests that run in isolation but not in the suite I've come to accept that the Singleton design is a bad idea. For this I'll pass it through to all the required places via function args (and create a new issue with ideas for a cleaner but more dramatic solution)

fitbenchmarking/controllers/bumps_controller.py
```diff
@@ -121,9 +122,11 @@ def setup(self):
         self.fit_order = param_order
 
         if self.minimizer == "lm-bumps":
-            self.minimizer = "lm"
+            self._minimizer = "lm"
```
AndrewLister-STFC, line 125: Add a comment for why here

fitbenchmarking/results_processing/base_table.py
```diff
@@ -288,9 +288,17 @@ def create_pandas_data_frame(self, html=False):
...
         str_results = self.get_str_dict(html)
-        row = next(iter(self.sorted_results.values()))
-        minimizers_list = [(r.software, r.modified_minimizer_name(False))
-                           for r in row]
+        # Check all rows incase first row has missing information
+        minimizers_list = []
+        for row in self.sorted_results.values():
+            for i, r in enumerate(row):
+                formatted = (r.software, r.modified_minimizer_name(False))
+                print(formatted)
```
AndrewLister-STFC, line 296: Remove print

tyronerees:
From our review meeting, we should add a link to `checkpointing.py` somewhere on the developers page, just so it's easy to follow the breadcrumbs to find where to add new options if and when necessary.

tyronerees:
The command `fitbenchmarking-cp` only seems to work from the base folder of fitbenchmarking -- running it from elsewhere I get:

image

Presumably this will break things if installed from pip?

tyronerees:
When we run `fitbenchmarking-cp -h`, the details on what report does are scant: 

image

Is it possible to point people to `fitbenchmarking-cp report -h` here for details?  I know it's in the sphinx docs, but it'd be good to be here too.

tyronerees:
Load checkpoint doesn't work with multiple datasets.  If you run:

image

then you get the error:

image

AndrewLister-STFC:
@tyronerees Running say:
```
> cd /var
> fitbenchmarking-cp -h
```
works for me

AndrewLister-STFC:
@tyronerees I think this should be ready... finally

tyronerees (approved):
Happy to approve!

PR #1061, opened 2022-09-26
Set certain options input via argparse

### LINKED ISSUE #672: Allow certain options input via `argparse`
It would be useful to be able to set some options via `argparse` rather than from an `ini` file or from the default options. These options include:

- `fitting.num_runs`
- `fitting.algorithm_type`
- `fitting.software`
- `fitting.jac_method`
- `fitting.cost_func_type`
- `jacobian.scipy`
- `jacobian.analytic`
- `plotting.make_plots`
- `plotting.comparison_mode`
- `plotting.table_type`
- `plotting.results_dir`
- `logging.file_name`
- `logging.append`
- `logging.level`
- `logging.external_output`

I don't suggest that all these options to be set via the argparse at once but maybe a subset. Some of the options about may not be suitable to set via `argparse`, for example, the Jacobian options

Create a function within the options class called `get_argparse_options` (or similar) which overrides options. Thus, the options will be set in the following order:

1. default
2. `ini` file
3. `argparse`

where `argparse` is the priority.

Shanrahan16:
#### Description of Work

Fixes #672 

#### Testing Instructions

1.
2.
3.

### DIFF
--- a/fitbenchmarking/cli/main.py
+++ b/fitbenchmarking/cli/main.py
@@ -66,7 +66,7 @@ of the Fitbenchmarking docs. '''
     parser.add_argument('-o', '--options-file',
                         metavar='OPTIONS_FILE',
                         default='',
-                        help='The path to a %(prog)s options file')
+                        help='The path to a %(prog)s options file.')
     parser.add_argument('-p', '--problem_sets',
                         nargs='+',
                         default=glob.glob(os.path.join(root,
@@ -81,12 +81,72 @@ of the Fitbenchmarking docs. '''
     parser.add_argument('-d', '--debug-mode',
                         default=False,
                         action='store_true',
-                        help='Enable debug mode (prints traceback)',)
+                        help='Enable debug mode (prints traceback).',)
+    parser.add_argument('-n','--num_runs',
+                        metavar='NUM_RUNS',
+                        type = int,
+                        default = 0,
+                        help = "Set the number of runs to average each fit over.")
+    parser.add_argument('-a', '--algorithm_type',
+                        metavar = 'ALGORITHM_TYPE',
+                        nargs='+',
+                        type = str,
+                        default = [],
+                        help = "Select what type of algorithm is used within a specific software.")
+    parser.add_argument('-s', '--software',
+                        metavar = 'SOFTWARE',
+                        nargs='+',
+                        type = str,
+                        default = [],
+                        help = "Select the fitting software to benchmark.")    
+    parser.add_argument('-j', '--jac_method',
+                        metavar = 'JAC_METHOD',
+                        nargs='+',
+                        type = str,
+                        default = [],
+                        help = "Set the Jacobian to be used.")
+    parser.add_argument('-c', '--cost_func_type',
+                        metavar = 'COST_FUNC_TYPE',
+                        nargs='+',
+                        type = str,
+                        default = [],
+                        help = "Set the cost functions to be used for the given data.")
 
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--do_make_plots', action = 'store_true')
+    group.add_argument('--dont_make_plots', action = 'store_true')
+
+    parser.add_argument('-m', '--comparison_mode',
+                        metavar = 'COMPARISON_MODE',
+                        default = '',
+                        help = "Select the mode for displaying values in the resulting table.")               
+    parser.add_argument('-t', '--table_type',
+                        metavar = 'TABLE_TYPE',
+                        nargs='+',
+                        type = str,
+                        default = [],
+                        help = "Select the type of table to be produced in FitBenchmarking.")                
+    parser.add_argument('-f', '--logging_file_name',
+                        metavar = 'LOGGING_FILE_NAME',
+                        default = '',
+                        help = "Specify the file path to write the logs to.") 
+
+    group2 = parser.add_mutually_exclusive_group()
+    group2.add_argument('--do_append', action = 'store_true')
+    group2.add_argument('--dont_append', action = 'store_true')
+
+    parser.add_argument('-l', '--level',
+                        metavar = 'LEVEL',
+                        default = '',
+                        help = "Specify the minimum level of logging to display on console during runtime.")  
+    parser.add_argument('-e', '--external_output',
+                        metavar = 'EXTERNAL_OUTPUT',
+                        default = '',
+                        help = "Select the amount of information displayed from third-parties.")                       
     return parser
 
 
-def _find_options_file(options_file: str, results_directory: str) -> Options:
+def _find_options_file(options_file: str, results_directory: str, additional_options : dict) -> Options:
     """
     Attempts to find the options file and creates an Options object for it.
     Wildcards are accepted in the parameters of this function.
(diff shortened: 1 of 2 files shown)


fitbenchmarking/utils/options.py
```diff
@@ -176,7 +176,7 @@ class Options:
                 'OUTPUT': DEFAULT_OUTPUT,
                 'LOGGING': DEFAULT_LOGGING}
 
-    def __init__(self, file_name=None, results_directory: str = ""):
+    def __init__(self, additional_options = {}, file_name=None, results_directory: str = ""):
```
AndrewLister-STFC, line 179: The style checker is upset with a default argument of `{}`. The details of this are a bit murky, but it doesn't like any object as a default argument, it's much happier with a literal.

The standard way to get around this is have a default argument of `None`, and then overwrite this if required:
```
def some_function(arg1, arg2={}, arg3=None):
    """
    arg2 and arg3 are both initialied to an empty dict if no value is given.
    The style checker will complain about arg2. arg3 is safer.
    """
    if arg3 is None:
        arg3 = {}
    ...
```

I also think additional_options should be after file_name

fitbenchmarking/cli/main.py
```diff
@@ -81,12 +81,72 @@ def get_parser():
...
+                        help = "Specify the minimum level of logging to display on console during runtime.")  
+    parser.add_argument('-e', '--external_output',
+                        metavar = 'EXTERNAL_OUTPUT',
+                        default = '',
+                        help = "Select the amount of information displayed from third-parties.")                       
     return parser
 
 
-def _find_options_file(options_file: str, results_directory: str) -> Options:
+def _find_options_file(options_file: str, results_directory: str, additional_options : dict) -> Options:
```
AndrewLister-STFC, line 149: The in-code documentation (text in the triple quotes) for this will need updating to explain what additional_options is.

fitbenchmarking/utils/options.py
```diff
@@ -230,14 +230,34 @@ def __init__(self, file_name=None, results_directory: str = ""):
...
-            fitting.getlist, 'algorithm_type')
-        self.software = self.read_value(fitting.getlist, 'software')
-        self.jac_method = self.read_value(fitting.getlist, 'jac_method')
+
+        if 'num_runs' in additional_options and additional_options['num_runs']:
+            self.num_runs = additional_options['num_runs']
+        else:
+            self.num_runs = self.read_value(fitting.getint, 'num_runs')
+
+        if 'algorithm_type' in additional_options and additional_options['algorithm_type']:
```
AndrewLister-STFC, line 239: This is good but I'm a bit concerned that typos could cause hard to detect errors. For example running the following produces no output:
```
fitbenchmarking -a levenburg-marqadt
```
whereas
```
fitbenchmarking -a levenberg-marquardt
```
works fine.

Lets have a think of how best to do this and discus it later this week, if we can use the check that's already in `Options.read_value()` it may be neat.

fitbenchmarking/cli/main.py
```diff
@@ -81,12 +81,72 @@ def get_parser():
...
+                        metavar='NUM_RUNS',
+                        type = int,
+                        default = 0,
+                        help = "Set the number of runs to average each fit over.")
+    parser.add_argument('-a', '--algorithm_type',
+                        metavar = 'ALGORITHM_TYPE',
+                        nargs='+',
+                        type = str,
+                        default = [],
+                        help = "Select what type of algorithm is used within a specific software.")
```
AndrewLister-STFC, line 95: The style checker is also upset about line length here, we can just break it up and it should be happy.
```suggestion
                        help="Select what type of algorithm is used within a "
                             "specific software.")
```

AndrewLister-STFC (changes requested):
Great work, it's very nice to be able to try something with just the new options instead of having to edit the file!

I've left a few comments below, please let me know if you have questions, comments, or disagree with any of them.

fitbenchmarking/cli/main.py
```diff
@@ -316,10 +376,39 @@ def main():
...
+        
+    #Check if log_append in options.py should be overridden, and if so, add to options_dictionary                    
+    if args.do_append:
+        options_dictionary['append'] = True
+    elif args.dont_append:
+        options_dictionary['append'] = False
+
+
     run(problem_sets=args.problem_sets,
         results_directory=args.results_dir,
```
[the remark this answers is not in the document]
Shanrahan16: Okay, I didn't edit that one at all as it already existed, so I'll take a look.

fitbenchmarking/cli/main.py
```diff
@@ -184,7 +244,7 @@ def _open_browser(output_file: str) -> None:
 
 
 @exception_handler
-def run(problem_sets, results_directory, options_file='', debug=False):
+def run(additional_options, problem_sets, results_directory, options_file='', debug=False):
```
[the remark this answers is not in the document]
Shanrahan16: I've changed it now, but previously it didn't seem to have a default, so what would it have done?

fitbenchmarking/utils/options.py
Shanrahan16: Done for all except `make_plots`

fitbenchmarking/utils/options.py
```diff
@@ -227,79 +234,115 @@ def __init__(self, file_name=None, results_directory: str = ""):
...
             raise OptionsError('\n'.join(self.error_message))
 
-    def reset(self):
-        """
-        Resets options object when running multiple problem groups.
-        """
-        self.__init__(self.stored_file_name, self.results_dir)
-
-    def read_value(self, func, option):
+    def read_value(self, func, option, additional_options_func):
```
AndrewLister-STFC, line 323: Did you intend to leave this as "additional_options_func"? I'm not sure the "_func" adds anything.

AndrewLister-STFC (changes requested):
I think this is good for the code - just one very minor point in the options.py file.

There's a red cross on coverage (which indicates how much of our code is tested). I'll double check if it's an issue but I don't think we need to worry about it as it's not a big drop in coverage.

The final thing to do is update the docs. [this page](https://fitbenchmarking--1061.org.readthedocs.build/en/1061/users/running.html) describes the `-r` (results_dir) option but with all of the new options I think we should just tell users how to use `-h` (help) instead.
The file to edit is `docs/src/users/running.rst`.

fitbenchmarking/utils/options.py
Shanrahan16: I got rid of the `func` and then added it back when it wasn't working, so I'll remove it again now

Shanrahan16:
> I think this is good for the code - just one very minor point in the options.py file.
> 
> There's a red cross on coverage (which indicates how much of our code is tested). I'll double check if it's an issue but I don't think we need to worry about it as it's not a big drop in coverage.
> 
> The final thing to do is update the docs. [this page](https://fitbenchmarking--1061.org.readthedocs.build/en/1061/users/running.html) describes the `-r` (results_dir) option but with all of the new options I think we should just tell users how to use `-h` (help) instead. The file to edit is `docs/src/users/running.rst`.

I'm thinking of putting the table of arguments for the options as an image in the docs?

AndrewLister-STFC (approved):
I think this is good to go! We just need to update the branch and wait for the tests to run again then we can merge 🎉

Issue #1055, opened 2022-07-26
Add progress bars


AndrewLister-STFC:
**Is your feature request related to a problem? Please describe.**
It can be hard to see how much computation remains.

**Describe the solution you'd like**
We should implement progress bars, either:
- tqdm: This is best if our steps are short.
- https://github.com/alexhroom/verbosemanager: This allows for much finer grained progress messages during execution.

PR #1063, opened 2022-09-30
Add a progress bar for the benchmark problems (#1055)

joshuadkitenge:
#### Description of Work
Add a progress bar to indicate how far the fitbenchmarking code is through the benchmark problems, starting values and software.

- The progress bar always exist for the any number of benchmark problems and still exist after completion i.e at 100%.  
- The progress bar only exists for the starting values if there are more than or equal to 2 starting values and does not exist after completion.  
- The progress bar only exists for the software if there are more than or equal to 3 software and does not exist after completion.   

The progress bars for the starting values and the software don't exist after completion as it is a sub process to the benchmark problems. Therefore it's tracking the progress within the benchmark problems, so when it has finished we would want to clear the progress bars so the progress bars don't persist on the terminal.    
 
Fixes

#### Testing Instructions

1. Test using more than or equal to 2 starting starting values and more than or equal to 3 software


fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -111,27 +113,29 @@ def loop_over_benchmark_problems(problem_group, options):
...
-        # Loops over starting values #
-        ##############################
-        problem_results, problem_fails, \
-            unselected_minimizers = \
-            loop_over_starting_values(
-                problem, options, grabbed_output)
-        results.extend(problem_results)
-        failed_problems.extend(problem_fails)
+    pb = tqdm(problems,colour='green',leave=True)
+    with logging_redirect_tqdm():
```
AndrewLister-STFC, line 117: Specifying the logger seems to get it working!
```suggestion
    with logging_redirect_tqdm(loggers=[LOGGER]):
```

joshuadkitenge: It does, cheers however it is printing the progress bar on every line now. 

image.

I edited the "tqdm()" so "leave" is = "false" and it hasn't changed anything.  

image

I'll investigate how to prevent this from happen and keep the progress bar to the bottom of command line output.

joshuadkitenge:
The progress bar work fine on anaconda prompt, command prompt and powershell

Powershell 
image

Command prompt 
image

Anaconda prompt
image

joshuadkitenge:
pbar_bp_sv

tyronerees:
Not sure why, but this fails for me: 

image

I have `tqdm` (v 4.50.0) installed:
image
which is a little old (but the same major version) -- do you think this might be the issue, or is there something else wrong with my setup?

tyronerees:
Upgrading `tqdm` got it to work 😃

fitbenchmarking/core/tests/test_fitting_benchmarking_software.py
```diff
@@ -108,20 +110,41 @@ def shared_test(self, expected_list_len, expected_minimizer_failed):
         dict_test(unselected_minimzers, expected_minimizer_failed)
 
     @unittest.mock.patch('{}.loop_over_minimizers'.format(FITTING_DIR))
-    def test_run_software(self, loop_over_minimizers):
+    def test_run_one_software(self, loop_over_minimizers):
         """
-        Checks that results are produced for all minimizers within the
+        Checks that results are produced for one minimizers within the
```
tyronerees, line 115:
```suggestion
        Checks that results are produced for one minimizer within the
```

fitbenchmarking/core/tests/test_fitting_benchmarking_software.py
```diff
@@ -108,20 +110,41 @@ def shared_test(self, expected_list_len, expected_minimizer_failed):
...
+        self.results_problem = \
+            [[fitbm_result.FittingResult(**self.result_args)
+              for i in range(self.scipy_len)]]
+        loop_over_minimizers.side_effect = self.mock_func_call
+        expected_list_len = self.scipy_len
+        expected_minimizer_failed = self.minimizer_failed
+        self.shared_test(expected_list_len, expected_minimizer_failed)
+
+    @unittest.mock.patch('{}.loop_over_minimizers'.format(FITTING_DIR))
+    def test_run_multitple_softwares(self, loop_over_minimizers):
```
tyronerees, line 130:
```suggestion
    def test_run_multiple_softwares(self, loop_over_minimizers):
```

tyronerees (changes requested):
This is such a quality of life improvement, looks really good -- thanks, Joshua!

I'm not entirely sure what the stats being printed out mean.  I think these should either be documented somewhere or, ideally, displayed in such a way that it's obvious what they mean.

For example, at the start: 
image
What does `00:02<?` mean?  What does `?Benchmark problems/s` mean?

In the middle:
image
what does `00:18<02:25` mean?

At the end:
image
what does `01:35/00:00` mean?  What does `10.56s/Benchmark problems` mean?

joshuadkitenge:
@tyronerees 
do you think the best place to put the additional documentation that describes what the tqdm statistics are is in  docs/source/users/running.rst  https://fitbenchmarking.readthedocs.io/en/stable/users/running.html

tyronerees:
@joshuadkitenge -- do you have a few minutes to chat about this?  In terms of the docs, that's probably the best place, but I think it would be better if it was self-documenting (as no-one's actually going to read all the docs before running it).  Maybe we could have a quick brainstorm and come up with a way to format it that wouldn't need formal documentation

joshuadkitenge:
image


I have added the additional text in "Fitbenchmarking --help" to describe the tqdm statistics.  Do you have any comments before I commit the changes. @tyronerees

PR #1076, opened 2022-10-31
Headings in options #678

### LINKED ISSUE #678: headings in options
In the options, the `[PLOTTING]` options all have `[FITTING]` as a header.

image

Also, it's not obvious which section each option is in -- having these laid out on one page may be more helpful to a user.

[PLOTTING] should be changed to [OUTPUT], as most of the options don't relate to plotting.

Shanrahan16:
#### Description of Work

Fixes #678


docs/source/users/options/output_option.rst
```diff
@@ -1,10 +1,10 @@
...
 
-##############
-Output Options
-##############
+################
+Plotting Options
+################
 
-The output section contains options to control how results are outputted.
+The plotting section contains options to control how results are outputted and presented.
```
tyronerees, line 7: I think we should merge `Plotting` into `Output`, not `Output` into `Plotting`; plotting is a form of output, but I'd never to think to look for where to set the output directory under a section called `Plotting`.

docs/source/users/options/output_option.rst
```diff
@@ -20,4 +20,123 @@ Default is ``fitbenchmarking_results``
 .. code-block:: rst
 
     [OUTPUT]
```
tyronerees, line 22: All the headers should be the same now -- either all `[OUTPUT]` or all `[PLOTTING]`

tyronerees (changes requested):
I think we should also merge the `SECTIONS` in `fitbenchmarking/utils/options.py` too -- breaking the convention of one doc page per section for these two sections is likely to confuse users.

Shanrahan16:
I think I also need to update the ini file templates

docs/source/extending/options_extend.rst
```diff
@@ -13,7 +13,7 @@ The options used can be changed using an `.ini` formatted file
 FitBenchmarking.
 
 To add a new option to one of the five sections ``FITTING``, ``MINIMIZERS``,
```
tyronerees, line 15: I love how the number of sections is now correct, since you've taken one out 😆

tyronerees (approved):
Works for me, and looks much cleaner -- thanks!

PR #1077, opened 2022-11-04
Avoid parsing Mantid function strings

jfowkes:
#### Description of Work

At present we attempt to parse Mantid function strings ourselves, this is a fool's errand as the general Mantid function string syntax is far too complicated for us to handle and should be left to Mantid. 

#### Testing Instructions

1. Try parsing a really complicated Mantid function string


AndrewLister-STFC:
As a comment, we did change to leaving the parsing to Mantid, but it looks like the entry in the input is still being being read by our parser (the result of which goes unused).

AndrewLister-STFC (commented):
Fix get_equation signature

Edit: This comment was an accident while editing on the online vscode

joshuadkitenge (approved):

PR #1085, opened 2022-11-15
update ralfit options

### LINKED ISSUE #1078: Expose more features in RALFit controller
RALFit can exploit higher order derivatives, and we don't take advantage of that in FitBenchmarking.  We should add extra options to the RALFit controller.

tyronerees:
#### Description of Work

Fixes #1078 

#### Testing Instructions

1. check coverage hasn't decreased
2. check new options are documented and work
3.


fitbenchmarking/utils/tests/test_options_minimizers.py
```diff
@@ -75,7 +75,8 @@ def test_minimizer_ralfit(self):
         """
         Checks valid ralfit minimizers are set correctly
         """
-        expected = ['gn', 'gn_reg', 'hybrid', 'hybrid_reg']
+        expected = ['gn', 'gn_reg', 'hybrid', 'hybrid_reg',
+                    'newton', 'newton_reg']
```
Shanrahan16, line 79: Does this only look at the default minimizers? I can't tell?

fitbenchmarking/utils/options.py
```diff
@@ -139,7 +141,8 @@ class Options:
          'matlab_opt': ['levenberg-marquardt', 'trust-region-reflective'],
          'matlab_stats': ['Levenberg-Marquardt'],
          'minuit': ['minuit'],
-         'ralfit': ['gn', 'gn_reg', 'hybrid', 'hybrid_reg'],
+         'ralfit': ['gn', 'gn_reg', 'hybrid', 'hybrid_reg',
+                    'newton', 'newton_reg'],
```
Shanrahan16, line 145: How do you decide which minimizers should be the default?

fitbenchmarking/controllers/ralfit_controller.py
```diff
@@ -69,6 +75,18 @@ def setup(self):
...
+            self._options[b"type_of_method"] = 1
+        elif self.minimizer == "newton_reg":
+            self._options[b"model"] = 2
+            self._options[b"type_of_method"] = 2
+        elif self.minimizer == "newton-tensor":
+            self._options[b"model"] = 4
+            self._options[b"type_of_method"] = 1
+        elif self.minimizer == "newton-tensor_reg":
+            self._options[b"model"] = 4
+            self._options[b"type_of_method"] = 2
```
Shanrahan16, line 89: I've found what the numbers mean on the RALFit documentation, these all look fine

fitbenchmarking/utils/tests/test_options_minimizers.py
tyronerees: It does (I only found this out as the tests failed when I put all valid minimizers in).  I think that means the doc strings are all wrong, but that seems to be the case throughout the file -- I'll open an issue to fix that.

fitbenchmarking/utils/options.py
tyronerees: `newton` and `newton_reg` take an age to run (by design).  I think this means you should actively think you want to test with them, rather than think "I'll run RALFit" and then grumble that it takes forever.  I'll update the docs for RALFit to explain this.

PR #1089, opened 2022-11-16
theseus controller #1052

### LINKED ISSUE #1052: Add support for Theseus NLS solver
**Add the Theseus NLS solver to the benchmarks**

Theseus is an efficient application-agnostic library for building custom nonlinear optimization layers in PyTorch to support constructing various problems in robotics and vision as end-to-end differentiable architectures but it is also a stand-alone NLS.

https://sites.google.com/view/theseus-ai/

Reference: https://arxiv.org/pdf/2207.09442.pdf

joshuadkitenge:
#### Description of Work

Fixes #1052

#### Testing Instructions

1.
2.
3.


tyronerees:
@tyronerees to run tests on a local docker container to see why the tests fail

tyronerees (changes requested):
Only one minor change needed to the docs that I can see.  

Theseus doesn't seem to be the most robust (it fails on quite a few of the NIST problems), but when it works it gets the right answer.  I can't see anything in this implementation that would make it more likely to fall over, so suspect this is just a fundamental issue with Theseus.

tyronerees (approved):

PR #1098, opened 2022-12-07
Add LMFIT software #1096

### LINKED ISSUE #1096: Add LMFIT
https://lmfit.github.io/lmfit-py/

joshuadkitenge:
#### Description of Work

Fixes #1096

#### Testing Instructions

1. Check the the packages can be installed correctly using setup.py
2. Check that the minimizer works as expected
3. Check the documentation for any error e.g. licensing etc


fitbenchmarking/controllers/lmfit_controller.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+
+        if self._status == 0:
+            self.flag = 0
+        else:
+            self.flag = 2
+        
+        for _, param in self.lmfit_out.params.items():
+              self.result.append(param.value)
+
```
AndrewLister-STFC, line 168: Similar to what we were looking for earlier. The result object persists so using append gives problems.
You could either clear it (it's not clear to me why it needs to be an attribute) or use something like the following which replaces the bound object with a new list.
```suggestion
        self.result = list(self.lmfit_out.params.values())
```

joshuadkitenge:
Broke this branch and opened a new branch https://github.com/fitbenchmarking/fitbenchmarking/pull/1109

PR #1108, opened 2022-12-22
Added paramonte software and Bayesian comparison framework

joshuadkitenge:
#### Description of Work

Fixes

Still to do:

- [x] Add tests
- [x] Add documentation

#### Testing Instructions

1.
2.
3.


AnthonyLim23:
could fit a gaussian or skewed gaussian and report the fit parameters. This might be misleading and knowing if its gaussian or skewed could be difficult

AnthonyLim23:
we could get the peak value, y, and calculate (I have randomly chosen 10%)

int_0.9*y ^1.1*y P(x) dx

to get the confidence interval that the parameter is within the range. i.e. 40% chance that the value is between 0.9*y and 1.1*y

AnthonyLim23:
After talking to @jess-farmer the following could work.

1. Get the "true results" from a traditional fit (e.g. curve fit)
2. Calculate the probability distributions
3. integrate the area under the distribution that is between (1-s)*y and (1+s)*y, where y is the parameter value from the fit and s is some acceptable deviation from the value (e.g. 10% or 1 sigma)
4. The integral will then give the chance of the parameter being within s% of the expected value. 

The (badly) drawn diagram below shows how it could work.
image

AnthonyLim23:
Been thinking about this some more. In my previous comment I said that the area would be the percentage chance, it is actually the confidence of a point being within the bounds (since the y axis is probability). 

The second thing is that this approach is not very helpful for comparing different fitting problems. Since it would produce N numbers for N free parameters. However, we could combine them by taking the product (probability of being between a and b for x and being between c and d for y):

int dx P(x) int dy P(y) = int dx int dy P(x)P(y) = int dx int dy P(x and y)

jess-farmer:
> Been thinking about this some more. In my previous comment I said that the area would be the percentage chance, it is actually the confidence of a point being within the bounds (since the y axis is probability).
> 
> The second thing is that this approach is not very helpful for comparing different fitting problems. Since it would produce N numbers for N free parameters. However, we could combine them by taking the product (probability of being between a and b for x and being between c and d for y):
> 
> int dx P(x) int dy P(y) = int dx int dy P(x)P(y) = int dx int dy P(x and y)

@AnthonyLim23 I had also been thinking that the suggested approach doesn't provide an easy metric for comparing different fitting problems, and I think we want to be trying to convey information about the fit as concisely as possible (I do think that it would be good to have a more detailed results page for each problem with pdfs for each parameter, and confidence that that parameter is within the bounds). So, the idea of taking the product is to give an overall confidence level for all parameters being within the specified bounds?

AnthonyLim23:
yes exactly. It gives us a single number

docs/source/users/algorithms/bayesian.rst
```diff
@@ -0,0 +1,104 @@
...
+            r=\frac{p(\theta^*|y)/J_t(\theta^*|\theta^{t-1})}{p(\theta^{t-1}|y)/J_t(\theta^{t-1}|\theta^*)}
+    c. Set
+
+        .. math::
+            \theta^t = \begin{cases}\theta^*, & \text{with probability}\ min(r,1) \\ \theta^{t-1}, & \text{otherwise} \end{cases}
+
+Points to consider
+******************
+- MCMC methods do not have an explicit goal of refining the solution, so often a large chain size is required for convergence.
+- It can take some time for the chain to converge to the target distribution and so it's common to discard a number of initial samples
```
AnthonyLim23, line 87: we dont discard the samples, we keep the same number for the whole simulation. We just dont measure the distribution during the burn in period.

fitbenchmarking/controllers/lmfit_controller.py
```diff
@@ -71,7 +71,8 @@ class LmfitController(Controller):
         'global_optimization': ['differential_evolution',
                                 'ampgo',
                                 'shgo',
-                                'dual_annealing']
+                                'dual_annealing'],
+        'MCMC': []
```
jfowkes, line 75: I think this needs some indentation?

fitbenchmarking/controllers/nlopt_controller.py
```diff
@@ -85,7 +85,8 @@ class NLoptController(Controller):
                                 'GD_STOGO',
                                 'GD_STOGO_RAND',
                                 'GN_AGS',
-                                'GN_ISRES']}
+                                'GN_ISRES'],
+        'MCMC': []}
```
jfowkes, line 89: Also needs some indentation?

jfowkes (approved):
Looks great, see minor suggestions below.

AnthonyLim23:
I think the run command needs updating to `mpiexec -n 4 python ...` https://mpi4py.readthedocs.io/en/stable/tutorial.html#running-python-scripts-with-mpi

AnthonyLim23:
That makes sense. You could do `-n 1` to make it run on one processor, but at that point you may as well use more. 

Maybe something like this could help by allowing you to add a response to the prompt in the script https://stackoverflow.com/questions/10479078/how-to-answer-to-prompts-automatically-with-python-fabric

jess-farmer:
As discussed in yesterday's FitBenchmarking meeting, the Paramonte controller tests are now skipped during Windows tests - this is because there is an issue with Paramonte detecting MPI libraries on Windows (the wrong path is searched for). When a user first imports Paramonte during a python session, they are asked if they want to install MPI libraries if an installation hasn't been detected. This is a problem for testing, as there isn't an obvious way to provide a response to this prompt.

jfowkes (approved):
Looks good to me.

PR #1109, opened 2022-12-23
Add lmfit software #1096

### LINKED ISSUE #1096: Add LMFIT

joshuadkitenge:
#### Description of Work

Fixes #1096

#### Testing Instructions

1.
2.
3.


docs/source/users/options/minimizer_option.rst
```diff
@@ -392,6 +392,72 @@ The `levmar` minimizer is set as follows:
...
+* `Maximum likelihood via Monte-Carlo Markov Chain <https://emcee.readthedocs.io/en/stable/>`__ (:code:`emcee`)
+* Simplicial Homology Global Optimization (:code:`shgo`)
+* Dual Annealing optimization (:code:`dual_annealing`)
+
+**Licence** LMFIT is available the new BSD-3 licence -- details can be found `here <https://lmfit.github.io/lmfit-py/installation.html#copyright-licensing-and-re-distribution>`__
+
+The `lmfit` minimizer is set as follows:
+
+.. code-block:: rst
+   [MINIMIZERS]
```
jess-farmer, line 433: I think a blank line is needed here for the code block to display
```suggestion
.. code-block:: rst

   [MINIMIZERS]
```

fitbenchmarking/controllers/lmfit_controller.py
```diff
@@ -0,0 +1,159 @@
+"""
+Implements a controller for the Ceres fitting software.
```
jess-farmer, line 2:
```suggestion
Implements a controller for the lmfit fitting software.
```

fitbenchmarking/controllers/lmfit_controller.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def setup(self):
+        """
+        Setup problem ready to be run with lmfit
+        """
+
+        if (self.value_ranges is None or np.any(np.isinf(self.value_ranges))) \
+           and self.minimizer in self.bound_minimizers:
+            raise MissingBoundsError(
+                    "Differential evolution requires finite bounds on all"
```
jess-farmer, line 117: This `MissingBoundsError` message should be dependent on the current minimizer, as it applies to both differential evolution and dual annealing

jess-farmer (changes requested):
When running the emcee minimizer, I get the error message `utf-8' codec can't decode byte 0xe2 in position 0: unexpected end of data` - is this an error that you've seen? I have emcee version 3.1.3 installed and have tried running the NIST and simple_tests problem sets.

joshuadkitenge:
> When running the emcee minimizer, I get the error message `utf-8' codec can't decode byte 0xe2 in position 0: unexpected end of data` - is this an error that you've seen? I have emcee version 3.1.3 installed and have tried running the NIST and simple_tests problem sets.

Yeah I've seen this error before. This should be fixed by lines 143 - 144 ([4405454](https://github.com/fitbenchmarking/fitbenchmarking/pull/1109/commits/44054540c65a13721a454b7c9feb973b0f11641c)) on the lmfit controller. I just checked now and it works locally for me. Can you try running it again because we have the same versions of (emcee  3.1.3) ?  also what version of lmfit do you have ? @jess-farmer I have 1.1.0

jess-farmer:
@joshuadkitenge I also have version 1.1.0 of lmfit - I think it must have been a local issue causing the error as I've pulled the latest changes and I'm now now getting the error message. However, I am now getting frequent warning messages from emcee saying `The chain is shorter than 50 times the integrated autocorrelation time for 2 parameter(s). Use this estimate with caution and run a longer chain!`. Having had a quick look at the [lmfit docs for emcee](https://lmfit.github.io/lmfit-py/fitting.html#lmfit.minimizer.Minimizer.emcee), it sounds like emcee is not intended for fitting - I've not yet read what emcee is actually doing/how it works, but might be worth not including it as a default minimizer/adding a warning in the docs that it may not perform well?

joshuadkitenge:
@jess-farmer can you algorithm_check in the controller? just to double check all the minimizers are in the right place and also can you double check the warning message I've written in the docs for emcee? Thanks :)

joshuadkitenge:
@jess-farmer you are right `slsqp` is an jacobian enabled solver so it isn't `deriv_free`. I've changed that in the lasest commit.

jess-farmer (approved):

PR #1120, opened 2023-01-17
Enhancement/883 Add horace parser

### LINKED ISSUE #883: Add SpinW Fits
Related to #492 

Horace is used to analyse data obtained with SpinW, and we should add a parser to allow these to be read into FitBenchmarking.

Instructions from Duc for running fits with SpinW:

> 
> There is currently an alpha release of "pace-python" - a Python module which uses a compiled Matlab library to run Horace and SpinW (both are Matlab programs): https://github.com/pace-neutrons/pace-python/releases - you can find installers in the "Assets" section - they're mostly there to install the Matlab Compiler Runtime (MCR - ~700MB download). The actual Python module is on the test-pypi repo (not actual PyPI yet since it's still a bit unstable: https://test.pypi.org/project/pace-python/ - you could install it with: 
> 
> pip install -i https://test.pypi.org/simple/ pace-python 
> 
> (note that there are only binary wheels for Linux and Windows Python 3.7 specifically [the Py 3.8 wheel doesn't work] - it's a restriction of the MCR and version of Matlab I'm using, R2020a; to create other wheels you have to have the Matlab Compiler SDK toolbox).
> 
> However, after installing the wheel, you need the MCR to run it (and also six and numpy but I think those should be pulled in).
> 
> There's an example ipynb here: https://nbviewer.jupyter.org/github/pace-neutrons/pace-python-demo/blob/main/demo.ipynb
> 
> The syntax for running a fit with Horace is in cell 7 and 8, and integration with SpinW is in cell 9-11.
>

joshuadkitenge:
#### Description of Work

Fixes #883 

#### Testing Instructions

1.
2.
3.


docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -7,6 +7,23 @@ SpinW File Format
...
+Examples of spinw problems are:
+
+.. literalinclude:: ../../../../examples/benchmark_problems/SpinW/1D_Gaussian_1.txt
+
+.. literalinclude:: ../../../../examples/benchmark_problems/SpinW/3D_Gaussian.txt
+
+.. literalinclude:: ../../../../examples/benchmark_problems/SpinW/PCSMO_at_001_data.txt
+
+.. note::
+The SpinW file format requires you to have ran the benchmark problem in Horace
```
oerc0122, line 19:
```suggestion
The SpinW file format requires you to have run the benchmark problem in Horace 
```

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -7,6 +7,23 @@ SpinW File Format
...
+.. literalinclude:: ../../../../examples/benchmark_problems/SpinW/1D_Gaussian_1.txt
+
+.. literalinclude:: ../../../../examples/benchmark_problems/SpinW/3D_Gaussian.txt
+
+.. literalinclude:: ../../../../examples/benchmark_problems/SpinW/PCSMO_at_001_data.txt
+
+.. note::
+The SpinW file format requires you to have ran the benchmark problem in Horace 
+using :code:`fit()` and :code:`simulate()` successfully. Here are some relevant links on 
+how to run `Multifit <https://pace-neutrons.github.io/Horace/unstable/manual/Multifit.html />`__ ,
```
oerc0122, line 21:
```suggestion
using :code:`fit()` and :code:`simulate()` successfully. Relevant links on 
how to run this are: `Multifit <https://pace-neutrons.github.io/Horace/unstable/manual/Multifit.html />`__ ,
```
Looks a bit more professional to me

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
-  e.g.::
-    cut = 'p1_bin=[-1,0.05,1],p2_bin=[-1,0.05,1],p3_bin=[-10,10],p4_bin=[10,20]'
-
-sample
-  The sample defines the target of the experiment and has several parameters.
-
-  This must have a class as "IX_Sample", the remaining arguments are passed
-  through to the IX_Sample class constructor.
-  Full details can be found `here <https://pace-neutrons.github.io/Horace/v3.6.2/user_guide/Resolution_convolution.html#the-tobyfit-class>`__.
+  For SpinW we require an SQW file or a MAT file containing the data from Horace.
```
oerc0122, line 35:
```suggestion
  For SpinW we require a ``.sqw`` or ``.mat`` file containing preprocessed, Horace-compatible data. 
```

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
-
-  This must have a class as "IX_Sample", the remaining arguments are passed
-  through to the IX_Sample class constructor.
-  Full details can be found `here <https://pace-neutrons.github.io/Horace/v3.6.2/user_guide/Resolution_convolution.html#the-tobyfit-class>`__.
+  For SpinW we require an SQW file or a MAT file containing the data from Horace. 
   
-  e.g.::
-    sample = 'class=IX_sample,xgeom=[0,0,1],ygeom=[0,1,0],shape='cuboid',ps=[0.01,0.05,0.01])
+.. note::
+  The MAT file should be used if you are loading in multiple objects. Make sure
```
oerc0122, line 38:
```suggestion
  The ``.mat`` file is the result of using ` ``save(file, sqw_objects)``  <https://uk.mathworks.com/help/matlab/ref/save.html>`__ 
  and should be used if you are loading in multiple ``sqw`` objects. Make sure 
```

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
   
-  e.g.::
-    sample = 'class=IX_sample,xgeom=[0,0,1],ygeom=[0,1,0],shape='cuboid',ps=[0.01,0.05,0.01])
+.. note::
+  The MAT file should be used if you are loading in multiple objects. Make sure 
+  to load the data in appropriately in the `wye_function`. 
+
+function
+  The function is defined by one or two matlab files which returns a SpinW model of the foreground or the foreground and background
+  repectively separated by a semicolon.
```
oerc0122, line 43:
```suggestion
  The function is defined by one or more matlab script (``.m``) files which return a model of the foreground or foreground and background  respectively. 
  These should be given as a semicolon-separated list.
```
Not sure what you mean by "separated by a semicolon", I've guessed here.

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
+
+  The format is again comma seperated key-value pairs, with the matlab files 
+  defined by the variable "foreground" and "background" and the remaining pairs defining starting
+  values as in the native parser.
```
oerc0122, line 47: The format of what? Parameters?
> Parameters should be specified as a comma-separated list of ``key=value`` pairs.

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
 
-  As with sample, details can be found `here <https://pace-neutrons.github.io/Horace/v3.6.2/user_guide/Resolution_convolution.html#the-tobyfit-class>`__.
   
-  e.g.::
-    instrument = 'class=maps_instrument,ei=70,hz=250,chopper=S'
+wye_function
+  The wye_function is defined by a matlab file which returns the `w` (this could be a sqw ,d1d, d2d, d3d and d4d object), 
+  `e` (this is the standard deviation arrays from the objects), `y` (intensity arrays from the objects), along with a  
+  mask array (`msk`) that indicates which elements are to be retained (where elements of msk are true, the corresponding elements of 
+  y and e are retained). This matlab file takes in the path of the datafile and the path of where the matlab functions are located.
```
oerc0122, line 68:
```suggestion
  The wye_function is defined by a matlab file which returns the: 

`w` 
    an ``sqw`` ,``dnd``, ``ix_dataset`` object or ``xye`` struct. (see `Multifit <https://pace-neutrons.github.io/Horace/unstable/manual/Multifit.html />`__)

`e` 
    standard deviation data

`y` 
    intensity data
    
`msk`
    logical mask array of which elements are to be retained
    
This function takes the path to the datafile and the path to the matlab functions as arguments.
```

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
 
-function
-  The function is defined by a matlab file which returns a SpinW model.
+  Explained example of the wye_function:
 
-  The format is again comma seperated key-value pairs, with the matlab file
-  defined by the variable "matlab_script" and the remaining pairs defining starting
-  values as in the native parser.
+  The first three lines adds the path to matlab functions need for fitting and loads the `w` object.
```
oerc0122, line 72:
```suggestion
  The first three lines add the path to matlab functions needed for fitting and loads the `w` object.
```

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
 
-  e.g.::
-    function = 'matlab_script=pcsmo_model.m,J=35,D=0,gam=30,temp=10,amp=300'
+  .. code-block:: rst
```
oerc0122, line 74: in reST the `rst` here (I believe) should be the language of the following code (i.e. matlab if a role has been defined). @AndrewLister-STFC may know better.

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
-  The following example retains only 1% of the data.
+  The next line gets the y and e from the `w` object. These are the true `y` and `e` values from the experiment.    
 
-  e.g.::
-    random_fraction_pixels = 0.01
+  .. code-block:: rst
+    
+    [spinw_y, spinw_e] = sigvar_get(w);
+
+  The next block of code runs :code:`simulate()` once for the sole purpose of getting the msk array
```
oerc0122, line 86: Is this still true? I thought you changed this

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
+    
+    pin=[100,50,7,0,0];
+    forefunc = @mftest_gauss_bkgd;
+    mf = multifit(w);
+    mf = mf.set_fun(forefunc);
+    mf = mf.set_pin (pin);
+    [wout,fitpar] = mf.simulate();
+    [ ~, ~, msk] = sigvar_get(wout);
+  
+  The last two lines of the wye_function applies the `msk` on the `y` and `e` data. As the `e` from retrieved above is the
```
oerc0122, line 98:
```suggestion
  The last two lines of the wye_function applies the `msk` to the `y` and `e` data. As the `e` from retrieved above is the
```

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
+  variance we have squared root the value to get the standard deviation.
```
oerc0122, line 99:
```suggestion
  variance we have taken the square root of the value to get the standard deviation.
```

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
+  forefunc = @mftest_gauss_bkgd;
+  mf = multifit(w);
+  mf = mf.set_fun(forefunc);
+  mf = mf.set_pin(fitpars);
+  [wout,fitpar] = mf.simulate();
+  [spinw_y, e] = sigvar_get(wout);
+  spinw_y=spinw_y(msk);
+
+.. note:: 
+  If the benchmark problem is `tobyfit` or uses monte carlo. A persisent seed needs to be set before simulate is ran.
```
oerc0122, line 130:
```suggestion
  If the benchmark problem uses random numbers in any way (e.g. `tobyfit`). A persisent seed needs to be set before simulate is run.
```

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
-spinwpars
-  SpinW models have some additional parameters which are passed to TobyFit.
-  These can be set for the model by adding them to the file as a new value
-  prefixed with "spinwpars_".
+.. literalinclude:: ../../../../examples/benchmark_problems/SpinW/m_scripts/simulate_functions/fb_simulate_pcsmo_test.m
 
-  It is left to the experienced SpinW user to know what their problem requires
-  in this regard.
+.. note::
+   All the functions needed in the fitting must be in the subdirectory of the benchmark problem.
```
oerc0122, line 159: Is this due to absolute vs. relative pathing or something more fundamental?

docs/source/users/problem_definition_files/spinw.rst
```diff
@@ -15,105 +32,132 @@ software, name, description
...
 
-  Some examples might be:
-  - spinwpars_mat = ['JF1', 'JA', 'JF2', 'JF3', 'Jperp', 'D(3,3)']
-  - spinwpars_hermit = False
-  - spinwpars_optmem = 1
-  - spinwpars_useFast = False
-  - spinwpars_... = ...
+.. note:: 
+  If you have a non standard installation of Horace please set the `HORACE_LOCATION` and the `SPINW_LOCATION`
+  as environment variables(i.e on IDAaaS).
```
oerc0122, line 163:
```suggestion
  as environment variables (e.g on IDAaaS).  
```

docs/source/users/problem_definition_files/spinw.rst
joshuadkitenge: In the SpinW parser we add the benchmark problem directory to the path. If the functions needed for fitting are not global functions in matlab we would expected it to be defined within the sub directory of the benchmark problem. If it is not the run will fail, stating that the functions can't be found.

docs/source/users/problem_definition_files/spinw.rst
joshuadkitenge: This is for the functions and the parameters. e.g.
`foreground=funcF(x1,x2 ,x3), x1=1, x2=2, x3=3  `    **;**         `background=funcB(x4,x5,x6), x4=4, x5=5, x5=5  `

fitbenchmarking/controllers/horace_controller.py
```diff
@@ -86,10 +99,15 @@ def cleanup(self):
...
+
+        # # Allow repeat calls to cleanup without falling over
+        # try:
+        #     self.eng.evalc(
+        #         'if not(any(cellfun(@(x) x=="horace", persistent_vars)));'
+        #         ' horace_off;'
+        #         'end;'
+        #     )
+        # except matlab.engine.MatlabExecutionError:
+        #     pass
```
AndrewLister-STFC, line 113: Is it safe to uncomment this now?
We will need it if someone wants to run a single (non-horace) problem with all matlab solvers

AndrewLister-STFC (changes requested):
We're finally almost there! I've put a few comment in and have tested that I can get results with:
```
fitbenchmarking -p examples/benchmark_problems/SpinW/ -s horace scipy_ls -j default
```
for the 1D problem sets.

I'll try running it with the other datasets when I'm not planning to turn my laptop off after work on Monday.

docs/source/users/problem_definition_files/spinw.rst
joshuadkitenge: I've changed it now.

fitbenchmarking/controllers/horace_controller.py
joshuadkitenge: The reason why I commented this out is because when fitbenchmarking tries to recompute the jacobians at the end. It fails due to horace being turned off.  

Here is the error message:
image

`hpc_config` is set when you turn horace_on().

For now I'll just add logic for it ignore those lines for spinw problems

fitbenchmarking/parsing/spinw_parser.py
```diff
@@ -0,0 +1,198 @@
...
+
+        name = self._entries["name"].replace(' ', '_')
+
+        self._spinw_w = f'w_{name}'
+        self._spinw_msk = f'msk_{name}'
+
+        data_file_path_mat = f"char('{data_file_path}')"
+        spinw_path_mat = f"char('{self._spinw_path}')"
+        eng.evalc(f'[{self._spinw_w}, y, e, {self._spinw_msk}] ='
+                  f'{func_name}({data_file_path_mat},{spinw_path_mat})')
```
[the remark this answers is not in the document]
joshuadkitenge: I tried it with double quotes and it works. I've changed it in the latest commit.

fitbenchmarking/parsing/spinw_parser.py
```diff
@@ -0,0 +1,198 @@
...
+
+        self._spinw_x = x
+        return {'x': x, 'y': y, 'e': e}
+
+    def _create_function(self) -> typing.Callable:
+        """
+        Process the SpinW formatted function into a callable.
+
+        Expected function format:
+        function='matlab_script=filename,p0=...'
```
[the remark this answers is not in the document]
joshuadkitenge: I also added in the expected format if you have a foreground and a background

fitbenchmarking/parsing/spinw_parser.py
```diff
@@ -0,0 +1,198 @@
...
+            bp_names = [k for k in pbg if k != 'background']
+            equations = "foreground: " + \
+                os.path.splitext(os.path.basename(pfg['foreground']))[0] +\
+                "<br>" + 'parameters: ' + ', '.join(p_names) +\
+                "<br>" + "background: " +\
+                os.path.splitext(os.path.basename(pbg['background']))[0] \
+                + '<br>' + 'parameters: ' + ', '.join(bp_names)
+            fg_starting_vals = {n: pfg[n] for n in p_names}
+            bg_starting_vals = {n: pbg[n] for n in bp_names}
+            start_values = [{**fg_starting_vals, **bg_starting_vals}]
```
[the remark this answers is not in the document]
joshuadkitenge: I added in a note in the docs stating that "All parameters must have unique names"

fitbenchmarking/controllers/horace_controller.py
AndrewLister-STFC: I thought that should be guarded by: `if not(any(cellfun(@(x) x=="horace", persistent_vars)));`
Maybe the persistent vars aren't working correctly then

fitbenchmarking/parsing/horace_parser.py
```diff
@@ -0,0 +1,194 @@
...
+                           'that HORACE_LOCATION is specfied as a environment '
+                           'variable')
+    elif "HORACE_LOCATION" in os.environ and \
+         "SPINW_LOCATION" not in os.environ:
+
+        raise ParsingError('Could not parse SpinW problem. Please ensure '
+                           'that SPINW_LOCATION is specfied as a environment '
+                           'variable')
+    else:
+        eng.evalc("horace_on")
```
oerc0122, line 40:
```suggestion
    eng.evalc("restoredefaultpath")

    horace_location = os.environ.get("HORACE_LOCATION", None)
    if horace_location:
        eng.evalc(f"addpath('{horace_location}')")

    try:
        eng.evalc("horace_on")
    except matlab.engine.MatlabExecutionError:
        raise ParsingError('Could not parse SpinW problem. Please ensure '
                           'that HORACE_LOCATION ("{horace_location}") is specfied as a environment '
                           'variable and contains "horace_on.m"')

    spinw_location = os.environ.get("SPINW_LOCATION", None)
    if spinw_location:
        eng.evalc(f"addpath('{spinw_location}')")

    try:
        eng.evalc("spinw_on")
    except matlab.engine.MatlabExecutionError:
        raise ParsingError('Could not parse SpinW problem. Please ensure '
                           'that SPINW_LOCATION ("{spinw_location}") is specfied as a environment '
                           'variable and contains "spinw_on.m"')

```

fitbenchmarking/parsing/horace_parser.py
```diff
@@ -0,0 +1,194 @@
...
+        :type data_file_path: str
+
+        :return: data
+        :rtype: dict<str, np.ndarray>
+        """
+        wye_data_f = self._parse_function(self._entries['wye_function'])
+        path = os.path.join(os.path.dirname(self._filename),
+                            wye_data_f[0]['matlab_script'])
+        eng.addpath(os.path.dirname(path))
+        func_name = os.path.basename(path).split('.', 1)[0]
```
oerc0122, line 75: See [pathlib](https://docs.python.org/3/library/pathlib.html?highlight=pathlib#module-pathlib) for a more modern/less verbose way of doing this.

fitbenchmarking/parsing/horace_parser.py
```diff
@@ -0,0 +1,194 @@
...
+
+        Expected function format:
+        function='foreground=filename,p0=...'
+        or
+        function='foreground=filename,p0=... ; background=filename,q0=...'
+
+        :return: A callable function
+        :rtype: callable
+        """
+        # pylint: disable=attribute-defined-outside-init
```
oerc0122, line 114: Don't disable the warning, just add it as `None` in the init. This lets people more easily see what to expect from your class

fitbenchmarking/parsing/horace_parser.py
```diff
@@ -0,0 +1,194 @@
...
+        if len(self._parsed_func) == 1:
+            pfg = self._parsed_func[0]
+            p_names = [k for k in pfg if k != 'foreground']
+            equations = "foreground: " + \
+                os.path.splitext(os.path.basename(pfg['foreground']))[0] +\
+                "<br>" + 'parameters: ' + ', '.join(p_names)
```
oerc0122, line 120:
```suggestion
                FMT_STR = "{loc}: {f_name}<br>parameters: {p_names}"
                equations = FMT_STR.format(loc="foreground",
                                           f_name=os.path.splitext(os.path.basename(pfg['foreground']))[0],
                                           p_names=', '.join(p_names))
```
And then you can use `FMT_STR` elsewhere (though use better names, I'm lazy.

fitbenchmarking/parsing/horace_parser.py
```diff
@@ -0,0 +1,194 @@
...
+        or
+        function='foreground=filename,p0=... ; background=filename,q0=...'
+
+        :return: A callable function
+        :rtype: callable
+        """
+        # pylint: disable=attribute-defined-outside-init
+        if len(self._parsed_func) == 1:
+            pfg = self._parsed_func[0]
+            p_names = [k for k in pfg if k != 'foreground']
```
oerc0122, line 117: More for your awareness rather than a fix
```suggestion
            p_names = filter(lambda k: k != 'foreground', pfg)
```
`pfg` isn't a terribly informative name

fitbenchmarking/parsing/horace_parser.py
```diff
@@ -0,0 +1,194 @@
...
+        Add set_tbf to the fitting problem.
+        """
+        def set_persistent_vars(path):
+            """
+            Update the persistent_vars from a matlab dump.
+
+            :param path: The file to update from
+            :type path: str
+            """
+            horace_on()
```
oerc0122, line 190: Is this meant to be indented like this?

fitbenchmarking/utils/matlab_engine.py
```diff
@@ -0,0 +1,40 @@
...
+    ENG.evalc(
+        r'if not(exist("persistent_vars"));'
+        r' persistent_vars = {"persistent_vars"};'
+        r'end;'
+    )
+    ENG.evalc(
+        f'if not(any(cellfun(@(x) x=="{name}", persistent_vars)));'
+        f' persistent_vars{{end+1}} = "{name}";'
+        r'end;'
+    )
```
oerc0122, line 25: Why are the first block raw, but the second block not? Need `rf`?

fitbenchmarking/parsing/horace_parser.py
joshuadkitenge: This implementation is not going to work. I specifically wrote function so you don't need to set the `horace_location` and the `spinw_location` if you have a normal installation of horace. This is mention in the docs. Unless you would prefer that the environment variables are required ? I have checked this locally and it doesn't work.

examples/benchmark_problems/Horace/m_scripts/functions/slow_func.m
```diff
@@ -0,0 +1,31 @@
+function y = slow_func (varargin)
```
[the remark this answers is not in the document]
joshuadkitenge: No, I'll change the names of the duplicate functions to be Fitbenchmarking specific (e.g. `slow_func_fb_test`) . I understand that MATLAB uses a global namespace however having all the function in the same space makes benchmark problems easy to debug if they fail. As well prevents future changes in horace affecting these tests.

fitbenchmarking/parsing/horace_parser.py
oerc0122: Yeah, I think it might not work because I left the extraneous `restoredefaultpath` in.

fitbenchmarking/controllers/horace_controller.py
```diff
@@ -41,12 +42,24 @@ def __init__(self, cost_func):
         super().__init__(cost_func)
         self._fit_params = None
 
+    def horace_on(self):
+        """
+        Turning Horace on in matlab
+        """
+        if "HORACE_LOCATION" in os.environ:
+            horace_location = os.environ["HORACE_LOCATION"]
+            self.eng.evalc("restoredefaultpath")
```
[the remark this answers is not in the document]
joshuadkitenge: Yes, however when the fitting function it ran `spinw_on` and `horace_on` are called. I'll add in logical to stop it calling `horace_on` if it is a horace problem as that solves this conflict.

fitbenchmarking/utils/matlab_engine.py
oerc0122: Also 
```
if not(any(cellfun(@(x) x=="{name}", persistent_vars)));
```
Could be
```
if ~ismember("{name}", persistent_vars)
```

fitbenchmarking/parsing/horace_parser.py
joshuadkitenge: I am not confident enough to put this in. This causes too much confusion with how to set up `spinw` as `spinw_on` is not a global variable locally(my pc) but it's global variable on IDAaaS (spinw problems work on both). I might just remove `spinw_on`  as it is not needed to run spinw problems and it is not mention in the SpinW documentation. The path is still needed (https://www.spinw.org/installation/). This would fix the the current error messages, however the  `restoredefaultpath` is needed for IDAaaS. The whole point of this logical is for it to work on both IDAaaS and locally seamlessly. Opinions ?

oerc0122: Ideally this will not be needed for IDAaaS (though admittedly it currently is), but I need to work out what I want to ask them for with regards to not loading default Horace on startup. `spinw_on` adds the required paths, that's its job.

joshuadkitenge: We are adding the required spinw paths manually anyways so `spinw_on` can just be removed then.   I'll probably keep the current implementation for now and create another issue suggesting that the `horace_on` function should to be changed to the code below once the  IDAaaS issue has been resolved.

```
    horace_location = os.environ.get("HORACE_LOCATION", None)
    if horace_location:
        eng.evalc(f"addpath('{horace_location}')")
    try:
        eng.evalc("horace_on")
    except matlab.engine.MatlabExecutionError:
        raise ParsingError('Could not parse SpinW problem. Please ensure '
                           f'that HORACE_LOCATION ("{horace_location}") is'
                           ' specfied as a environment '
                           'variable and contains "horace_on.m"')
    spinw_location = os.environ.get("SPINW_LOCATION", None)
    if spinw_location:
        eng.evalc(f"addpath('{spinw_location}')")
```

oerc0122: Sounds like a reasonable solution. The main reason for the suggestion was just that I think it's easier to parse what's intended and I suspect `else: horace_on` might catch people by surprise

docs/source/users/problem_definition_files/spinw.rst
AndrewLister-STFC: Yes, something like:

```
.. codeblock:: python
```

In this case I guess python is the closest but I doubt syntax highlighting will give any insight so `rst` is probably fine.

fitbenchmarking/parsing/fitbenchmark_parser.py
```diff
@@ -1,10 +1,11 @@
 """
 This file implements a parser for the Fitbenchmark data format.
 """
-from collections import OrderedDict
-
 import os
 import typing
+from collections import OrderedDict
```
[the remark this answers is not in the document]
AndrewLister-STFC: I thihnk this is still used elsewhere in the file, it only shows up as a change because the autoformatter moved it I think. This should probably be included in a new issue like #1122

fitbenchmarking/utils/matlab_engine.py
AndrewLister-STFC: The raw was so it's single `{` so I dont think rf is useful. Although this probably went through a few iterations and it might be cleaner with the double `{{`.

fitbenchmarking/parsing/horace_parser.py
AndrewLister-STFC: Yes, it's defining a dynamic function which is added to the fitting problem.

fitbenchmarking/parsing/fitbenchmark_parser.py
```diff
@@ -235,101 +236,166 @@ def _get_data_problem_entries(self):
...
+        :type value: str
+        :raises ParsingError: If the value has any unexpected characters
+        :return: The parsed value
+        :rtype: int, float, or str
+        """
+        if value.lower() == 'true':
+            return True
+        if value.lower() == 'false':
+            return False
+        for convert in [int, float, str]:
```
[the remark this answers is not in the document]
AndrewLister-STFC: You're right on the string conversion, for the `int`, I think Mantid cares about int vs float so we do need it.

fitbenchmarking/parsing/fitbenchmark_parser.py
```diff
@@ -235,101 +236,166 @@ def _get_data_problem_entries(self):
...
+
+        :param data_file_path: The path to the file to load the points from
+        :type data_file_path: str
+
+        :return: data
+        :rtype: dict<str, np.ndarray>
+        """
+
+        with open(data_file_path, 'r') as f:
+            data_text = f.readlines()
```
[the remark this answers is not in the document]
AndrewLister-STFC: I don't think this should be refactored as part of this pr.

I think this only came up as it was reordered so that methods more likely to be overridden are nearer the top of the file.

I will create an issue from the comment though as the numpy one is probably what we want.

fitbenchmarking/controllers/matlab_opt_controller.py
```diff
@@ -119,4 +119,5 @@ def cleanup(self):
         else:
             self.flag = 2
 
-        self.final_params = self.result[0]
+        self.final_params = np.array(self.result[0],
```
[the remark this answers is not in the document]
joshuadkitenge: The varies different controllers (e.g. scipy, scipy_ls, theseus etc ) return output types are unique to the controller. They need to converted according to a list or a numpy array. The line in question is not recurring in all controllers in fitbenchmarking.

AndrewLister-STFC (approved):

PR #1121, opened 2023-01-18
allow plot scale options #1064

### LINKED ISSUE #1064: Allow plot scale option
At the moment the `format_plot` function has hard-coded that for sasview problems, the plots log-log, and other problems are linear:
https://github.com/fitbenchmarking/fitbenchmarking/blob/c535470981d868af5deb251b69e12340f31f1a46/fitbenchmarking/results_processing/plots.py#L100-L103

This should be refactored so that it's an option that can be passed in with the dataset, rather than individual sources of data being hard-coded into the plotting

joshuadkitenge:
#### Description of Work

Fixes #1064

#### Testing Instructions

1.
2.
3.


AndrewLister-STFC:
Docs should robably be on this page: https://fitbenchmarking.readthedocs.io/en/latest/users/problem_definition_files/native.html
It might also be useful to put the new option in older files -- Muon/Neutron/SASView/etc -- as I know I would copy one of these and edit it if writing a new one...

docs/source/users/problem_definition_files/native.rst
```diff
@@ -75,6 +75,10 @@ function
   Inside FitBenchmarking, this is passed on to the specified software and, as
   such, the format is specific to the package we wish to use, as described below.
 
+plot_scale
```
AndrewLister-STFC, line 78: The below **IVP**, **Mantid**,... is still part of the funtion section above, so this should probably be below or above all of that.

Can you change default to linear in the options as well? We've had a lot of confusion with calling default options "default" in the config file.

Also a very minor point but I'd say x and y rather than y and x.

AndrewLister-STFC (changes requested):
I've just had a quick scan through, did yo ucheck the plots for the loglog/logx files you edited or was it a random smattering?

I can run them later to check too.

joshuadkitenge:
> I've just had a quick scan through, did yo ucheck the plots for the loglog/logx files you edited or was it a random smattering?
> 
> I can run them later to check too.

Yeah it was random, I've changed them back to linear now. All the Sasview problems are "loglog" as that was the format before the refactor.

docs/source/users/problem_definition_files/native.rst
```diff
@@ -69,6 +69,10 @@ input_file
   a number. An example of this can be seen in
   ``examples/benchmark_problems/Data_Assimilation/data_files/lorentz.txt``
 
+plot_scale
+  The scale of the x and y axis for the plots. The options are 'loglog', 'logy', 'logx' and 'linear'. If this 
+  is not set it will default back to default.
```
AndrewLister-STFC, line 74:
```suggestion
  is not set it will default to 'linear'.
```

AndrewLister-STFC (changes requested):
Missed a 'default'. I'll just commit it and merge! :shipit:

AndrewLister-STFC (approved):

PR #1152, opened 2023-04-04
Improve support for missing data in profiler and local_min

### LINKED ISSUE #1082: Poisson problem set with nlls and poisson cost functions- x and y must have same first dimension, but have shapes (5,) and (8,)
**Description of the error**
When running:
`fitbenchmarking -p examples/benchmark_problems/Poisson/ -c nlls poisson`

The following error is produced when producing the output;
` INFO main.py: Producing output for the Poisson problem set`
`ERROR exception_handler.py: Unknown exception. Exiting.`
`ERROR exception_handler.py: x and y must have same first dimension, but have shapes (5,) and (8,)
`

What is the result you expect from running the steps described above?
It to produce the output without an error.

` INFO main.py: Producing output for the Poisson problem set`
`ERROR exception_handler.py: Unknown exception. Exiting.`
`ERROR exception_handler.py: x and y must have same first dimension, but have shapes (5,) and (8,)
`

**Suggested fix**

**Additional details**

AndrewLister-STFC:
#### Description of Work

Performance profiler:
- Allows for differing number of steps per line.
  This could be confusing -- do we need to add any docs about this? Or will this be covered if we do a report on how to read profilers?
  Another thought is we could have separate images for each cost function.

Local min tables:
- If the cost function is incompatible with this table it now displays 'N/A' for the entry.

Todo:
- [x] Add docs in table description for the N/A behavior.

Fixes #1082

#### Testing Instructions

1. Run the example from the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ docs/build
 docs/source/_build
 docs/source/contributors/module_index
 *.log
+*.DS_Store
--- a/fitbenchmarking/results_processing/base_table.py
+++ b/fitbenchmarking/results_processing/base_table.py
@@ -419,18 +419,17 @@ class Table:
         table = self.create_pandas_data_frame(html=False)
         return table.to_csv()
 
-    def get_description(self, html_description):
+    def get_description(self):
         """
         Generates table description from class docstrings and converts them
         into html
-
-        :param html_description: Dictionary containing table descriptions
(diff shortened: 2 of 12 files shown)


jfowkes (approved):
Looks good to me, testing with the example from the issue works and the results table looks as expected.

PR #1161, opened 2023-05-22
Add environmental impact table

### LINKED ISSUE #1158: Add an environmental impact table
Integrate codecarbon, and produce a new table to show which solver has least environmental impact by this metric.

jess-farmer:
#### Description of Work

Fixes #1158 

I've added a new emissions table which shows the average CO2 emissions (over `num_runs`) for each minimizer, measured using the [codecarbon](https://mlco2.github.io/codecarbon/index.html) module.

A `.codecarbon.config` file has also been added to set a few parameters which can then be edited by a user if required. Alternatively, the codecarbon parameters can be set when the tracker is initialized.

#### Testing Instructions

1. Run Fitbenchmarking and check that emissions table is produced
2. 
3.


fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -547,18 +549,20 @@ def perform_fit(controller, options, grabbed_output):
...
-            runtime_list = timeit.Timer(
-                setup=controller.prepare,
-                stmt=controller.execute
-            ).repeat(num_runs, 1)
+            with EmissionsTracker() as tracker:
+                runtime_list = timeit.Timer(
+                    setup=controller.prepare,
+                    stmt=controller.execute
+                ).repeat(num_runs, 1)
+            emissions = tracker.final_emissions / num_runs
```
AndrewLister-STFC, line 565: Do we want to track the prepare call? Currently this is left out of timings.

requirements.txt
```diff
@@ -6,3 +6,4 @@ pylint==2.9.5
 flake8==3.9.2
 pylint-exit==1.2.0
 pandas==1.4.0
+codecarbon==2.2.1
```
AndrewLister-STFC, line 9: This should go in the install_requires section of setup.py -- requirements.txt is just for test/CI deps (and should probably be renamed at some point!)

AndrewLister-STFC (changes requested):
I've just skimmed the code and all looks good. I'll try running next week.

fitbenchmarking/core/fitting_benchmarking.py
jess-farmer: This is a good point - I had set up the emissions tracking this way to avoid having to initialize a new tracker for each run (as each `EmissionsTracker` object can only be started and stopped once), but tracking the prepare call is not ideal - I'll have a look at whether there's a better solution.

jess-farmer: I've made some changes so that a new `EmissionsTracker` is created for each run - but am finding this is causing Fitbenchmarking to run very slowly - on my computer it takes ~5 seconds to create an `EmissionsTracker` object so it really is painfully slow... will continue to think if there's a way to measure only the call to `controller.fit` without instantiating multiple tracking objects

AndrewLister-STFC:
This looks great, I tested it and was a bit disappointed that the table is almost identical to the runtime table. In fact it looks like on WSL code-carbon can't access cpu power usage so the metric is based on an estimated average power consumption from what I can tell.

I'm wondering if we should warn about that and the (hard to find) notes [here](https://mlco2.github.io/codecarbon/methodology.html#cpu) that on windows/mac there's something else to install?

Looking forward to running this on windows directly/some linux box to see some real values!

jess-farmer:
I've not been able to get the emissions tracking running any faster, so I have removed emissions tracking from the regression tests as well as one of the test usage examples. I've also now added in a note about the `Intel Power Gadget` to the docs.

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -553,18 +554,23 @@ def perform_fit(controller, options, grabbed_output):
...
             controller.validate()
+            controller.prepare()
             # Calls timeit repeat with repeat = num_runs and number = 1
             runtime_list = timeit.Timer(
-                setup=controller.prepare,
                 stmt=controller.execute
             ).repeat(num_runs, 1)
+            if controller.track_emissions:
+                # stop emissions tracking after all runs have completed
+                emissions = controller.emissions_tracker.stop() / num_runs
```
AndrewLister-STFC, line 573: Good spot that we can just put prepare on the outside... I guess this means we can put the tracking back here instead of in the controller?

I'll put in a commit so you don't have to change it back again (if you think the commit shouldn't be there we can always git reset it away).

.codecarbon.config
```diff
@@ -0,0 +1,4 @@
+[codecarbon]
+log_level = ERROR
+tracking_mode = process
+save_to_file = False
```
AndrewLister-STFC, line 4: I'm not sure this will work for people installing it (it looks like they support having it in either the home or working directory...). That said I'm not sure there's an easy fix so we should just merge for now and I'll raise an issue to flag it.

AndrewLister-STFC (dismissed):
Happy to merge if this passes

AndrewLister-STFC (approved):

PR #1162, opened 2023-06-01
Add joss badge, remove OS badges

### LINKED ISSUE #1155: Add JOSS badge and CFF file
Add the JOSS badge to our README (might need to generate it from this with shield.io: [DOI](https://doi.org/10.21105/joss.03127)

Add a CFF file: https://citation-file-format.github.io/#:~:text=CITATION.cff%20files%20are%20plain%20text%20files%20with%20human-,CITATION.cff%20files%20is%20the%20Citation%20File%20Format%20%28CFF%29.

AndrewLister-STFC:
#### Description of Work

Fixes #1155 

#### Testing Instructions

1. Check the badges look correct
2. Check the JOSS link works
3. Look for citation info on github!


AndrewLister-STFC:
This also makes releases a bit more complicated as we need to create a draft release to preregister a DOI that we can update the files with for the release?

tyronerees (dismissed):
This looks good to me.  

I'm not sure how to differentiate between the JOSS paper and the software's doi -- both seem to be referenced here (but have different author lists).  I think the best is to merge, and adapt the default isn't what we want/expect.

tyronerees:
> This also makes releases a bit more complicated as we need to create a draft release to preregister a DOI that we can update the files with for the release?

I don't think this is a problem -- the doi is for the latest version of the software, with all the previous versions listed under the same doi

AndrewLister-STFC (commented):
Update zenodo doi

tyronerees (approved):

PR #1163, opened 2023-06-06
Updated default method for computing jacobians and hessians to "analytic"

### LINKED ISSUE #1083: Behaviour of analytic Jacobian/Hessian options
If you run fitbenchmarking on a test set that doesn't support analytic jacobians (or hessians), then the tests do not run:
image

We therefore use `scipy:2-point` as the default jacobian, as this will run on all test problems.

However, it would be better to use the best available (`analytic` if possible, or `scipy:2-point` if not)

I propose changing the behaviour of `analytic` to switch to `scipy:2-point` if analytic Jacobians are not implemented on this data (with a message to the user that this is happening).  This would then allow us to change the default behaviour so that we can use the best derivatives available.

The same should also be done with the Hessians.

RabiyaF:
Closes #1083

#### Description of Work
- [x] Update the default options for Jacobians and Hessians to `analytic`.
- [x] Have a fall back to `scipy` if `analytic` Jacobians are not available.
- [x] Have a fall back to `default` if `analytic` Hessians are not available.
- [x] Update tests and add new tests.
- [x] Update docs.
- [x] Update `_handle_fallback_tags()` to handle edge cases
- [x] Write tests for the edge cases.


fitbenchmarking/core/results_output.py
```diff
@@ -199,6 +202,171 @@ def preprocess_data(results: "list[FittingResult]"):
     return best_results, sorted_results
 
 
+def _handle_fallback_tags(results,
```
AndrewLister-STFC, line 205: I ran an edge case and it broke. I think it's because I've chosen to compare multiple hessians but have kept the jacobian as the default and in the dataset one problem has analytic jacobians whereas the other doesn't. 

I don't know what the best way to fix this function would be. I'll send you the checkpoint file so you don't have to run it unless you want to change something on the data generation side of fb.

ini file
```
[FITTING]
hes_method: scipy
            numdifftools

[MINIMIZERS]
ralfit: hybrid
```

command
```
fitbenchmarking -p fitbenchmarking/test_files/default_parsers_set/ -o test.ini -s ralfit -d -n 1
```

output
```
Unknown exception. Exiting.
  File "/home/alister/venvs/fitbenchmarking/bin/fitbenchmarking-cp", line 33, in <module>
    sys.exit(load_entry_point('FitBenchmarking', 'console_scripts', 'fitbenchmarking-cp')())
  File "/home/alister/fitbenchmarking/fitbenchmarking/cli/checkpoint_handler.py", line 126, in main
    generate_report(args.options_file,
  File "/home/alister/fitbenchmarking/fitbenchmarking/cli/exception_handler.py", line 29, in wrapped
    return f(*args, **kwargs)
  File "/home/alister/fitbenchmarking/fitbenchmarking/cli/checkpoint_handler.py", line 99, in generate_report
    directory = save_results(
  File "/home/alister/fitbenchmarking/fitbenchmarking/utils/write_files.py", line 29, in wrapper
    return function(*args, **kwargs)
  File "/home/alister/fitbenchmarking/fitbenchmarking/core/results_output.py", line 58, in save_results
    best_results, results_dict = preprocess_data(results)
  File "/home/alister/fitbenchmarking/fitbenchmarking/core/results_output.py", line 200, in preprocess_data
    best_results[r][c] = _process_best_results(cat)
  File "/home/alister/fitbenchmarking/fitbenchmarking/core/results_output.py", line 431, in _process_best_results
    if best.accuracy > result.accuracy:
AttributeError: 'NoneType' object has no attribute 'accuracy'
```

AndrewLister-STFC (changes requested):
I found a situation where this doesn't quite work although I'm not sure on the best approach to fixing it. Happy to discuss.

I also had some minor suggestions to make it clearer what's happening in a few of the functions.

fitbenchmarking/core/results_output.py
```diff
@@ -199,6 +202,171 @@ def preprocess_data(results: "list[FittingResult]"):
...
+    jacobian and hessian fallback
+
+    :param results: The list of results to find the tags for and
+                   check for repetition
+    :type results: list[FittingResult]
+    :param sort_order: The sort order of the tags
+    :type sort_order: list[list[str]]
+    :param col_sorting: The components in order of importance that will be
+                        used to generate the col tag.
+    :type col_sorting: list[str]
```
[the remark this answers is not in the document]
RabiyaF: updated the typo. `col_sorting` -> `cat_sorting`

fitbenchmarking/core/results_output.py
RabiyaF: This is a very interesting edge case find. It is worth adding to the tests for results_output(). 

I had one question about the following command. Are you able to run the equivalent of this command (with analytic jacobians) on the master branch ?

`fitbenchmarking -p fitbenchmarking/test_files/default_parsers_set/ -o test.ini -s ralfit -d -n 1`

I am loading the new `checkpoint.json` file and calling `preprocess_data()` _with the new functions on this branch commented out and undoing the changes between line 142 to 151._

sorted_results is still populated with `None` values. So, I wanted to confirm if this command was possible before. If yes, could you kindly share a picture of the output table. I just want to see how it should look.

RabiyaF: I have updated the logic of `_handle_fallback_tags()`. The updated logic is more compact.

**But there is a new issue with the ralfit edge case**

These are the fallback columns in the ralfit checkpoint.json file.

`['ralfit:hybrid:analytic:scipy 2-point',
   'ralfit:hybrid:analytic:numdifftools central', 
   'ralfit:hybrid:scipy 2-point:scipy 2-point', 
   'ralfit:hybrid:scipy 2-point:numdifftools central']`

The renaming convention that we discussed should rename both Jacobin and Hessian to `best_available` in this case. This is exactly what happens. All the tags are renamed as `ralfit:hybrid:best_available:best_available`. 

However, during the subsequent processing in `preprocess_data()`. sorted_results is created with a shape of 3x1 instead of 3x2. _And I think the results are mislabelled because of overwritting._

I have tried rethinking the renaming convention but I can't think of anything that would work for this edge case.

RabiyaF: **This is just for illustration**. The ralfit results without column renaming would have this structure. 

image

The alternative to renaming the columns would be if we handle the `None` values on the frontend. It might make the tables messier if there are a lot of fallbacks. But at least all the information would be displayed on the frontend.

RabiyaF: Can you try this command again with the latest commits ?

`fitbenchmarking -p fitbenchmarking/test_files/default_parsers_set/ -o test.ini -s ralfit -d -n 1`.

I updated the `_handle_fallback_tags()` again. **I don't think we should ever update both the jacobian and the hessian tags to `best_avaliable` at the same time**. This will lead to multiple results mapping to the same columns.

 I have added an additional condition to update just one tag (I have chosen to update the Jacobian tag for now).

AndrewLister-STFC: > These are the fallback columns in the ralfit checkpoint.json file.
> 
> `['ralfit:hybrid:analytic:scipy 2-point', 'ralfit:hybrid:analytic:numdifftools central', 'ralfit:hybrid:scipy 2-point:scipy 2-point', 'ralfit:hybrid:scipy 2-point:numdifftools central']`
> 
> The renaming convention that we discussed should rename both Jacobin and Hessian to `best_available` in this case. This is exactly what happens. All the tags are renamed as `ralfit:hybrid:best_available:best_available`.
> 

In this case (which is why I chose it to test), the hessians are not `best_available`. They are user chosen hessians.

I think the correct result would be 
`['ralfit:hybrid:best_available:scipy 2-point', 'ralfit:hybrid:best_available:numdifftools central', 'ralfit:hybrid:best_available:scipy 2-point', 'ralfit:hybrid:best_available:numdifftools central']`


> I updated the _handle_fallback_tags() again. I don't think we should ever update both the jacobian and the hessian tags to best_avaliable at the same time. This will lead to multiple results mapping to the same columns.

I think we can update both at the same time, we just have to be careful to make sure that it is the case that the used the fallback.


I'll have a look at the update this afternoon. We could always add 2 bools to the fitting result if it's using a jacobian fallback or hessian fallback?

RabiyaF: With the latest changes, this edge case should be processed correctly and should not raise an error. I will also add the tests for this edge case today.

However, If there are other edge cases, then I can have a look at the fallback bools in `FittingResults`. I considered it. I also added them but removed them later because I thought the logic would be more complicated and would require iterating over the list of results multiple times. I abandoned the approach because I could not think of an easy way of doing it at that time but it can be revisited if problems with edge cases persist.

_What I meant about not updating the hessian and jacobian tags at the same time was_ - I don't think that we should every allow a column tag to be renamed as `s1:m1:best_avaliable:best_avaliable`. Because in this case we would end up mapping the results of 4 partially filled columns into 1 column. We would overwrite and possibly mislabel the results.

fitbenchmarking/core/results_output.py
```diff
@@ -199,6 +201,183 @@ def preprocess_data(results: "list[FittingResult]"):
...
+                match_str = ':'. join(["[^:]*" if key == check_tag else
+                                      unpacked_column_tag[key]
+                                      for key in col_order])
+                matches = _find_matching_tags(match_str, fallback_columns)
+                matches_summary[check_tag] = matches
+
+            rename_jac_and_hes = len(matches_summary['jacobian']) == 2 \
+                and len(matches_summary['hessian']) == 2
+
+            if rename_jac_and_hes:
```
AndrewLister-STFC, line 254:
```suggestion
            rename_jac = len(matches_summary['jacobian']) == 1
            rename_hes = len(matches_summary['hessian']) == 1
            rename_both = len(matches_summary['hessian']) == 1
            
            if rename_jac and rename_hes:
```

fitbenchmarking/core/results_output.py
```diff
@@ -199,6 +201,183 @@ def preprocess_data(results: "list[FittingResult]"):
...
+                # If both jacobian and hessian tags can be renamed
+                matches_summary['jacobian'].remove(tag['col'])
+                col_count = all_count[tag['col']]
+                jac_count = all_count[''.join(matches_summary['jacobian'])]
+
+                # If jacobian renaming leads to row count matching
+                if jac_count + col_count == rows_count:
```
AndrewLister-STFC, line 261: One worry with this is that we could have an unfortunate even number:

|problem | j1-h1 | j1-h2 | j2-h1 | j2-h2 | description|
|-|-|-|-|-|-|
|p1 | ❌ | ✔️ | ❌ | ✔️ | hess fallback |
|p2 | ❌ | ✔️ | ❌ | ✔️ | hess fallback |
|p3 | ✔️ | ❌ | ✔️ | ❌ | analytic hess |
|p4 | ✔️ | ❌ | ✔️ | ❌ | analytic hess |

I think this logic would try to combine the j1 columns?

AndrewLister-STFC (commented):
I think this covers the remaining cases where both fallback in at least one problem?

fitbenchmarking/core/results_output.py
RabiyaF: Will incorporate this change.

fitbenchmarking/core/results_output.py
RabiyaF: Yes, It will try to combine the j1 and j2 tags. In this case, we would have two columns tags

- `best_avaliable:h1`
- `best_avaliable:h2`. 

The following part of the code would decide whether it should update the jacobian or the hessians. But the preference is given to the jacobian update.

However, this was an arbitrary choice. We can prioritize hessian renaming so that we end up with

- `j1:best_avaliable`
-  `j2:best_avaliable`

```
# If both jacobian and hessian tags can be renamed
matches_summary['jacobian'].remove(tag['col'])
col_count = all_count[tag['col']]
jac_count = all_count[''.join(matches_summary['jacobian'])]

# If jacobian renaming leads to row count matching
if jac_count + col_count == rows_count:
      # Update jacobian tag
      unpacked_column_tag['jacobian'] = \
            results[result_ix].jacobian_tag = column_rename
else:
       # Update hessian tag
       unpacked_column_tag['hessian'] = \
             results[result_ix].hessian_tag = column_rename
```

In my opinion both are okay. I will try to add a test_case for this as well. However, It is similar to edgecase2.json.  I think it should work even if it is an even number.

fitbenchmarking/core/results_output.py
```diff
@@ -199,6 +201,183 @@ def preprocess_data(results: "list[FittingResult]"):
...
+            elif len(matches_summary['jacobian']) == 2:
+                # Only jacobian tag need to be renamed
+                unpacked_column_tag['jacobian'] = \
+                    results[result_ix].jacobian_tag = column_rename
+
+            elif len(matches_summary['hessian']) == 2:
+                # Only hessian tag need to be renamed
+                unpacked_column_tag['hessian'] = \
+                    results[result_ix].hessian_tag = column_rename
+
```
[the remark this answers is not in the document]
RabiyaF: I did not commit these changes directly because it is causing some of the earlier test_cases to fail.

fitbenchmarking/core/results_output.py
AndrewLister-STFC: I think the problem is we'd need to know where the missing results are (the ❌'s).
So `if (jac_count + col_count == rows_count) and ((results[result_ix] is not None) != (results[fallback_col_ix] is not None))` then do the jac column?

RabiyaF: I think you are right. The logic should automatically identify the right columns to match. In this case `best_avaliable:h1` and `best_avaliable:h2.` is the wrong answer. I have added this edgecase to the tests.

fitbenchmarking/core/results_output.py
```diff
@@ -199,6 +201,287 @@ def preprocess_data(results: "list[FittingResult]"):
...
+    # Find col tags with each sm_tag
+    sm_summary = {}
+    for sm in sm_list:
+        tags = set()
+        for tag in all_result_tags:
+            up = dict(zip(col_order, tag["col"].split(":")))
+            if ':'.join([up['software'], up['minimizer']]) == sm:
+                tags.add(tag['col'])
+        sm_summary[sm] = tags
```
AndrewLister-STFC, line 248: Is sm_list only used to generate the sm_summary? If yes, we can eliminate a loop by not generating this first.

I'd also suggest using tuples (will need to avoid the other places where it is joined and split)

Comparison works just the same with tuples
```
('foo', 'bar') == ('foo', 'bar') #True
('hello', 'world') == ('foo', 'bar') #False
```


```suggestion
    # Find col tags with each sm_tag
    sm_summary = {}
    for tag in all_result_tags:
        up = tag["col"].split(":")
        sm = (up[col_order.index('software')],
              up[col_order.index('minimizer')])
        if sm not in sm_summary:
           sm_summary[sm] = set()
        sm_summary[sm].add(tag['col'])
```

fitbenchmarking/core/results_output.py
```diff
@@ -199,6 +201,287 @@ def preprocess_data(results: "list[FittingResult]"):
...
+            jac_tag = set()
+            hes_tag = set()
+            for tag in tags:
+                up = dict(zip(col_order, tag.split(":")))
+                jac_tag.add(up['jacobian'])
+                hes_tag.add(up['hessian'])
+            update_summary[sm] = 'jacobian' if len(jac_tag) != 1 \
+                and len(hes_tag) == 1 else \
+                ('hessian' if len(jac_tag) == 1
+                 and len(hes_tag) != 1 else 'both')
```
AndrewLister-STFC, line 335: Have I missed something or is this not checking whether any of the columns involved are fallback ones?

fitbenchmarking/core/results_output.py
```diff
@@ -199,6 +201,287 @@ def preprocess_data(results: "list[FittingResult]"):
...
+
+        # Three types of tags for software and minimizer
+        # Update both tags
+        elif len(tags) == 3:
+            update_summary[sm] = 'both'
+
+        # Four types of tags for software and minimizer
+        # Update jacobian, hessian or both tags
+        elif len(tags) == 4:
```
AndrewLister-STFC, line 344: What happens for more tags?
E.g hessian has fallback for some problems and we have the following tags
|jacobian|hessian|
|-|-|
|scipy|scipy|
|scipy|analytic|
|analytic|scipy|
|analytic|analytic|
|numdifftools|scipy|
|numdifftools|analytic|

fitbenchmarking/core/results_output.py
```diff
@@ -199,6 +201,287 @@ def preprocess_data(results: "list[FittingResult]"):
...
+
+                # Unpack the tag
+                unpacked_column_tag = dict(zip(col_order, tag.split(":")))
+
+                # Find jacobian and hessian matches
+                matches_summary = {}
+                for check_tag in ['jacobian', 'hessian']:
+                    match_str = ':'. join(["[^:]*" if key == check_tag else
+                                           unpacked_column_tag[key]
+                                           for key in col_order])
```
AndrewLister-STFC, line 356: If both jac and hessian fallback, we will have `analytic:analytic` and `scipy:scipy`.

In this case, neither of the cols will have matches so we'll also need to check the version that looks like `"[^:]*:[^:]*"`

RabiyaF:
```
        # Two types of tags for software and minimizer
        # Update jacobian, hessian or both tags
        if len(tags) == 2:
            jac_tag = set()
            hes_tag = set()
            for tag in tags:
                up = dict(zip(col_order, tag.split(":")))
                jac_tag.add(up['jacobian'])
                hes_tag.add(up['hessian'])
            update_summary[sm] = 'jacobian' if len(jac_tag) != 1 \
                and len(hes_tag) == 1 else \
                ('hessian' if len(jac_tag) == 1
                 and len(hes_tag) != 1 else 'both')
```

This part of the logic will be able to handle the case where both fallback. I added one of those scenarios to `edgecase6.json`.

- `{'row': 'p1', 'col': 's1:m4:j1:h1', 'cat': 'WeightedNLLSCostFunc', 'result_ix': 12}`
- `{'row': 'p2', 'col': 's1:m4:j1:h1', 'cat': 'WeightedNLLSCostFunc', 'result_ix': 13}`
- `{'row': 'p3', 'col': 's1:m4:j2:h2', 'cat': 'WeightedNLLSCostFunc', 'result_ix': 14}`
- `{'row': 'p4', 'col': 's1:m4:j2:h2', 'cat': 'WeightedNLLSCostFunc', 'result_ix': 15}`

In this case, the col_tags are renamed as  `'s1:m4:best_avaliable:best_avaliable'`.

I did not incorporate the suggested change last time because it was not being used subsequently to set the conditions. But let me see if I can clean up the logic further and simplify it now.

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -503,7 +507,10 @@ def loop_over_hessians(controller, options, grabbed_output, checkpointer):
                 cost_func.hessian = hessian
             except NoHessianError as excp:
                 LOGGER.warning(str(excp))
-                continue
+                LOGGER.info('Using default method instead for hessian')
+                if hes_method == 'analytic':
+                    hes_method = 'default'
+                    cost_func.hessian = None
```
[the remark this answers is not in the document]
RabiyaF: Done.

fitbenchmarking/core/results_output.py
RabiyaF: (I presumed) If tags are longer than 1, then the column experienced fallback. 

I had the following condition at the start of the for loop but removed it because linting was complaining about too many branches. I can add it now because I reduced the number of branches in a recent commit. 

`if len(tags) == 1:
continue
`
       
This function would just tell you for a given software:minimzer tag whether you need to update
- ` jacobian`
- ` hessian ` or
- `both` tags

In some of the earlier commits, I had logic where deciding what to update and the actual updating of values was being done in the same for loop. But I thought the code was getting messy as each of the branches was duplicating the code to update the right values. 

So, I decided to do it in two for loops. 

Once you get the `update_summary`, `_handle_fallback_tags()` updates the tags. 

Let me see if I can explicitly add a check for tag in fallback_columns here as well.

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -437,7 +437,11 @@ def loop_over_jacobians(controller, options, grabbed_output, checkpointer):
                 jacobian = jacobian_cls(cost_func.problem)
             except NoJacobianError as excp:
                 LOGGER.warning(str(excp))
-                continue
+                LOGGER.info('Using Scipy instead for jacobian')
+                if jac_method == 'analytic':
+                    jac_method = 'scipy'
+                    jacobian_cls = create_jacobian(jac_method)
+                    jacobian = jacobian_cls(cost_func.problem)
```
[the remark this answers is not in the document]
RabiyaF: Done.

fitbenchmarking/core/results_output.py
RabiyaF: Added this as edgecase7.json.

fitbenchmarking/core/results_output.py
RabiyaF: I have simplified the logic of ` _find_tag_to_rename()` further.

fitbenchmarking/core/tests/test_results_fallback.py
```diff
@@ -0,0 +1,660 @@
...
+        """
+        Helper function to call _find_non_full_columns
+        and match outputs
+        """
+        actual_column_tags = _find_non_full_columns(columns,
+                                                    expected_count,
+                                                    columns_with_errors)
+        self.assertEqual(actual_column_tags, expected_column_tags)
+
+    def test_find_non_full_columns(self):
```
AndrewLister-STFC, line 583: Can you put these in subtests.
Usually we have 1 class per function and several tests per class.

E.g.
```
class TestNonFullColumns:
    ...
    def test_all_columns_full(self):
        ...
    def test_some_full_columns(self):
        ...
    def no_full_columns(self):
        ...
```

But I think this with subtests is equivalent.
```
def test_non_full_columns(self):
    ...
    with self.subTest('all_columns_full'):
        ...
    ...
```

fitbenchmarking/core/tests/test_results_fallback.py
```diff
@@ -0,0 +1,660 @@
...
+        expected_column_tags = ['s1:m11:j0:',
+                                's1:m11:j1:',
+                                's1:m11:j0:h0',
+                                's1:m11:j1:h1']
+        self.run_and_match_find_non_full_columns(columns,
+                                                 expected_count,
+                                                 columns_with_errors,
+                                                 expected_column_tags)
+
+    def test_handle_fallback_tags(self):
```
AndrewLister-STFC, line 645: What is this unit test testing?
Is it the case where `_handle_fallback_tags` is given no fallbacks?

I think this function could do with multiple unit tests with descriptive names (subtests are fine)

fitbenchmarking/core/tests/test_results_fallback.py
```diff
@@ -0,0 +1,660 @@
...
+                               'cat': 'cf1', 'result_ix': 11},
+                              {'row': 'prob_1', 'col': 's1:m11:j1:',
+                               'cat': 'cf1', 'result_ix': 12},
+                              {'row': 'prob_1', 'col': 's0:m01:j1:',
+                               'cat': 'cf1', 'result_ix': 13},
+                              {'row': 'prob_1', 'col': 's0:m00:j1:',
+                               'cat': 'cf1', 'result_ix': 14}]
+        self.expected_repeating_tags = []
+
+    def test_get_all_result_tags(self):
```
AndrewLister-STFC, line 553: This too could benefit from some more unit tests which explain what functionality the test is trying to confirm.

fitbenchmarking/core/tests/test_results_fallback.py
```diff
@@ -0,0 +1,660 @@
...
+                                                 _find_non_full_columns,
+                                                 _handle_fallback_tags,
+                                                 preprocess_data)
+
+from fitbenchmarking.core.tests.test_results_output import load_mock_results
+
+
+class EdgeCaseTests(unittest.TestCase):
+    """
+    Unit tests for the edge cases.
```
AndrewLister-STFC, line 17: I'm not sure this is a unit test as it's testing 4 functions work together. Would integration tests be more accurate?

fitbenchmarking/core/tests/test_results_fallback.py
```diff
@@ -0,0 +1,660 @@
...
+
+from fitbenchmarking.core.results_output import (_get_all_result_tags,
+                                                 _find_non_full_columns,
+                                                 _handle_fallback_tags,
+                                                 preprocess_data)
+
+from fitbenchmarking.core.tests.test_results_output import load_mock_results
+
+
+class EdgeCaseTests(unittest.TestCase):
```
AndrewLister-STFC, line 15: I think if this fails it will just be called `EdgeCaseTests` in the log - it might be worth choosing a longer name to help highlight which part of the code is being tested for edge cases.

AndrewLister-STFC (changes requested):

fitbenchmarking/core/tests/test_results_fallback.py
RabiyaF: This is the original test I wrote for `checkpoint.json`. It was a longer test. I was manipulating `checkpoint.json ` for another scenario. But that scenario is now covered by one of the edge cases. 

I think I can rename the TestEdgeCase class as you mentioned in a later comment for a more descriptive name and merge the two classes now.

`checkpoint.json` is a test with no fallback columns but it has a result with error_flag = 4. None of the other edge cases have error_flag = 4. It is useful for testing `_find_non_full_columns()` and `_get_all_result_tags()`. So that is why that test existed but I think I can merge the test into one function that tests all the files.

fitbenchmarking/core/tests/test_results_fallback.py
```diff
@@ -0,0 +1,660 @@
...
+    """
+    Unit tests for the edge cases.
+    """
+    def setUp(self):
+        """
+        Defining the edge case results
```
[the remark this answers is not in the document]
RabiyaF: I have updated the class structure and added more detailed doc strings.

fitbenchmarking/core/results_output.py
```diff
@@ -199,6 +201,258 @@ def preprocess_data(results: "list[FittingResult]"):
...
+    Find columns where the number of occurrences is less than the number
+    of rows.
+
+    :param columns: The dict of column tags and their count
+    :type columns: dict[str, str]
+    :param expected_count: The expected results count for each
+                           minimizer
+    :type expected_count: int
+    :param columns_with_errors: The column tags with errors
+    :type expected_count: dict[str, str]
```
AndrewLister-STFC, line 436:
```suggestion
    :type expected_count: dict[str, int]
```

AndrewLister-STFC (changes requested):
Looking good, I think the testing and functionality is all there. I spotted a couple of outdated docstrings (I guess from earlier iterations). And a redundant if (which might have been from a suggestion I gave earlier).

Can you also update the title of the PR to something that would work in release notes please. Then I think we can merge! 🚀

AndrewLister-STFC (approved):

PR #1164, opened 2023-06-06
Move images to Plotly

### LINKED ISSUE #1047: Use plotly instead of matplotlib
Add plotly instead of matplotib in the fitbenchmarking results pages

tyronerees:
#### Description of Work

Fixes #1047 

Todo
- [ ] Change the legend to being inline
- [ ] Tests
- [ ] Add/test support for multiple cost functions
- [ ] Remove matplotlib code
- [ ] Add plotly support for summary fitting plots
- [ ] Add plotly support for performance profiles

#### Testing Instructions

1.
2.
3.

### DIFF
--- a/fitbenchmarking/results_processing/plots.py
+++ b/fitbenchmarking/results_processing/plots.py
@@ -6,6 +6,10 @@ import os
 
 import matplotlib
 import numpy as np
+import plotly
+import plotly.express as px
+import plotly.graph_objects as go
+import pandas as pd
 
 from fitbenchmarking.utils.exceptions import PlottingError
 
@@ -158,13 +162,13 @@ class Plot:
 
                 self.fig.canvas.draw()
 
-    def plot_initial_guess(self):
+    def plot_initial_guess(self,df):
         """
         Plots the initial guess along with the data and stores in a file
 
         :return: path to the saved file
         :rtype: str
-        """
+        """        
         self.plot_data(errors=False,
                        plot_options=self.ini_guess_plot_options,
                        x=self.x,
@@ -173,9 +177,32 @@ class Plot:
         file = f"start_for_{self.result.sanitised_name}.png"
         file_name = os.path.join(self.figures_dir, file)
         self.fig.savefig(file_name)
-        return file
 
-    def plot_best(self, result):
+        # Plotly implementation below
+        fig = px.line(df.query("minimizer in ['Starting Guess']"),
+                      x="x",
+                      y="y",
+                      color="minimizer",
+                      title=self.result.name,
+                      markers=True)
+        # add the raw data as a scatter plot
+        fig.add_trace(go.Scatter(x=df.query("minimizer == 'Data'")["x"].to_list(),
+                                 y=df.query("minimizer == 'Data'")["y"].to_list(),
+                                 mode='markers',
+                                 name='Data',
+                                 marker=dict(symbol='x', color='black')))
+        htmlfile = f"start_for_{self.result.sanitised_name}.html"
+        html_file_name = os.path.join(self.figures_dir, htmlfile)
+        
+        plotly.offline.plot(
+            fig,
+            filename=html_file_name,
+            auto_open=False
+        )
+        
+        return htmlfile
+
+    def plot_best(self, result, df):
         """
         Plots the fit along with the data using the "best_fit" style
         and saves to a file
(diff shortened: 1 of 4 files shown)


fitbenchmarking/results_processing/plots.py
```diff
@@ -263,58 +180,73 @@ def plot_summary(cls, categories, title, options, figures_dir):
...
-                    fig.set_tight_layout(True)
-
-        fname = f'summary_plot_for_{first_result.sanitised_name}.png'
-        fig.savefig(os.path.join(figures_dir, fname))
-        plt.close(fig)
-        return fname
+                    line = cls.summary_best_plot_line \
+                        if result.is_best_fit else cls.summary_plot_line
+
+                    line["color"] = colour
```
AndrewLister-STFC, line 217: `line` is redefined in both branches of the if-else below

fitbenchmarking/results_processing/plots.py
```diff
@@ -263,58 +180,73 @@ def plot_summary(cls, categories, title, options, figures_dir):
...
+                        line = cls.summary_best_plot_line
+                        line["color"] = colour
+                    else:
+                        line = cls.summary_plot_line
+                        transparency = 0.5
+                        line["color"] = 'rgba' + colour[3:-1] + ', ' + \
+                            str(transparency) + ')'
+
+                    plotlyfig.add_trace(go.Scatter(
+                        x=first_result.data_x[first_result.sorted_index],
```
AndrewLister-STFC, line 229: If we're not setting x up before the loop I think it's more readable to just use result
```suggestion
                        x=result.data_x[result.sorted_index],
```

fitbenchmarking/results_processing/plots.py
```diff
@@ -263,58 +180,73 @@ def plot_summary(cls, categories, title, options, figures_dir):
...
+                    ))
+
+                    plotlyfig.update_layout(
+                        title=title
+                        )
+
+                    if result.plot_scale in ["loglog", "logx"]:
+                        plotlyfig.update_xaxes(type="log")
+                    if result.plot_scale in ["loglog", "logy"]:
+                        plotlyfig.update_yaxes(type="log")
```
AndrewLister-STFC, line 244: Can this be un-indented so it's not in every loop? Looks like that was an oversight of the original implementation

fitbenchmarking/results_processing/tests/test_plots.py
```diff
@@ -6,6 +6,7 @@
 import os
 import unittest
 from tempfile import TemporaryDirectory
+import pandas as pd
```
AndrewLister-STFC, line 9: Newline between python libs and 3rd party libs
```suggestion

import pandas as pd
```

fitbenchmarking/results_processing/plots.py
```diff
@@ -68,182 +45,122 @@ def __init__(self, best_result, options, figures_dir):
...
+            fig.update_yaxes(type="log")
+
+        htmlfile = f"start_for_{self.result.sanitised_name}.html"
+        html_file_name = os.path.join(self.figures_dir, htmlfile)
+
+        offline_plot(
+            fig,
+            filename=html_file_name,
+            auto_open=False
+        )
```
AndrewLister-STFC, line 87: Is this the same as [fig.write_html](https://plotly.com/python/interactive-html-export/?_gl=1*8nljbz*_ga*NjUxMzg3MzY1LjE2OTA1MzcxMTU.*_ga_6G7EE0JNSC*MTY5MDUzNzExNC4xLjEuMTY5MDU0MDEzNy4xNi4wLjA.)?

fitbenchmarking/results_processing/plots.py
```diff
@@ -68,182 +45,122 @@ def __init__(self, best_result, options, figures_dir):
...
+                fig = px.line(df.query(f"minimizer == '{minimizer}'"),
+                              x="x",
+                              y="y",
+                              color="minimizer",
+                              title=self.result.name,
+                              markers=True)
+                if not df.query(f"minimizer == '{minimizer}'")["best"].any():
+                    # add the best plot
+                    name = 'Best Fit (' + \
+                        f'{df.query("best")["minimizer"].unique()})'
```
AndrewLister-STFC, line 129: This seems to give an odd name -- I think unique is returning a list:
image

By construction will this only ever have 1 value? If so can we just take the first one? And in that case do we need the above `any()` as that's also just the same as the first value?

AndrewLister-STFC (changes requested):

AndrewLister-STFC:
I found this when looking at results 
image
It looks a bit alarming because it's using sqrt(y) and y~=0.2. Do you think we should only show error bars when it's a user given error? (out of scope for this PR)

fitbenchmarking/core/tests/test_results_output.py
```diff
@@ -179,13 +179,21 @@ def test_create_plots_with_params(self, plot_mock):
...
+        expected_plot_best = "plotly_fit"
+        expected_plotly_fit = {'m00_[s0]': "plotly_fit",
+                               'm01_[s0]_jj0': "plotly_fit",
+                               'm01_[s0]_jj1': "plotly_fit",
+                               'm10_[s1]_jj0': "plotly_fit",
+                               'm10_[s1]_jj1': "plotly_fit",
+                               'm11_[s1]_jj0': "plotly_fit",
+                               'm11_[s1]_jj1': "plotly_fit",
+                               'm00_[s0]_jj0': "plotly_fit",
+                               'm00_[s0]_jj1': "plotly_fit"}
```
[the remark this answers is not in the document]
tyronerees: This breaks the tests - look in the checkpoint file to see which one should be the best fit.

fitbenchmarking/results_processing/plots.py
```diff
@@ -68,182 +45,122 @@ def __init__(self, best_result, options, figures_dir):
...
+        # Plotly implementation below
+        fig = px.line(df.query("minimizer in ['Starting Guess']"),
+                      x="x",
+                      y="y",
+                      color="minimizer",
+                      title=self.result.name,
+                      markers=True)
+        # add the raw data as a scatter plot
+        fig.add_trace(go.Scatter(
+            x=df.query("minimizer == 'Data'")["x"].to_list(),
```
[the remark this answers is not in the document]
tyronerees: Indexing seems to be a bit faster, on a test of two runs: 

| `query`             | image | image |
|---------------|------------|-------------|
| `index`          | image | image |

fitbenchmarking/results_processing/plots.py
tyronerees: Seems to be, and this is much faster for some reason

AndrewLister-STFC: This adds the full plotly lib to each html file.
We should do one of:
- Return a div in this block (then include in the html file). Could use as a string but it would take too much storage.
- Generate html without javascript. Requires adding the css/js to the html pages.

AndrewLister-STFC (approved):

PR #1187, opened 2023-10-04
Added problem size info to results tables, fitting reports and problem summary pages

### LINKED ISSUE #884: Optional extra columns for displaying additional info, such as number of parameters and number of data points in result tables
Discussed with @BenPetersRAL recently, and based on questions from Nick Gould after a presentation by @jess-farmer recently, where Nick was asking from viewing table results what are the number of fitting parameters for each of the Benchmark problems? 

You can get this information by clicking on each of the Benchmark problems. However, here the feature request is to make this optionally more visible. Specifically the feature request is to allow optional extra columns in the tables, which e.g. can be: the number of parameters, number of data points, and so on, although in this issue suggest limit this to: number of parameters and number of data points. 

The main user facing output of FitBenchmarking is the table outputs and the purpose of this issue is the enhance this experience; allowing the user to optionally add extra columns, which are important to them and for when they present results to otherwise/or themselves.

I am unsure how to best implement this. Options include: 1) add options which can be set in `.ini` setting files such as `all_software.ini` and `options_template.ini` 2) allow users to interactively add such columns in table result web pages.

**Describe alternatives you've considered**
For the Benchmark problems we ship, create a document page with CI that displays summary information about the benchmark problems we support. Not as nice as if this info displayed directly in tables, but maybe fine. This would require creating a script for this and maintaining this script.

Letizia97:
## Description of work

Fixes #884

Changes : 
- Problem size information, specifically number of parameters and number of data points, was added to the results tables, fitting reports and problem summary pages.
- In the tables, this information was added in a new header column, to the right of the header showing the problem names.
- A checkbox has also been added that allows to hide or show this new column (by default the column is shown).
- CSS for the tables has been modified to allow the new column to slide behind the header showing the problem names.

## Testing Instructions

1. Running for example "fitbenchmarking -o examples/options_template.ini \
-p examples/benchmark_problems/NIST/low_difficulty" should give tables showing the new problem-size column.
2. The new column can be toggled using the checkbox present next to the dropdowns in the table pages.
3. If the table is large and the table can slide horizontally, then the new column should slide behind the header with problem names.
4. Clicking on problem sizes present in the new column should bring you to the corresponding problem summary page.
5. Fitting reports and problem summary pages should show the number of data points and number of parameters (just before the list of parameters).

## Checklist


fitbenchmarking/results_processing/base_table.py
```diff
@@ -328,10 +329,32 @@ def create_pandas_data_frame(self, html=False):
...
+                              for m, n in zip(n_params, n_data_points)]
+
+        single_index = list(str_results.keys())
+
+        multi_index = pd.MultiIndex.from_tuples(zip(single_index,
+                                                    self.problem_sizes))
+
+        results_only = np.array(list(str_results.values()))
+        reshaped_results = results_only.reshape((len(multi_index),
+                                                 len(multi_columns)))
```
AndrewLister-STFC, line 352: I haven't used numpy arrays in a while but I'd have expected them to maintain the right shape. Did you check that it's not just that the array needs transposing (`results_only.T`)?

fitbenchmarking/results_processing/base_table.py
```diff
@@ -380,15 +403,26 @@ def to_html(self):
...
+        for b, i in zip(self.best_results.values(), self.problem_sizes):
             b = next(iter(b.values()))
             rel_path = os.path.relpath(
                 path=b.problem_summary_page_link,
                 start=self.group_dir)
-            index.append(f'<a class="problem_header" href="{rel_path}">{i}</a>'
-                         )
-        table.index = index
+            index2.append(f'<a class="problem_header_lev1" '
+                          f'href="{rel_path}">{i}</a>')
```
AndrewLister-STFC, line 422: I think both `index1` and `index2` can be updated in the same loop. If we make `index` have entries which are tuples then there's no need to zip. `index.append((f'<a class="problem_header"...', ...))`

fitbenchmarking/utils/fitbm_result.py
```diff
@@ -179,6 +179,24 @@ def __eq__(self, other):
                 return False
         return True
 
+    def get_n_parameters(self):
+        """
+        Returns number of parameters for the result.
+
+        :return: String for number of parameters
+        :rtype: str
```
AndrewLister-STFC, line 187: As written these return ints
```suggestion
        :return: Number of parameters
        :rtype: Int
```

fitbenchmarking/utils/fitbm_result.py
```diff
@@ -179,6 +179,24 @@ def __eq__(self, other):
...
+        """
+        return len(self.initial_params)
+
+    def get_n_data_points(self):
+        """
+        Returns number of data points for the result.
+
+        :return: String for number of data points
+        :rtype: str
```
AndrewLister-STFC, line 196:
```suggestion
        :return: Number of data points
        :rtype: Int
```

fitbenchmarking/results_processing/base_table.py
Letizia97: Just checked this, it looks like reshaping (lines 351 and 252) is not necessary, as I get the same table if I build it directly with the array "result_only". So shall we just remove the reshaping and use "result_only" for building the table?

Letizia97: Also, I tried using results_only.T but that raises an error saying "Shape of passed values is (6, 16), indices imply (16, 6)"

AndrewLister-STFC (approved):
All looks good to me. Let's merge it! 🚀

PR #1189, opened 2023-10-11
Changed rendering of html plotly plots (to reduce size of html files)

### LINKED ISSUE #1183: Change rendering of html plotly plots
Plotly plots introduced in #1164 all embed the full plotly.js in the html file, meaning each file is ~3MB big.

This can be avoided by shipping our own library.

We can also just generate the div needed, rather than the full html page, which would remove the need for an iframe.

This is a useful [StackOverflow post](https://stackoverflow.com/questions/36262748/save-plotly-plot-to-local-file-and-insert-into-html)

Letizia97:
## Description of work

Fixes #1183 .

This PR modifies the function used to save the html files for the plotly plots. The reason is to avoid including the whole plotly.js information within the html file as this considerably increases its size. The new function used to save each html file is write_html which allows to point to a single plotly.js file (also added through this PR). This way all the html files for all figures point to the same plotly.js file, rather than repeating the same info.

## Testing Instructions

1. Run fitbenchmarking and check that the plotly plots are shown.
2. Check the size of the html files for the plotly plots in results/support_pages/figures. Each file should be much smaller than 3000KB (e.g.~10-40KB).

## Checklist


fitbenchmarking/results_processing/plots.py
```diff
@@ -45,6 +44,26 @@ def __init__(self, best_result, options, figures_dir):
...
+        :type fig: plotly.graph_objs._figure.Figure
+        :param figures_dir: The directory to save the figures in
+        :type figures_dir: str
+        :param htmlfile: Name of the figure
+        :type htmlfile: str
+
+        :return: None
+        """
+
+        plotly_path = '../../../js/plotly.js'
```
tyronerees, line 63: Rather than hard link the path, can you use something like we do here:
https://github.com/fitbenchmarking/fitbenchmarking/blob/4757fb1b0fa3ba0c664632d9cf646a2439f5ff31/fitbenchmarking/utils/misc.py#L66C40-L66C82
(maybe even calling that file?)

I can't think of any cases where what you've done will fail, but this protects us from needing to hunt and change lots of hard coded paths if we ever change where the js files live in the future.

tyronerees (changes requested):
Tested locally, and all looks good to me.  Only one minor comment from me -- please let me know if it doesn't make sense.  Thanks, Letizia!

tyronerees (approved):
Thanks!  Sorry it was a lot more work to put put the directory in -- this looks great now 👍

PR #1192, opened 2023-10-23
Replaced matplotlib with plotly as the engine for generating performance profiles

### LINKED ISSUE #1180: Use plotly for performance profiles

Letizia97:
## Description of work

Fixes #1180 .

- Updated the code so that performance profiles are generated using plotly rather than matplotlib.
- Added a warning for the users in the table templates (just below the performance profiles). This is specify that the performance profile plot does not get recalculated when they remove a certain line/minimizer from the graph by clicking on the legend (removing line is now possible thanks to plotly).

## Testing Instructions

Just run fitbenchmarking and check the performance profiles in the table pages. It should be possible to:
1. Zoom in and out
2. Remove traces and add them back in using the legend
3. See points coordinates and minimizer name wheh hovering over the lines

## Checklist

- [x] My PR fully fixes the issue linked. If new issues have been created, what are they? New issue about creating a toggle to allow users to see the performance profile either with a linear x axis or a log x axis (issue #1194 ) .
- [x] I have added the appropriate tests to cover code that has been added (I modified existing tests but not sure whether we need to add more).
- [x] I have updated the documentation in the relevant places to cover the changes  (I modified existing docstrings, don't think we need to add more).

### DIFF
--- a/fitbenchmarking/core/results_output.py
+++ b/fitbenchmarking/core/results_output.py
@@ -61,7 +61,8 @@ def save_results(options, results, group_name, failed_problems,
 
     best_results, results_dict = preprocess_data(results)
 
-    pp_locations = performance_profiler.profile(results_dict, fig_dir)
+    pp_locations = performance_profiler.profile(results_dict, fig_dir,
+                                                supp_dir, options)
 
     if options.make_plots:
         create_plots(options, results_dict, best_results, fig_dir, supp_dir)
--- a/fitbenchmarking/results_processing/performance_profiler.py
+++ b/fitbenchmarking/results_processing/performance_profiler.py
@@ -103,7 +103,7 @@ def plot(acc, runtime, fig_dir):
         if max_value < linear_upper_limit:
             use_log_plot = False
 
-        # Get second largest value and use it to set an x limit on the graph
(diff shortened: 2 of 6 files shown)


AndrewLister-STFC (changes requested):
No comments on the code 👍 

I did spot a few issues:
- [x] When the perf profiles are created my browser opens them in a new tab. This doesn't happen with the other plotly plots so I'd guess it's an easy fix.
- [x] The graphs don't have gridlines or tick marks which can make them hard to follow. (I don't know that we need minor gridlines as it's meant to show trends not exact values but the major ones are handy)
- [x] In some cases the x axis is way too long and means the interesting bit is very squashed. We can zoom in so not a big deal but the ticks also don't extend.
- [x] I think the use of LaTeX script for the axes was neater/more professional. It looks like it should be possible to use `"$10^4$"` for example by following here and adding mathjax to our list of js libs: https://plotly.com/python/LaTeX/

AndrewLister-STFC:
Blocked on #1189 being merged first

AndrewLister-STFC (changes requested):
Looks great, 1 import that might be unused? Otherwise I think it's ready to merge

AndrewLister-STFC (approved):

PR #1200, opened 2023-11-22
Added a more interactive (Dash) version of performance profiles

### LINKED ISSUE #1194: Add toggle to show performance profiles either with linear x axis or log x axis
Once PR #1192 is merged , the performance profiles will be generated in plotly. The x axis will be shown as a log axis in all performance profiles and the user cannot change this (so they won't be able to switch the axis to linear).

It would be good to have a toggle which allows the user to change the x axis in the performance profiles from log to linear and vice versa. If this is done without Dash, probably it would be necessary to create all plots in the background (e.g. 4 plots, of which 2 with log x axis and 2 with linear x axis), and then display only the version the user selects. Probably, it will be easier to do this with Dash, so this is something we could do later when we start to use Dash for the plots.

Letizia97:
## Description of work

Fixes #1194 .

Main changes:
- Added Dash app within fitbenchmarking, to be run after the results are produced. Unless the user stops the app, the app keeps running after fitbenchmarking has finished.
- Added Dash plots to the accuracy and runtime tables - these are only available if the app is running
  - The user can switch between the new Dash plots and the old plotly plots (always available, no matter whether Dash is running), by clicking a button.
  - Dash plots are more interactive than plotly plots - however, in this PR the only added interactivity available in the Dash plots is that the user is able to switch between linear and log x-axis scale.
- Modified docs to inform the user of the changes
- Added option to specify whether to run the app or not (default is yes), which can be done through the command line or options file
- Added option to specify which port to use to run dash.

## Testing Instructions

First run with "fitbenchmarking" only, to check Dash (Dash will run by default). Can check:
1. That you have two buttons (one for Online and one for Offline graphs), and they work
2. That in the Dash plot, one can change the x axis
3. Try stopping Dash to see what happens (changing log axis stops working, and after changing to offline plot, one cannot obtain the online one any more)

Could then check the new options through the command line:
1. --run_dash and --dont_run_dash
2. --port for specifying the Dash port

Then one could check that all works when passing more than one problem set. For example using:
fitbenchmarking   -o examples/options_template.ini -p examples/benchmark_problems/NIST/low_difficulty  examples/benchmark_problems/NIST/average_difficulty .

Could also check that the option --load-checkpoint still works too, as that code was modified too.

## Checklist


AndrewLister-STFC (changes requested):

fitbenchmarking/results_processing/performance_profiler.py
```diff
@@ -225,4 +279,143 @@ def create_plot(step_values: 'list[np.ndarray]', solvers: 'list[str]'):
                        )
             )
 
-    return fig
+    data_df = create_df(all_solvers,
+                        all_solver_values,
+                        all_plot_points)
```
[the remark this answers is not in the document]
Letizia97: Yes, I think I see what you mean. This will need to change when we introduce the recalculation of the values. I left it simple for now because they are not being recalculated. But once we introduce that, we will need to save more values (or in other words bring more of the "raw" values along) as the function that recalculates the values to be plotted will be called directly by the callback. Does this answer your doubt? Or you meant something else?

fitbenchmarking/results_processing/tests/test_performance_profiler.py
```diff
@@ -33,6 +43,30 @@ def load_mock_results():
...
+    """
+
+    with open(html_path, 'r', encoding='utf-8') as f:
+        read_lines = f.readlines()
+
+    processed_lines = []
+    for str_i in read_lines:
+        pattern_for_ids = r"\b((?:[a-z]+\S*\d+|\d\S*[a-z]+)[a-z\d_-]*)\b\w+"
+        line_without_ids = re.sub(pattern_for_ids, '', str_i)
+        pattern_for_src = r"\b(([A-Za-z._-]+[\\/]+){1,}([A-Za-z.\s_-]+))\b"
```
AndrewLister-STFC, line 63: I'm not sure I quite understand the format you're extracting with these. In the first, matching `\S` seems excessive when we know the only symbols will be `-` (as far as I can tell?).

In the second, I think the `{1,}` syntax is the same as `+`? I'm also not sure the source will have any whitespace in the filename?

I'd suggest something like the following is more specific but I don't know what you were testing it against?

```
r'"(?:(?:\d+[a-z]+|[a-z]+\d+)[a-z\d]*-)+[a-z0-9]+"'
and
r'"(?:[A-Za-z_-]+[\\/]+)+[A-Za-z_-]+\.[A-Za-z]+"'
```

fitbenchmarking/results_processing/tests/test_performance_profiler.py
```diff
@@ -99,18 +163,158 @@ def test_correct_prepare_profile_data(self):
...
+
+        Plot.write_html_with_link_plotlyjs(fig=plot,
+                                           figures_dir='',
+                                           htmlfile=output_plot_path,
+                                           options=self.options)
+
+        processed_achieved_lines = remove_ids_and_src(output_plot_path)
+        processed_exp_lines = remove_ids_and_src(expected_plot_path)
+
+        assert set(processed_exp_lines) == set(processed_achieved_lines)
```
AndrewLister-STFC, line 223: Why convert to a set, I'd have thought the order mattered?

AndrewLister-STFC (changes requested):

fitbenchmarking/results_processing/tests/test_performance_profiler.py
Letizia97: I tried the patterns you suggested, I had to change the one for the id to r"(?:(?:\d+[a-z]+|[a-z]+\d+)[a-z\d-]*)+[a-z0-9]+",  but like this looks like it works

AndrewLister-STFC:
I had to run this with an environment variable (maybe it's a wsl thing): `QT_QPA_PLATFORM=offscreen`
I guess since we're not opening a window to house the new plots, it's fine to use this for all of dash in fitbenchmarking?

Something like this at the top of the file (outside of any functions) fixes it. Does that still work on your remote machine?
```
import os.environ
os.environ("QT_QPA_PLATFORM")=offscreen
```

AndrewLister-STFC (approved):

PR #1212, opened 2023-12-08
Added the Fit benchmarking class

RabiyaF:
## Description of work

Fixes #

## Testing Instructions

1.
2.
3.

## Checklist


fitbenchmarking/core/tests/test_fitting_benchmarking.py
```diff
@@ -0,0 +1,824 @@
...
+    controller.parameter_set = 0
+    result = mock_loop_over_hessians_func_call(controller)
+    return result
+
+
+def set_up_controller(file, options):
+    """
+    Sets up the controller for the PerformFitTests
+    """
+    data_file = DATA_DIR + file
```
Letizia97, line 123:
```suggestion
    data_file = os.path.join(DATA_DIR, file)
```

And similarly for all other cases that use DATA_DIR + .. .

fitbenchmarking/core/tests/test_fitting_benchmarking.py
```diff
@@ -0,0 +1,824 @@
...
+
+    def test_benchmark_method(self):
+        """
+        This regression test checks fitbenchmarking with the default
+        software options.The results of running the new class are matched
+        with the orginal benchmark function. The /NIST/average_difficulty
+        problem set is run with scipy_ls software.
+        """
+        results_dir = ROOT + \
+            "/fitbenchmarking/test_files/regression_checkpoint.json"
```
Letizia97, line 758:
```suggestion
        results_dir = os.path.join(TEST_FILES_DIR, "regression_checkpoint.json")
```
We could change the name of  "regression_checkpoint.json" to something that includes the name of the test_file, something like  "fitting_benchmarking_regression_cp.json" or so

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -33,661 +33,544 @@
...
+                dict[str, list[str]]
+        """
+        problem_group = misc.get_problem_files(self._data_dir)
+
+        problems, results, name_count,  = [], [], {}
+
+        for p in problem_group:
+            try:
+                with self.__grabbed_output:
+                    LOGGER.info('Parsing problems')
```
Letizia97, line 99: LOGGER.info could go before the for loop so it's not executed every time

```suggestion
        LOGGER.info('Parsing problems')
        for p in problem_group:
            try:
                with self.__grabbed_output:
```

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -33,661 +33,544 @@
...
+        :return: all results,
+                 problems where all fitting failed, and
+                 minimizers that were unselected due to algorithm_type
+        :rtype: list[fibenchmarking.utils.fitbm_result.FittingResult],
+                list[str],
+                dict[str, list[str]]
+        """
+        problem_group = misc.get_problem_files(self._data_dir)
+
+        problems, results, name_count,  = [], [], {}
```
Letizia97, line 94: I think we don't need to define results here, as afterwards we overwrite it (line 133)

```suggestion
        problems, name_count = [], {}
```

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -33,661 +33,544 @@
...
+                    # are stored in a list i.e. we have multiple results
+                    for i in range(len(accuracy)):
+                        result_args['dataset'] = i
+                        result = fitbm_result.FittingResult(**result_args)
+                        results.append(result)
+                        self._checkpointer.add_result(result)
+                else:
+                    result = fitbm_result.FittingResult(**result_args)
+                    results.append(result)
+                    # self._results.append(result)
```
Letizia97, line 455: As discussed, this comment is leftover from previous work

```suggestion

```

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -33,661 +33,544 @@
...
-                                unselected_minimizers=unselected_minimizers)
+        benchmark_pbar = tqdm(problems, colour='green',
+                              desc="Benchmark problems",
+                              unit="Benchmark problem", leave=True)\
+            if self._options.pbar else problems
 
-    return results, failed_problems, unselected_minimizers
+        name_index = {key: 0 for key in name_count}
 
+        LOGGER.info('Running problems')
```
Letizia97, line 118: This LOGGER.info is repeated twice, here and a few lines above, so we could remove this one

```suggestion

```

Letizia97 (changes requested):
I think the fitting_benchmarking code is much easier to read now thanks to this refactoring. I've added a few suggestions for:
- new tests
- trying to fix the Windows tests
- small changes/deletions in the fitting_benchmarking file

RabiyaF:
@jess-farmer If I revert from the _**explicit object**_ introduced in #1244 back to the _**context manager**_ way of using codecarbon then the Windows test will pass. This [link](https://mlco2.github.io/codecarbon/usage.html) says that the **explicit object** should be preferred in Jupyter notebooks. If I remember correctly - A speed up in generating the emission tables was mentioned as a reason for preferring the _**explicit object**_ style. Do you think it is best to continue with the _**explicit object**_ style ? Or is it possible to change it in the short term ? The only quick fix to get the Windows tests to pass on the runners might be to revert the changes.

jess-farmer:
> @jess-farmer If I revert from the _**explicit object**_ introduced in #1244 back to the _**context manager**_ way of using codecarbon then the Windows test will pass. This [link](https://mlco2.github.io/codecarbon/usage.html) says that the **explicit object** should be preferred in Jupyter notebooks. If I remember correctly - A speed up in generating the emission tables was mentioned as a reason for preferring the _**explicit object**_ style. Do you think it is best to continue with the _**explicit object**_ style ? Or is it possible to change it in the short term ? The only quick fix to get the Windows tests to pass on the runners might be to revert the changes.

I think if we can avoid reverting the changes introduced in #1244 that would be preferable - the speed up achieved from switching from the context manager way of using codecarbon was significant (system tests went from taking ~20 mins to run to ~3 mins). I'll have a look today and see if there's any alternative ways we can use codecarbon where these Windows tests will pass that also isn't painfully slow.

fitbenchmarking/core/tests/test_fitting_benchmarking.py
```diff
@@ -0,0 +1,824 @@
...
+                                              UnknownMinimizerError,
+                                              IncompatibleMinimizerError,
+                                              NoJacobianError,
+                                              NoHessianError,
+                                              ValidationException)
+
+FITTING_DIR = "fitbenchmarking.core.fitting_benchmarking"
+ROOT = os.getcwd()
+DATA_DIR = ROOT + \
+    "/fitbenchmarking/benchmark_problems/NIST/average_difficulty/"
```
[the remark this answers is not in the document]
RabiyaF: I am little apprehensive about converting the `benchmark_problems` folder into a python module just to import it in this way. That folder contains no python code and is only being imported by this test file. I think it might be better to get the files simply this way.

`DATA_DIR = os.path.join(os. getcwd(),
                        'fitbenchmarking',
                        'benchmark_problems',
                        'NIST',
                        'average_difficulty')`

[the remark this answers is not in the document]
Letizia97: I think the main reason for doing it that way (i.e. by adding the init file) was to check whether os.getcwd() was the reason for the Windows test error. If you've found out that the error is due to something else, then I think the way you're suggesting here is fine too

RabiyaF:
Just some additional information about the codecarbon package logs.

The test are working on my **Mac** laptop and the logs look like this:
image

The test are not working on my **Windows** laptop and the logs look like this (The same tests are failing on my Windows laptop as in the GitHub runners with the same error messages):
image

The **Windows** falls back to the estimating power consumption using the constant mode. And when it is trying to prepare the emissions data there is division by zero that is raising the `float division by zero` error msg.

The `ZeroDivisionError` originates from `codecarbon` > `emissions_tracker.py` > `BaseEmissionsTracker` > `_prepare_emissions_data` [here](https://github.com/mlco2/codecarbon/blob/dbc423085c3355754cabe57fb773b24c0b809093/codecarbon/emissions_tracker.py#L600)

This causes the `self.__emissions_tracker.start_task()` to fail but the exception is caught in the function and this why accuracy is set as `inf` and the test assertions fail.

fitbenchmarking/core/tests/test_fitting_benchmarking.py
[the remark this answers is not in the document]
AndrewLister-STFC: Quick one on the difference between `os.getcwd()` and the `import benchmark_problems` implementations is that the latter does not depend on where the script is run:
`cd fitbenchmarking/core; pytest .` would only work with the import option.

You could use a relative path from `__FILE__` (I think that's the right var) which should be a bit more robust than `getcwd` if you didn't want to make it a python module.

AndrewLister-STFC:
I think we can have our cake and eat it too if we just call the `EmissionsTracker.start()` function when we make the object?

jess-farmer:
Looking at [this example](https://github.com/mlco2/codecarbon/blob/dbc423085c3355754cabe57fb773b24c0b809093/examples/mnist_inference.py) on the CodeCarbon repo, there is a `TaskEmissionsTracker` class, which allows for the tracking of individual tasks using the context manager - so I think we could still instantiate the `EmissionsTracker` class in `__init__` (avoiding the overhead of creating a new tracker for each fit), but still use the context manager approach in `__perform_fit`

Something like this could work?
```
task_name = uuid.uuid4().__str__()
with TaskEmissionsTracker(task_name=task_name, tracker=self.__emissions_tracker):
    runtimes = timeit.Timer(
    stmt=controller.execute
    ).repeat(num_runs, 1)
if track_emissions:
    emissions = self.__emissions_tracker._tasks[task_name].emissions_data.emissions
```

RabiyaF:
Thank you @jess-farmer and @AndrewLister-STFC for the suggestions. I tried them both but I could not get it to work. This is one of the temp commit with the [changes](https://github.com/fitbenchmarking/fitbenchmarking/pull/1212/commits/162189fea175d29e088d56120f438ca20a1a0796#diff-9fc78853568b402b0ee9ae340c246fd786fff55fa059deb299acb7ff592e0239) suggested. For now, I have temporarily got the tests to pass by either using **_context manager_** or **_explicit defination_** based on the OS.

PR #1216, opened 2024-01-02
Fixed bug where the emissions table showed a runtime performance profile

### LINKED ISSUE #1196: Performance profiles for emissions
**Description of the error**
When looking at the emissions table I see a runtime pp

We should have a separate emissions pp

**Suggested fix**
Copy code for runtime and accuracy to make emissions a first class citizen

AndrewLister-STFC:
## Description of work

Fixes #1196

Adds a performance profiler for the emissions table (and generalises the code so that extending this will be easier in future).

As far as I'm aware this is a bug that was unexpected so no docs changes needed.

Open questions this raises:
- Are we happy with just having acc and runtime on the compare table page?
  I don't think we should add more to the table but the pp could be put in easily if we wanted to.
- Is it worth doing something differently with the local min?
  Currently it shows acc and runtime pp. It would make more sense to show a bar chart of successes since it's a binary result.

## Testing Instructions

1. Look for the new emissions table after generating the results. It should be similar but marginally different to runtime.

## Checklist

### DIFF
--- a/fitbenchmarking/results_processing/acc_table.py
+++ b/fitbenchmarking/results_processing/acc_table.py
@@ -34,9 +34,8 @@ class AccTable(Table):
         :param group_dir: path to the directory where group results should be
                           stored
         :type group_dir: str
-        :param pp_locations: tuple containing the locations of the
-                             performance profiles (acc then runtime)
-        :type pp_locations: tuple(str,str)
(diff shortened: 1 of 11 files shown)


jess-farmer (approved):
All looks good to me - thanks for the fix! As discussed in this morning's meeting, new issues will be created to address the open questions raised in this PR.

PR #1218, opened 2024-01-15
Introduced a limit to the number of solvers in performance profile plots.

Letizia97:
## Description of work

Fixes # 976

This PR limits the number of solvers shown in performance profiles.
When the number of solvers is higher than a certain number (15), the offline plots is not shown and the user is asked to rely on the Dash plot. The Dash plot will only allow to select 15 of the solvers that have been used, and they can be selected in the dropdown.

## Testing Instructions

1. Run fitbenchmarking with a large number of solvers (>15) to test.
2. After running fitbenchmarking, in the tables, the offline plot should not appear. There should be a red warning asking the user to rely on the offline plot.
3. After switching to offline plot, one should see that a maximum number of 15 solvers are shown. Under the dropdown, there should be a warning asking to unselect some solvers to select others, if needed. When the max number of solvers are shown in the plot, the others should appear as greyed out.
4.  Also the writing that generally appears under all performance profile plots should disappear when no offline plot is shown (and reappear when using the dash plot).

## Checklist


fitbenchmarking/templates/table_template.html
```diff
@@ -139,44 +139,48 @@ <h2 class="handle">
...
                                         <div>
-                                        <button class="btn default" id="offline_plot{{ index }}"
-                                                data-value1="{{ offline_plot }}" data-value2="{{ index }}"
-                                                onclick="load_src(this)">Offline version</button>
-
-                                        <button class="btn default" id="online_plot{{ index }}"
-                                                data-value1="{{ dash_plot }}" data-value2="{{ index }}"
-                                                onclick="load_src(this)">Online version</button>
-                                        </div>
+                                            <div>
```
AndrewLister-STFC, line 148: Is this `<div>` spare?

AndrewLister-STFC (changes requested):
This looks nice, I'm a bit cautious about the change to pylint settings. What was the offending block of code?

I also need to go through the js a bit more to double check I understand it but I think it's all good.

I merged the other PR and now there's some conflicts.

fitbenchmarking/templates/table_template.html
Letizia97: I updated this part of code now that I rebased

Letizia97: Also, the change to pylint settings is for the changes in tables.py , where I'm adding the variable n_solvers_large around line 122 (so that means 31 local variables rather than 30). Not sure whether there's another way around that

fitbenchmarking/templates/table_template.html
```diff
@@ -139,44 +139,50 @@ <h2 class="handle">
...
+                                        <br>
+                                        <br>
+                                        <b>
+                                            The plot is such that lines can be removed by clicking on the respective minimizer name in the legend.
+                                            However, the whole graph highly depends on the best minimizer (the one in the top left corner), that is,
+                                            each line is computed with respect to the best minimizer and should only be compared to the latter (rather
+                                            than to any other line in the graph). Currently, if a line is removed, the graph is not recalculated, which
+                                            means care should be taken in interpreting results after having removed lines.
+                                        </b>
+                                    </p>
```
AndrewLister-STFC, line 185: I think your editor put in some indents?
```suggestion
                                <p id="profiles_info" style="display:block">
                                    Fits are taken from all benchmarks, so if FitBenchmarking is run with
                                    ``n`` problems and ``m`` cost functions, the resulting profile plots will have
                                    ``n*m`` steps on the y-axis. <br>
                                    This is a Dolan-Moré performance profile.
                                    The solvers appearing in the top left corner may be considered the best
                                    performing on this test set.
                                    See <a href="https://doi.org/10.1007/s101070100263">Dolan and Moré (2001)</a> for more information.
                                    <br>
                                    <br>
                                    <b>
                                        The plot is such that lines can be removed by clicking on the respective minimizer name in the legend.
                                        However, the whole graph highly depends on the best minimizer (the one in the top left corner), that is,
                                        each line is computed with respect to the best minimizer and should only be compared to the latter (rather
                                        than to any other line in the graph). Currently, if a line is removed, the graph is not recalculated, which
                                        means care should be taken in interpreting results after having removed lines.
                                    </b>
                                </p>
```

AndrewLister-STFC (commented):
This looks great. I'll run some manual testing in the morning and merge if nothing breaks! 🚀

AndrewLister-STFC (approved):

PR #1230, opened 2024-01-23
Refactored performance profilers

AndrewLister-STFC:


AndrewLister-STFC:
@Letizia97 This has a few changes compared to the main one:
- Single set of online/offline buttons
- Single set of dash controls
- More generic performance profiler code

Let's chat about if we want to keep any of these aspects...

A couple of cons:
- The height code would need to be altered to handle having multiple plots. We could probably do this when we define the iframes.
- For multiple plots the controls get lost off the top of the screen when you scroll down. Not sure how big of a UI problem this is but something to note.

PR #1231, opened 2024-01-26
Added support for Mantid Jacobians and created a Mantid development parser 

### LINKED ISSUE #1228: Mantid improvements

At present fitbenchmarking gives a significant bias towards mantid minimizers if you use a mantid function. This defeats the object of having a fair comparison.

In the mantid parser https://github.com/fitbenchmarking/fitbenchmarking/blob/master/fitbenchmarking/parsing/mantid_parser.py#L111 gives mantid a shortcut. Commenting it out levels the playing field. We could have both dev and user versions.

The other advantage mantid minimizers have is that they are the only ones that can get Jacobians for mantid problems. We can get the Jacobian from mantid using the script:

```
from mantid.simpleapi import *
from mantid.api import FunctionDomain1DHistogram
# these will always be the same
N_ params = 3
x_values = [0, 1, 2, 3, 4]
domain = FunctionDomain1DHistogram(x_values)

# create a fit function with the correct parameters
g = Gaussian(Height=7.5, Sigma=1.2, PeakCentre=10)

Jac = g.functionDeriv(domain)

for i in range(len(x_values)-1):
    for j in range(N_params):
        print(i, j, Jac.get(i,j))


```

AnthonyLim23:
## Description of work
This PR adds Jacobians to the mantid parsers. However, it does not work for multistart problems. 

It also adds a new `mantiddev` parser, which removes the advantage from the mantid software when evaluating mantid functions. This does not remove the advantage from mantid jacobians. 
Fixes #1228 

## Testing Instructions

I can send some test problems that use both mantid and mantiddev. It also shows a comparison with nist problems and has similar order of results. 

## Checklist


fitbenchmarking/parsing/mantiddev_parser.py
```diff
@@ -0,0 +1,236 @@
...
+        for i in range(self._N_x):
+            for j in range(len(self._params_dict.keys())):
+                self._jac[i, j] = J.get(i, j)
+        return self._jac
+
+    def _update_params(self, *p):
+        update_dict = dict(zip(self._params_dict.keys(), p))
+        self._params_dict.update(update_dict)
+        return self._params_dict
+
```
AndrewLister-STFC, line 109: Is this just left over from something you tried but didn't use in the end?
Or is it used somewhere I've missed?
```suggestion
```

fitbenchmarking/parsing/mantiddev_parser.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        for param, key in zip(*args, self._params_dict.keys()):
+            self._mantid_function[key] = param
+        # get mantid Jacobian
+        J = self._mantid_function.functionDeriv(self._cache_x)
+        # set np Jacobian values
+        for i in range(self._N_x):
+            for j in range(len(self._params_dict.keys())):
+                self._jac[i, j] = J.get(i, j)
+        return self._jac
```
AndrewLister-STFC, line 103: I'm a bit concerned about using the same object here. If we call this twice on 2 different inputs both will be the jacobian of the second input?

```
x1 = ...
j1 = fitting_prob.jacobian(x1, params)
x2 = ...
j2 = fitting_prob.jacobian(x2, params)

j1 == j2
```

Will this be an issue on finite difference hessians?

AndrewLister-STFC (commented):
Looks good. 1 question about the cached jacobians, but I think that's it

fitbenchmarking/parsing/mantiddev_parser.py
AnthonyLim23: Is there a way to test finite difference Hessians? I can just try it and see if everything goes crazy. 

I have done it this way so we dont need to create a new Jacobian every time. I could change that or use a deepcopy

AndrewLister-STFC: If you run with `software=scipy` or `ralfit` and set `hes_method=numdifftools` or `scipy`. It makes sense to save the cost of allocating a new array but it might not be worth it here as they'd have to be copied later anyway.

AndrewLister-STFC: In the table the one column with the awful times is using the fd hessians. I suspect, it's finding that H=0 so just doing something like gradient descent which we'd expect to perform poorly.

It's probably a hole in the docs that we don't say which solvers are able to use hessians and which aren't. We also need to have a better description of the column titles (a `j:...` indicates the jacobian used, if it's omitted, no jacobian was given to the solver. Similarly for hessians with `h:`)

For scipy we have:
```
    hessian_enabled_solvers = ['Newton-CG',
                               'trust-ncg',
                               'trust-exact',
                               'trust-krylov',
                               'trust-constr',
                               'dogleg']
```

AnthonyLim23: I just changed the code to
```
        _jac = np.zeros((self._N_x, len(self._params_dict.keys())))
        for i in range(self._N_x):
            for j in range(len(self._params_dict.keys())):
                _jac[i, j] = J.get(i, j)
        return _jac
```
and the run time is slower! I got 95.77 and 160.1 for mantid and mantidev respectively, for the Newton-CG: J:an analytic h:numdifftools central

AnthonyLim23: The slow down is dues to generating the hessian from scratch every time. However, even the NIST problems are very slow. This combined with the above suggests that the mantid jacobians are being copied (the slowness is similar to  NIST's)

AndrewLister-STFC: Yeah it must be being copied somewhere! I think I'll add a comment so if a new hessian is added and doesn't copy, it's a bit easier to track down.

I'll put the new parser on the regression tests too.

AndrewLister-STFC:
All working now apart from that I can't use lru cache with BiComplex... will have to have a think about how best to do this.

AndrewLister-STFC:
Seems like the caching isn't needed in numdifftools (identical accuracy tables and these runtimes):
image
vs
image

AnthonyLim23:
> @AnthonyLim23 I think this is ready? Can you check my understanding is right on the docs update I added: https://fitbenchmarking--1231.org.readthedocs.build/en/1231/users/options/jacobian_option.html

yep. Looks good to me.

PR #1239, opened 2024-01-31
Switch to pyproject.toml

### LINKED ISSUE #952: Change build system to use newer pyproject.toml approach

AndrewLister-STFC:
## Description of work

Fixes #952 

## Testing Instructions

1. Check the install in a new venv

## Checklist


pyproject.toml
```diff
@@ -0,0 +1,150 @@
...
+[tool.pylint."messages control"]
+disable = [
+    "too-many-instance-attributes",
+    "invalid-name",
+    "too-few-public-methods",
+    "duplicate-code",
+]
+
+[tool.pylint.reports]
+output-format = 'colorised'
```
AndrewLister-STFC, line 144: American spelling!
```suggestion
output-format = 'colorized'
```

pyproject.toml
```diff
@@ -0,0 +1,150 @@
...
+    'configparser',
+    'codecarbon<2.2.4',
+    'dash',
+]
+
+[project.optional-dependencies]
+bumps = ['bumps>=0.9.0']
+DFO = ['DFO-LS', 'dfogn']
+gofit = ['gofit']
+gradient_free = ['gradient-free-optimizers']
```
AndrewLister-STFC, line 43:
```suggestion
"gradient_free" = ['gradient-free-optimizers']
```

tyronerees (commented):
This fails for me in linux with the editable command:
image
(using python 3.8.10, in case it matters).  Does this mean we can't install things that are editable? Or just that the docs need updating? Or is it just that my python is too old (and hence needs a minimum version?)

In windows, this gets further (Python 3.12.1), but then the install also fails. However, I think that is independent of this PR: 
image

AndrewLister-STFC:
For the editable install, I think it needs `pip>=21.3`. If you already have that, I'll have to dig a bit more!

Not sure what's going wrong on the windows one, but maybe it's the same issue?

tyronerees:
I get the same issue on linux as windows -- Python 3.12 doesn't seem to be compatible with our requirement that `numpy<1.24`.  do you remember why we put that in?

PR #1243, opened 2024-02-02
Added recalculation of performance profiles

### LINKED ISSUE #1217: Recalculate performance profiles when deselecting solvers in Dash
When selecting/deselecting solvers in the Dash performance profiles, the plots should be recalculated.

Letizia97:
## Description of work

Fixes #1217 .

This PR is to allow performance profiles to be recalculated when deselecting or deselecting a solver from the dropdown.
In the compare table one dropdown allows to control both graphs.

## Testing Instructions

1. Run fitbenchmarking
2. Try and deselect the one that looks like the best solver from performance profiles - The rest of the lines should get recalculated
3. Try and deselect all solvers and then re-adding them back - nothing should break

## Checklist

- [x] My PR fully fixes the issue linked. If new issues have been created, what are they? New issue created  #1246 .


fitbenchmarking/results_processing/performance_profiler.py
```diff
@@ -360,15 +409,26 @@ def __init__(self, profile_name, pp_df, group_label):
...
+        self.current_styles = {}
+        self.avail_styles = {}
+
+        self.colors = ['#1f77b4', '#ff7f0e', '#2ca02c', '#d62728', '#9467bd',
+                       '#8c564b', '#e377c2', '#7f7f7f', '#bcbd22', '#17becf']
+        self.linestyles = ['solid', 'dash', 'dashdot']
+
+        for comb in list(itertools.product(self.linestyles, self.colors)):
+            comb_id = comb[0]+comb[1]
+            self.avail_styles[comb_id] = comb
```
AndrewLister-STFC, line 430: I'm not sure I see the benefit of this being a dict, now that I reread it. I'll put a few suggestions in and hopefully you can see what I mean.
```suggestion
        self.avail_styles = list(itertools.product(self.linestyles, self.colors))
```

fitbenchmarking/results_processing/performance_profiler.py
```diff
@@ -382,17 +442,105 @@ def layout(self):
...
+        if len(previous_solvers) == 0:
+            for i, solver in enumerate(solvers):
+                linestyle = self.linestyles[(i % len(self.linestyles))]
+                linecolor = self.colors[(i % len(self.colors))]
+                self.current_styles[solver] = {
+                    'linestyle': linestyle,
+                    'linecolor': linecolor,
+                }
+                comb_id = linestyle+linecolor
+                del self.avail_styles[comb_id]
```
AndrewLister-STFC, line 475: We can pop the next available style each time we need one rather than choosing it and then removing it.
```suggestion
                comb = self.avail_styles.pop()
                self.current_styles[solver] = comb
```

fitbenchmarking/results_processing/performance_profiler.py
```diff
@@ -382,17 +442,105 @@ def layout(self):
...
+            newly_added_solvers = set(solvers).difference(previous_solvers)
+
+            for solver in newly_added_solvers:
+                chosen_style = list(self.avail_styles.values())[-1]
+                self.current_styles[solver] = {
+                    'linestyle': chosen_style[0],
+                    'linecolor': chosen_style[1],
+                }
+                comb_id = chosen_style[0]+chosen_style[1]
+                del self.avail_styles[comb_id]
```
AndrewLister-STFC, line 489: I'm also not sure that the `current_styles` need to be a dict?
```suggestion
                chosen_style = self.avail_styles.pop()
                self.current_styles[solver] = chosen_style
```

fitbenchmarking/results_processing/performance_profiler.py
```diff
@@ -382,17 +442,105 @@ def layout(self):
...
+        # If a solver has been removed
+        elif len(solvers) < len(previous_solvers):
+
+            solvers_to_remove = set(previous_solvers).difference(solvers)
+
+            for solver in solvers_to_remove:
+                linestyle = self.current_styles[solver]['linestyle']
+                linecolor = self.current_styles[solver]['linecolor']
+                self.avail_styles[linestyle+linecolor] = (linestyle, linecolor)
+                self.current_styles.pop(solver)
```
AndrewLister-STFC, line 500: `pop` gives the values and removes it at the same time. 
```suggestion
                comb = self.current_styles.pop(solver)
                self.avail_styles.append(comb)
                
```

fitbenchmarking/results_processing/performance_profiler.py
```diff
@@ -436,10 +582,10 @@ def create_graph(self, x_axis_scale, solvers):
                     mode='lines',
                     line={
                         "shape": 'hv',
-                        "dash": linestyles[(i % len(linestyles))],
-                        "color": colors[(i % len(colors))]
+                        "dash": self.current_styles[solver]['linestyle'],
+                        "color": self.current_styles[solver]['linecolor'],
```
AndrewLister-STFC, line 586: This is then a little less obvious but I think it's neater in general?
```suggestion
                        "dash": self.current_styles[solver][0],
                        "color": self.current_styles[solver][1],
```

fitbenchmarking/results_processing/tests/test_performance_profiler.py
```diff
@@ -313,15 +303,61 @@ def test_get_plot_path_and_data_returns_dict_for_data(self):
...
-            get_plot_path_and_data(bounds,
-                                   self.fig_dir,
-                                   self.options)
+        paths = performance_profiler.get_plot_path(bounds,
+                                                   self.fig_dir,
+                                                   self.options)
+        expec_paths = {}
+        for name, _ in bounds.items():
+            this_filename_html = os.path.join(self.fig_dir,
+                                              f"{name}_profile.html")
```
AndrewLister-STFC, line 312: I mentioned this but just so everything is on the review, I'd be careful that tests aren't the same code as what they're testing. This should work (it's quite simple) but it may be better to write out the expected paths in full.

fitbenchmarking/templates/table_template.html
```diff
@@ -174,14 +174,10 @@ <h2 class="handle">
...
-                                        However, the whole graph highly depends on the best minimizer (the one in the top left corner), that is,
-                                        each line is computed with respect to the best minimizer and should only be compared to the latter (rather
-                                        than to any other line in the graph). Currently, if a line is removed, the graph is not recalculated, which
-                                        means care should be taken in interpreting results after having removed lines.
-                                    </b>
+                                    The plot is such that lines can be removed by clicking on the respective minimizer name in the dropdown
+                                    or hidden by unselecting the minimizer in the legend. If a solver is removed through the dropdown,
+                                    the graph gets recalculated based on the best minimizer among the ones shown. On the other hand, if a solver
+                                    is unselected in the legend, the line gets temporarily hidden and the plot is not recalculated.
                                 </p>
```
AndrewLister-STFC, line 181: Not sure if we need to be more explicit... what do you think?
```suggestion
                                    The plot is such that lines can be removed by clicking on the respective minimizer name in the dropdown
                                    or hidden by unselecting the minimizer in the legend. If a solver is removed through the dropdown,
                                    the graph gets recalculated based on the best minimizer among the ones shown. On the other hand, if a solver
                                    is unselected in the legend, the line gets temporarily hidden and the plot is not recalculated.
                                    We recommend only analysing graphs where lines are removed via the dropdown.
                                </p>
```

AndrewLister-STFC (commented):
I've put a few suggestions in, let me know if you disagree or want to discuss.

I think the `avail_styles` could be a different datatype -- the key is essentially the same as the value (and since tuples can be used as dict keys, it could be the same). I've suggested a standard list with `append` and `pop` to treat it like a stack. It might be worth defining it as `reversed` depending on the order we want in the final plot!

fitbenchmarking/results_processing/performance_profiler.py
Letizia97: I think I'm using that when I do : 

        previous_solvers = list(self.current_styles.keys())
        if len(solvers) > len(previous_solvers):
            .....

Although it could be written differently

AndrewLister-STFC (approved):
This looks great, makes the table pages way more useful for the reports!

AndrewLister-STFC:
Failed test is a change in dependencies and out of scope for this PR.

PR #1244, opened 2024-02-02
Added code carbon task-based tracking

### LINKED ISSUE #1241: Update emissions tracking to use new codecarbon features
The emissions tracking in FitBenchmarking is currently verrrry slow due to the fact that a new `EmissionsTracker` object is created for each minimizer. A new version of codecarbon now includes a task manager to track emissions of small pieces of code [outlined here](https://mlco2.github.io/codecarbon/usage.html#explicit-object) - we should use this in Fitbenchmarking.

jess-farmer:
## Description of work

Fixes #1241 

## Testing Instructions

1.
2.
3.

## Checklist


fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -156,14 +162,19 @@ def loop_over_benchmark_problems(problem_group, options, checkpointer):
...
                                           options=options,
                                           grabbed_output=grabbed_output,
-                                          checkpointer=checkpointer)
+                                          checkpointer=checkpointer,
+                                          emissions_tracker=emissions_tracker)
             results.extend(problem_results)
             failed_problems.extend(problem_fails)
 
+            if emissions_tracker:
+                _ = emissions_tracker.stop()
```
AndrewLister-STFC, line 171: Is this only needed after a `start()`?
We can probably run it after the `try` block, where we clear runtime and `clean_matlab` on ~ln660.

setup.py
```diff
@@ -30,7 +30,7 @@
                       'pandas>=1.3',
                       'jinja2',
                       'configparser',
-                      'codecarbon<2.2.4',
+                      'codecarbon',
```
AndrewLister-STFC, line 33: Is it worth putting a `>=x.x.x` on this?

fitbenchmarking/core/fitting_benchmarking.py
jess-farmer: I followed the [example in the codecarbon docs](https://mlco2.github.io/codecarbon/usage.html#explicit-object) which runs a final `stop()` after the emissions for each task have been tracked. I think moving this to where runtime/matlab are cleared would mean that a new `EmissionsTracker` object would have to be created for the next fit?

AndrewLister-STFC: Ahh yeah, I was thinking of `task_stop()`. So this calls the final `stop()` after each problem?
Not sure if you need to call `stop_task()` before you can `start_task()` again but if there's an error in the try block, then the task won't stop.

jess-farmer: Ah yeah that's a good point - I'll take another look at this

jess-farmer: I've added an extra line after the calls to `clear_matlab` and clearing runtime which stops the emissions tracker task if the `emissions_tracker` has been created but the value for `emissions` hasn't been set

PR #1269, opened 2024-03-22
Added support for SScanSS inverse kinematics problems

### LINKED ISSUE #1040: Add SScanSS data parser
SScanSS is a slightly different problem to the standard FitBenchmarking problem as follows:

Given a robot , parameters , and a target , solve .

Functionally, the robot is considered to be a series of connected joints, either rotational of translational, with fixed axes and bounds on the range of motion, where the parameters represent the amount of rotation or translation. As joints earlier in the series move, this will affect all of the joints' positions further down the series. The solution is found when the position of the last joint of the robot reaches the desired position and orientation - the target position. Any position in the system is represented by a 3-vector for location and a quaternian for orientation. There may be multiple solutions but any will suffice.

One additional note is that the error being minimised is a slightly more complicated function where the location is differenced as expected, but the orientation is not a simple subtraction -- the rotation is first converted to 3 scalar rotations along the axes then the difference is taken.

The main difference here is that we have no input data to the robot, and that the output from f is multidimensional. For example, a usual fitbenchmarking problem (in the simplest case) would be:

Where x and y have the same number of elements.

If we treat the target as our y, then the norm will require some attention to ensure the difference in the orientation parts are handled correctly -- alternatively this could be bundled into the definition of f and y could be set to 0.

Another practical issue to overcome is the intended use and outputs. Where one problem for our existing applications requires a single data file with many corroborating data points to which we wish to aquire an estimate of some hidden property in noisy data. Inverse kinematics seeks to solve many positions for a single robot, by definition these will all have different solutions but would be of little use as individual results as the aim is to find a solver that is robust across the whole range of the robot -- perhaps it's best to consider a single problem as a full set of targets for a robot in the outputs from FitBenchmarking.

A demo of the inverse kinematics problem from SScanSS can be found here: https://bitbucket.org/SteveNneji/demo-for-fitbenchmarking/src/master/main.py

AndrewLister-STFC:
## Description of work

Adds a parser for inverse kinematics problems.

Fixes #1040

@RabiyaF @StephenNneji can you have a look at this and see if it's what you'd expect? I think this is the best way to get it to fit into the fitbenchmarking framework.

The main downside is that there's not much to show in the individual fitting plots. I could split it up so it showed aspects of the error (e.g. dx, dy, dz, angle in each axis) although I don't know if that's any more useful!
I think it also makes sense to have different targets for the robots as different fits in the table and to rely on the summaries. This is something that we could look at doing via the parser (single input file generates multiple versions of the fitting problem) if they all go from a single starting position.
It might get a bit more tricky if you wanted to run a chain of movements on a single robot. 

For the first draft, I've just copied the files you provided (with some slight edits to get them to pass on our different linting guidelines). Is there a package that this would normally be imported from?

## Testing Instructions

1.
2.
3.

## Checklist

### DIFF
--- a/fitbenchmarking/parsing/fitbenchmark_parser.py
+++ b/fitbenchmarking/parsing/fitbenchmark_parser.py
@@ -4,6 +4,7 @@ This file implements a parser for the Fitbenchmark data format.
 import os
 import re
 import typing
+
 import numpy as np
 
 from fitbenchmarking.parsing.base_parser import Parser
@@ -19,6 +20,7 @@ class FitbenchmarkParser(Parser):
     Parser for the native FitBenchmarking problem definition (FitBenchmark)
     file.
     """
+    PARAM_IGNORE_LIST = []
 
     def __init__(self, filename, options):
         super().__init__(filename, options)
@@ -145,13 +147,10 @@ class FitbenchmarkParser(Parser):
         :return: The starting values for the problem.
         :rtype: list
         """
-        # SasView functions can have reserved keywords so ignore these
-        ignore = ['name']
-
         starting_values = [
             {name: val
              for name, val in self._parsed_func[0].items()
-             if name not in ignore}]
+             if name not in self.PARAM_IGNORE_LIST}]
 
         return starting_values
 
@@ -174,7 +173,6 @@ class FitbenchmarkParser(Parser):
             self.fitting_problem.end_x = fit_ranges[0]['x'][1]
 
     def _set_additional_info(self) -> None:
-        # pylint: disable=no-self-use
         """
         Sets any additional info for a fitting problem.
         """
@@ -329,7 +327,7 @@ class FitbenchmarkParser(Parser):
                 count -= 1
             if count == 0:
                 value = string[:i]
-                rem = string[i+1:].strip(',')
+                rem = string[i + 1:].strip(',')
                 break
         else:
             raise ParsingError('Not all brackets are closed in function.')
@@ -376,7 +374,7 @@ class FitbenchmarkParser(Parser):
         :rtype: dict[str, np.ndarray]
         """
 
-        with open(data_file_path, 'r') as f:
+        with open(data_file_path, 'r', encoding='utf-8') as f:
             data_text = f.readlines()
 
         first_row = _find_first_line(data_text)
--- a/fitbenchmarking/parsing/sscanss_parser.py
+++ b/fitbenchmarking/parsing/sscanss_parser.py
@@ -1,0 +1,95 @@
+"""
+This file implements a parser for SScanSS problem sets.
+"""
+import math
+import os
+import sys
+import typing
+from importlib import import_module
+
+import numpy as np
+
+from fitbenchmarking.parsing.fitbenchmark_parser import FitbenchmarkParser
+
+
+class SScanSSParser(FitbenchmarkParser):
+    """
+    Parser for a SScanSS problem definition file.
+
+    Function calculates the error from a current pose to the target pose.
+    """
+    PARAM_IGNORE_LIST = ['robot', 'module', 'target']
+
+    def _create_function(self) -> typing.Callable:
+        """
+        Process the forward kinematicsfor the given robot into a callable.
+
+        Expected function format:
+        function='module=robots,robot=model1,target=target1,p1=...,p2=...,etc'
+
+        :return: A callable function
+        :rtype: callable
+        """
+
+        pf = self._parsed_func[0]
+        path = os.path.join(os.path.dirname(self._filename), pf['module'])
+        sys.path.append(os.path.dirname(path))
+        module = import_module(os.path.basename(path))
+        robot = getattr(module, pf['robot'])
+        target = getattr(module, pf['target'])
+
+        return self.inverse_kinematics_error(robot, target)
+
+    @staticmethod
+    def inverse_kinematics_error(robot, target):
+        """
+        Create a function for calculation the error in a forward kinematics
+        transform.
+        """
+        def error_func(*params):
+            cur_pose = robot.fkine(params)
+            pos_diff = cur_pose[:3, 3] - target[:3, 3]
+            angle_diff = SScanSSParser.emap(
+                target[:3, :3] @ cur_pose[:3, :3].transpose())
+            error = np.sum(pos_diff**2 + angle_diff**2)
+            return np.array([error])
+
+        return error_func
+
+    @staticmethod
+    def emap(r):
+        """
+        Convert a rotation matrix to an exponential map
+
+        :param r: the rotation matrix to convert
+        :type r: 3x3 matrix
+        """
+        b = r - np.identity(3)
+        _, _, v = np.linalg.svd(b)
+        axis = v[-1, :]
+
+        twocostheta = np.trace(r) - 1
+        twosinthetav = [r[2, 1] - r[1, 2],
+                        r[0, 2] - r[2, 0], r[1, 0] - r[0, 1]]
+        twosintheta = np.dot(axis, twosinthetav)
+        angle = math.degrees(math.atan2(twosintheta, twocostheta))
+        return angle * axis
+
+    def _get_equation(self) -> str:
+        """
+        Returns the equation in the problem definition file.
+
+        :return: The equation in the problem definition file.
+        :rtype: str
+        """
+        pf = self._parsed_func[0]
+        return f'{pf["robot"]}_{pf["target"]}'
+
+    def _get_data_file(self):
+        return ["no_file_required"]
+
+    @staticmethod
+    def _get_data_points(_filename):
+        # pylint: disable=unused-argument
+        return {'x': np.array([1]),
+                'y': np.array([0])}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,7 +110,7 @@ source = "versioningit"
 
 [tool.pylint.main]
 extension-pkg-allow-list = ["gofit"]
-ignore = ["benchmark_problems", "fonts", "test_files", "templates"]
+ignore = ["benchmark_problems", "examples", "fonts", "test_files", "templates"]
 ignored-modules = [
     "levmar",
     "lmfit",
@@ -147,3 +147,9 @@ ignored-argument-names = "arg|args|kwargs"
 
 [tool.versioningit]
 default-version = '0.0.0dev1'
+
+[tool.isort]
+skip = "benchmark_problems"
+
+[tool.autopep8]
+ignore = ['E401', 'E402'] # handled by isort
--- a/examples/benchmark_problems/InverseKinematics/data/kinematics.py
+++ b/examples/benchmark_problems/InverseKinematics/data/kinematics.py
@@ -1,0 +1,30 @@
+"""
+Definitions for the inverse kinematics problem files
+"""
+import sys
+from pathlib import Path
+
+import numpy as np
+
+# pylint: disable=wrong-import-position
+sys.path.append(str(Path(__file__).parent.parent / 'utils'))
+from robotics import enginx_combo, goniometer, table, two_link  # noqa:E402
+
+twolink_robot = two_link()
+table_robot = table()
+goniometer_robot = goniometer()
+enginx_combo_robot = enginx_combo()
+
+tl1 = twolink_robot.fkine([np.pi / 2, -np.pi / 2])
+tl2 = twolink_robot.fkine([1, 0.001])
+
+t1 = table_robot.fkine([109.23, np.pi / 2, -100, 40])
+t2 = table_robot.fkine([53.81, -np.pi / 4, -10, 40.9])
+
+g1 = goniometer_robot.fkine([np.pi / 2, -np.pi / 3, -np.pi / 4])
+g2 = goniometer_robot.fkine([np.pi / 3, np.pi / 4, np.pi / 3])
+
+ec1 = enginx_combo_robot.fkine(
+    [24, np.pi / 5, -103, -125, -np.pi / 2, -np.pi / 2, np.pi / 1.1])
+ec2 = enginx_combo_robot.fkine(
+    [12, -np.pi / 5, -113, -15, np.pi / 1.5, -np.pi / 20, np.pi / 4])
--- a/examples/benchmark_problems/InverseKinematics/enginx_combo.txt
+++ b/examples/benchmark_problems/InverseKinematics/enginx_combo.txt
@@ -1,0 +1,5 @@
+# Fitbenchmark Problem
+software = 'sscanss'
+name = 'Enginx Combo - target 1'
+description = 'Testing the enginx combo on target pose 1'
+function = 'module=data/kinematics,robot=enginx_combo_robot,target=ec1,p1=0,p2=0,p3=0,p4=0,p5=0,p6=0,p7=0'
--- a/examples/benchmark_problems/InverseKinematics/enginx_combo2.txt
+++ b/examples/benchmark_problems/InverseKinematics/enginx_combo2.txt
@@ -1,0 +1,5 @@
+# Fitbenchmark Problem
+software = 'sscanss'
+name = 'Enginx Combo - target 2'
+description = 'Testing the enginx combo on target pose 2'
+function = 'module=data/kinematics,robot=enginx_combo_robot,target=ec2,p1=0,p2=0,p3=0,p4=0,p5=0,p6=0,p7=0'
--- a/examples/benchmark_problems/InverseKinematics/goniometer.txt
+++ b/examples/benchmark_problems/InverseKinematics/goniometer.txt
@@ -1,0 +1,5 @@
+# Fitbenchmark Problem
+software = 'sscanss'
+name = 'Goniometer - target 1'
+description = 'Testing the goniometer on target pose 1'
+function = 'module=data/kinematics,robot=goniometer_robot,target=g1,p1=0,p2=0,p3=0'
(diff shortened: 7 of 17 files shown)


StephenNneji:
@AndrewLister-STFC Thanks for your work on the draft, I had a look, and have a few comments

1. The accuracy result values from running the example are really large. I am not sure if this is caused by the absence of bounds or if something else is going on. 
Screenshot 2024-04-15 130548

2. The matrix, vector etc files are present in [sscanss.core](https://github.com/ISISNeutronMuon/SScanSS-2/tree/master/sscanss/core), the problem is sscanss is not pip installable since our users typically install via exe. I will look into adding a setup.py or something

3. It would be nice to be able to run a list of targets for a single robot instead of creating a problem file ('Enginx Combo - target 1', 'Enginx Combo - target 2') for each target.

4. I think a plot showing the convergence trend i.e. best fit for each iteration, would be more useful than the current one

AndrewLister-STFC:
> @AndrewLister-STFC Thanks for your work on the draft, I had a look, and have a few comments
> 
> 1. The accuracy result values from running the example are really large. I am not sure if this is caused by the absence of bounds or if something else is going on.
> 2. The matrix, vector etc files are present in [sscanss.core](https://github.com/ISISNeutronMuon/SScanSS-2/tree/master/sscanss/core?rgh-link-date=2024-04-15T13%3A07%3A59Z), the problem is sscanss is not pip installable since our users typically install via exe. I will look into adding a setup.py or something
> 3. It would be nice to be able to run a list of targets for a single robot instead of creating a problem file ('Enginx Combo - target 1', 'Enginx Combo - target 2') for each target.
> 4. I think a plot showing the convergence trend i.e. best fit for each iteration, would be more useful than the current one

Thanks for the comments:

1. I've just improved the results (I messed up the func eval call) and improved the plots (a bit) - now the points from left to right are `x,y,z,*emap`
image
image


2. As long as it is installable and we can get it via an import, that's fine. Since the bit you shared was pure python, I just assumed it would be on Pypi.

3. 👍 

4. For these problems I agree (and in general it's a good feature request) but I think it might be something to discuss as a separate feature. We'll talk about it in the weekly meeting tomorrow.

Another thought (now that the fits are better) these systems are exactly solvable, are they always likely to be? If yes, this will need some additional development around our handling of an exact fit!

StephenNneji:
_Another thought (now that the fits are better) these systems are exactly solvable, are they always likely to be?_ 

No, real world problems will have some unreachable or unsolvable poses

fitbenchmarking/parsing/sscanss_parser.py
```diff
@@ -0,0 +1,131 @@
...
+        Create a function for calculation the error in a forward kinematics
+        transform.
+        """
+        def error_func(x, *params):
+            # pylint: disable=unused-argument
+            cur_pose = robot.fkine(params)
+            pos_diff = cur_pose[:3, 3] - target[:3, 3]
+            angle_diff = SScanSSParser.emap(
+                target[:3, :3] @ cur_pose[:3, :3].transpose())
+            return np.array(list(pos_diff) + list(angle_diff))
```
RabiyaF, line 88:
```
    current_pose = ri.fkine(q)
    error = np.zeros(6)
    error[:3] = target_pose[:3, 3] - current_pose[:3, 3]
    
    diff = target_pose[:3, :3] @ current_pose[:3, :3].transpose()
    error[3:6] = matrix_to_emap(diff)

    err = np.dot(error, error)
    return err
```

This is the snippet from the original [code](https://bitbucket.org/SteveNneji/demo-for-fitbenchmarking/src/master/main.py). Is the dot product computed outside the function ? 

I ran it once just to check the output (Like Stephen did earlier).

image

Some errors are too small and some are large. Some problems are exactly solvable (Two link - Target 1 and 2). But the error computed is not the same for exactly solvable problems. Would this be expected ?

fitbenchmarking/parsing/sscanss_parser.py
```diff
@@ -0,0 +1,131 @@
...
+    def _get_data_file(self):
+        return ["no_file_required"]
+
+    @staticmethod
+    def _get_data_points(_filename):
+        # pylint: disable=unused-argument
+        # x data is the spatial and agular coordinates for the pose
+        # y data is the error in each coord at the fit (as defined in
+        #     inverse_kinematics_error)
+        return {'x': np.array(["x", "y", "z", "alpha", "beta", "gamma"]),
```
RabiyaF, line 130:
```suggestion
        return {'x': np.array(["x", "y", "z", "ex", "ey", "ez"]),
```
Stephen suggested that these might be more appropriate names

fitbenchmarking/results_processing/plots.py
```diff
@@ -173,8 +175,10 @@ def plotly_fit(self, df):
                         error_y=self._error_dict,
                         mode='markers',
                         name='Data',
-                        marker=self._data_marker))
-                fig.update_layout(legend=self._legend_options)
+                        marker=self.data_marker))
+                fig.update_layout(legend=self.legend_options)
```
RabiyaF, line 179:
```suggestion
                fig.update_layout(legend=self._legend_options)
```

fitbenchmarking/results_processing/plots.py
```diff
@@ -97,8 +97,10 @@ def plot_initial_guess(self, df):
             error_y=self._error_dict,
             mode='markers',
             name='Data',
-            marker=self._data_marker))
-        fig.update_layout(legend=self._legend_options)
+            marker=self.data_marker))
+        fig.update_layout(legend=self.legend_options)
```
RabiyaF, line 101:
```suggestion
        fig.update_layout(legend=self._legend_options)
```

RabiyaF (commented):
I have had a look at the parser. The errors are still high in some cases. I also saw different error for perfectly solvable problems. I am not sure if that should be the expected behaviour.

(A small issue) The parameter values are always zero. It does not show what the starting values are in target 1 and 2.
image

fitbenchmarking/parsing/sscanss_parser.py
AndrewLister-STFC: I think all of the problems are exactly solvable, in the global sense. It could be that there are local minima that some solvers get stuck in.

The dot product is done as part of the cost function:
$$c = \Vert f(x; p) - y \Vert^2$$

For this problem, the error $f(x; p) - y$ is defined by the `error_func`. Since it's in a norm, the outcome is the same.

For some solver this makes a big difference as they can get a gradient for each of the error terms.

I'll rerun it with the solvers listed. It could be that bounds would help as you mention and they should probably all be bounded (or at least the linear actuators should be!). Running with `scipy_ls` gives good results which is what I was testing with. From an initial observation, `scipy` takes much longer than `scipy_ls`.

AndrewLister-STFC: Looking at the graphs from the `scipy` runs, the accuracy seems to be way off for some of them. I'll have a look for why.

AndrewLister-STFC: @RabiyaF can you try again with an unweighted cost function. Since the weighting is being set to $\max(\sqrt{y},10^-8)$ and $y=0$, it's just scaling up the chisq for no reason. I think I'll have to explicitly set $e=0$.

RabiyaF: I ran it with `scipy_ls` and `nlls`. In this case, you are right, the errors are reasonable. Maybe `scipy: Nelder-Mead` is just not suitable for this problem. SScanSS uses the `nlopt` package. I would just like to verify that nlopt minimisers are giving good results. I will try to do this soon if I can. But I was having some trouble with installing `nlopt` on a Mac. But I might look into it later this week.  

image

AndrewLister-STFC: There're quite a few minimizers for nlopt. Here's what I get on the first few:
image

AndrewLister-STFC:
@AndrewLister-STFC  to remove kinematics files and install SScanSS-2 https://github.com/ISISNeutronMuon/SScanSS-2

AndrewLister-STFC:
@RabiyaF there's a bit of an issue in updating the docker which is that I need to push the changes so that I can have tests that work on this PR. Can you review the docker changes and let me know if you're happy for me to push them (so that they are ready when this PR runs its tests!).

RabiyaF:
At first glance, the steps to install SScanSS-2 looks good.

I think these changes should work unless there are some incompatibility issues with python 3.12 with the full install softwares. 

I just had a question about the docker files (because I haven't worked on this part of the code yet). The **BasicInstall** Docker file is `fitbenchmarking/fitbenchmarking-deps:latest` which you then use as the base image to build `FullInstall` Docker Image ? And the FullInstall is the `fitbenchmarking/fitbenchmarking-extras:latest` in the GitHub runners ?

If that is the case, then I think down the line the **BasicInstall** can be removed and those steps can be done directly on the runners. No?  

I also noticed that the mantid version mentioned in the docker file is 6.2 and I think the latest is 6.10. Maybe all this can be looked at later in a different PR.

AndrewLister-STFC:
> At first glance, the steps to install SScanSS-2 looks good.
👍 
 
> I think these changes should work unless there are some incompatibility issues with python 3.12 with the full install softwares.
Good spot, I'll revert it if there're too many complaints and raise an issue with the output when building.

> I just had a question about the docker files (because I haven't worked on this part of the code yet). The **BasicInstall** Docker file is `fitbenchmarking/fitbenchmarking-deps:latest` which you then use as the base image to build `FullInstall` Docker Image ? And the FullInstall is the `fitbenchmarking/fitbenchmarking-extras:latest` in the GitHub runners ?
Yes. I think it's changed a bit recently which has simplified it (thanks to your work on the tests!). It was meant to be 1 docker container that was set up to install the pip installables on, and one with the non-trivial installs on, so that we don't have to build everything everytime we run tests.

> If that is the case, then I think down the line the **BasicInstall** can be removed and those steps can be done directly on the runners. No?
Agreed, the basic one is now redundant and could be removed (maybe we should bring it up on tuesday).

> I also noticed that the mantid version mentioned in the docker file is 6.2 and I think the latest is 6.10. Maybe all this can be looked at later in a different PR.
Good spot. Mantid is often out of date in our tests as it's so annoying to install unless you use their conda distribution... I'll look at updating to the latest

AndrewLister-STFC:
Just putting some statuses in, this is blocked awaiting #1312

pyproject.toml
```diff
@@ -143,3 +143,9 @@ quote-style = "double"
 
 [tool.versioningit]
 default-version = '0.0.0dev1'
+
+[tool.isort]
+skip = "benchmark_problems"
+
+[tool.autopep8]
+ignore = ['E401', 'E402'] # handled by isort
```
AndrewLister-STFC, line 151:
```suggestion
```

AndrewLister-STFC:
@RabiyaF Do you know if sscanss has any randomness? I'm not sure why this isn't deterministic...

RabiyaF:
Sorry, I missed your comment. I can have a closer look at the issue tomorrow and get back to you.

AndrewLister-STFC:
@RabiyaF I think this is ready for review  now (no rush). If you're happy with how his looks I can move the temporary changes in the ci to the container and we can finally get this merged

RabiyaF:
I can have a look at this on Monday :) I don't think I will have a lot to say regarding the code structure. When I had a cursory look at it this week, All looks good. I will just have a look at it again in detail next week and can let you know.

Regarding the container - I was having a look at removing the `numpy < 2.0` from the Container file this week. I can't do that just yet because SScanSS need `numpy < 2.0`

docs/source/users/install_instructions/externals.rst
```diff
@@ -109,6 +109,13 @@ RALFit is available to use as fitting software.
 
 Instructions on how to build the python interface are at `<https://ralfit.readthedocs.io/projects/Python/en/latest/install.html>`_
 
+SScanSS-2
```
RabiyaF, line 112: To test this PR, I cloned **SScanSS-2** and then ran `pip install .` in that **SScanSS-2** directory. I also ran `ruff check .` in the fitbennchmarking folder. I saw a lot of errors because sscanss is formatted differently. 
image

Maybe something like this needs to be added to pyproject.toml.

image

fitbenchmarking/parsing/tests/test_parsers.py
```diff
@@ -517,8 +524,8 @@ def test_parse_problem_file(self):
         """
         Tests the parse_problem_file method
         """
-        filename = Path(__file__).parent / "nist" / "basic.dat"
-        fitting_problem = parse_problem_file(filename, OPTIONS)
+        filename = os.path.join(os.path.dirname(__file__), "nist", "basic.dat")
+        fitting_problem = parse_problem_file(filename, OPTIONS)[0]
```
RabiyaF, line 528:
```suggestion
        filename = Path(__file__).parent / "nist" / "basic.dat"
        fitting_problem = parse_problem_file(filename, OPTIONS)[0]
```

Again I prefer slowly changing the code to use `Path` over `os`.

fitbenchmarking/test_files/all_parsers_set/Models/kinematics.py
```diff
@@ -0,0 +1,213 @@
+"""
```
RabiyaF, line 1: This file is the duplicate of the file in the examples folder. Will it be removed later ?

RabiyaF (changes requested):
Just wanted to suggest a few changes. They are mainly in the formatting of the code. There are a few pylint references in the code base that can be removed now that we are using ruff. I also suggested some changes that use `Path` instead of `os`.

fitbenchmarking/controllers/mantid_controller.py
```diff
@@ -111,6 +111,8 @@ def __init__(self, cost_func):
 
         if self.problem.multifit:
             # Multi Fit
+            if self.data_x[0].dtype not in [float, int]:
+                self.data_x = [np.array(range(len(d))) for d in self.data_x]
```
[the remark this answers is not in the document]
AndrewLister-STFC: The new sscanss parser puts strings in (It makes the graphs make much more sense)

docs/source/users/install_instructions/externals.rst
AndrewLister-STFC: Did you clone SScanSS-2 into the fitbenchmarking directory? I get this in the src dir:
```
(fitbenchmarking) alister@SCLT344:~/fitbenchmarking$ cd fitbenchmarking
(fitbenchmarking) alister@SCLT344:~/fitbenchmarking/fitbenchmarking$ ruff check .
All checks passed!
```

fitbenchmarking/test_files/all_parsers_set/Models/kinematics.py
AndrewLister-STFC: I think this is needed for systests - all of the systests duplicate from the examples but they're only a subset of the available ones

fitbenchmarking/parsing/sscanss_parser.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def parse(self) -> "list[FittingProblem]":
+        template = super().parse()
+
+        out = []
+
+        pf = self._parsed_func[0]
+        path = os.path.join(os.path.dirname(self._filename), pf["module"])
+        sys.path.append(os.path.dirname(path))
+        module = import_module(os.path.basename(path))
```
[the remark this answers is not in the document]
AndrewLister-STFC: Agreed, pathlib is nicer

fitbenchmarking/parsing/fitbenchmark_parser.py
```diff
@@ -109,6 +111,8 @@ def _is_multifit(self) -> bool:
         :return: True if the problem is a multi fit problem.
         :rtype: bool
         """
+        if "input_file" not in self._entries:
+            return False
```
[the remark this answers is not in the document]
AndrewLister-STFC: Yes, I think this is a condition that used to be in place but got broken while refactoring. I'll update the tests

fitbenchmarking/parsing/tests/test_parsers.py
AndrewLister-STFC: Oops, I think I missed the Path change when rebasing on this one! Good spot

fitbenchmarking/controllers/mantid_controller.py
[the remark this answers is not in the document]
RabiyaF: Ohh are you saying in the case where sscanss problems are run with mantid minimizers ? I did not test that. I was asking because I have to refactor the mantid controller next and add the tests. If that is the case then maybe this condition is not required because sscanss problems won't be multifit. And it is okay to have the next condition ?

docs/source/users/install_instructions/externals.rst
RabiyaF: I usually run the ruff check in the main fitbenchmarking directory. I don't cd into fitbenchmarking. Because it also checks conftest.py is in the main directory and the docs folder. But this can work as well. We can cd into fitbenchmarking and run the ruff checks and if there are any issues with the other files it will be caught by the pre commit bot.

AndrewLister-STFC: I get the same thing in the root of the repo:
```
(fitbenchmarking) alister@SCLT344:~/fitbenchmarking$ pwd
/home/alister/fitbenchmarking
(fitbenchmarking) alister@SCLT344:~/fitbenchmarking$ ruff check .
All checks passed!
```

From your screenshot it looks like SScanSS-2 is in the fitbenchmarking root folder, but since it's not in the repo I don't think we need to add it to the pyproject.toml

RabiyaF (dismissed):
I am approving the PR so that it can be tested with the updates to the container. But it needs a rebase and it would be best to check that the coverage for the fitbench_parser and mantid_controller is 100% before merging.

jess-farmer:
@AndrewLister-STFC when you have a chance, please could you check the status of the unresolved comments on this PR, and also remove the sscanss parser from the coverage report?

PR #1281, opened 2024-05-13
Added new controller for Scipy leastsq

### LINKED ISSUE #1225: Add scipy's `leastsq` to FitBenchmarking
[`scipy`'s legacy `leastsq` routine](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.leastsq.html) is a wrapper around `MINPACK`, and this is what is called in BUMPS.

However, there are some differences between the wrappers, in particular in how the Jacobians are handled.

We should add a new controller to give FitBenchmarking access to this routine.

This controller should allow derivatives to be held by either columns or rows (see `col_deriv` parameter).

RabiyaF:
## Description of work

Fixes #1225

Adds a controller for [scipy.optimize.leastsq](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.leastsq.html). Also refactored the `controller` and `minimizer` tests.

## Testing Instructions

1. Test the new controller using `fitbenchmarking -s scipy_leastsq`
2. Run the test using `pytest fitbenchmarking/controllers/tests/test_controllers.py --test-type default`

## Checklist


jess-farmer (commented):
Thanks for adding in the new controller and making the changes to `test_controllers`! I just have a couple of points/questions:

Is there a reason for the different minimizer options for `scipy_leastsq`? From the [docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.leastsq.html) it looks like there's no option to use different methods, and the different minimizer options aren't referred to within the controller.

It would be good to add relevant documentation to the Minimizer Options page, as well as add the software to the regression tests - see point 4 [here](https://fitbenchmarking.readthedocs.io/en/latest/extending/controllers.html)

jess-farmer (approved):
Thanks for making these changes Rabiya - all looks good to me!

PR #1283, opened 2024-05-14
Fixes bug that causes results output crash

RabiyaF:
Fixes the crash caused when Accuracy is `Infinite` and runtime is `finite`. Although, the reports will be generated, they will have cell styling issues. These will be fixed later.

Screenshot 2024-05-14 at 10 11 47 AM

### DIFF
--- a/fitbenchmarking/results_processing/fitting_report.py
+++ b/fitbenchmarking/results_processing/fitting_report.py
@@ -28,7 +28,7 @@ def create(results, support_pages_dir, options):
     """
 
     for prob_result in results:
-        if np.isinf(prob_result.accuracy):
+        if np.isinf(prob_result.accuracy) and np.isinf(prob_result.runtime):
             continue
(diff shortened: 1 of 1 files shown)


AndrewLister-STFC (dismissed):

RabiyaF:
Here is an another example. The screenshots are from a mantid analysis using a single dataset with variable starting parameters. `Gofit multistart` consistently gives the best results. But if I looked at the `local_min` table it seems to suggest that it was all errors because the error flag is non zero and parameter bounds were not respected.

## Before
Screenshot 2024-05-14 at 3 52 15 PM

## After
Screenshot 2024-05-14 at 3 51 03 PM

AndrewLister-STFC (approved):
Changes all look good, thanks for the quick fix! The tests pass althouh it failed to submit to coveralls 🙃 I think we just merge anyway 🚀

PR #1287, opened 2024-05-20
Fixed emissions display when not measured

### LINKED ISSUE #1221: Emissions listed as inf when not activated
We should list an NA in the fitting report or remove the row, rather than inf.

jess-farmer:
## Description of work

Fixes #1221 

Changes fitting report so that Emissions value is listed as 'N/A' when the emissions table option is not selected. I have also updated the `options_template.ini` so that the emissions table is not deactivated, as measuring emissions is no longer painfully slow.

## Testing Instructions

1. Run Fitbenchmarking without emissions table option, and check fitting report.
2. 
3.

## Checklist

### DIFF
--- a/examples/options_template.ini
+++ b/examples/options_template.ini
@@ -26,11 +26,6 @@ make_plots: no
 
 results_dir: new_results/
 
-table_type: acc
-            runtime
-            compare
-            local_min
-
 [LOGGING]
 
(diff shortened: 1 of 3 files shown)


RabiyaF (approved):
I tested it with and without the emissions tables and it works as expected 👍🏼

PR #1297, opened 2024-07-23
Added cost function type to logs

### LINKED ISSUE #1293: Cost function type not printed in logs
Currently when running FitBenchmarking with more than one `cost_func_type` it isn't clear in the logs that multiple cost functions are being run (in the example below both `nlls` and `weighted_nlls` cost functions are selected, but from the output shown it just looks like everything has run twice for no reason) - an extra line should be printed which states the cost function being used.

image

jess-farmer:
## Description of work

Fixes #1293 

## Testing Instructions

1.
2.
3.

## Checklist

### DIFF
--- a/fitbenchmarking/core/fitting_benchmarking.py
+++ b/fitbenchmarking/core/fitting_benchmarking.py
@@ -8,9 +8,9 @@ fitting software.
 
 import os
 import platform
+import time
 import timeit
 import warnings
-import time
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


RabiyaF (changes requested):
Thank you for adding the logs for the cost function and the minimizer. I just wanted to suggest a minor change.

RabiyaF (approved):
Thank you for fixing this logging issue :smile: 👍🏼

PR #1300, opened 2024-07-29
Fixed output formatting errors

### LINKED ISSUE #1195: Output formatting errors
When I run FitBenchmarking, I get a lot of max time ratio warnings.  the formating seems all off for these:

image

- The progress bar is copied to the screen
- The line of code raising the warning is printed
- there's a stray `warnings.warn(` at the end of the message

Can we fix this so it just outputs: 

> UserWarning: The ratio of the max time to the min is 4.763145328754864, which is larger than the tolerance of 4. The min time is 0.0009025829999700363. This can indicate that the fitting engine is caching results. If the min time is small this may just indicate that other non-FitBenchmarking CPU activities are taking place that affects the timing results

Possibly another issue, but these are now far too frequent (and not very helpful -- the times are so tiny that they tell you nothing, but this could be offputting to a user)

jess-farmer:
## Description of work

Fixes #1195 

Switches out `warnings.warn` for `LOGGER.warning` to avoid the progress bar being printed and other formatting issues.

## Testing Instructions

1.
2.
3.

## Checklist

### DIFF
--- a/fitbenchmarking/core/tests/test_fitting_benchmarking.py
+++ b/fitbenchmarking/core/tests/test_fitting_benchmarking.py
@@ -277,9 +277,10 @@ class PerformFitTests(unittest.TestCase):
                       checkpointer=self.cp)
 
             _ = fit._Fit__perform_fit(controller)
-            self.assertTrue("ratio of the max time to the min is 5.0"
-                            in ''.join([log.output[i]
-                                        for i in range(0, len(log.output))]))
+            self.assertTrue(("The ratio of the max time to the min is "
+                             "5.00000000, which is larger than the tolerance "
+                             "of 4. The min time is 1.00000000.")
+                            in log.output[0])
 
     @patch("fitbenchmarking.controllers." +
            "base_controller.Controller.validate")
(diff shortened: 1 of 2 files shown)


fitbenchmarking/core/tests/test_fitting_benchmarking.py
```diff
@@ -276,8 +277,9 @@ def test_perform_fit_runtime_warning(self, mock1, mock2, mock3):
                       checkpointer=self.cp)
 
             _ = fit._Fit__perform_fit(controller)
-            assert ("ratio of the max time to the min is 5.0"
-                    in ''.join([str(w[i].message) for i in range(0, len(w))]))
+            self.assertTrue("ratio of the max time to the min is 5.0"
+                            in ''.join([log.output[i]
+                                        for i in range(0, len(log.output))]))
```
RabiyaF, line 282:
```suggestion
            self.assertTrue(("The ratio of the max time to the min is "
                             "5.00000000, which is larger than the tolerance "
                             "of 4. The min time is 1.00000000.")
                            in log.output[0])
```
You can check the whole statement to verify that the ratio and min time are printing to the right number of significant figures. I also removed the join statement because the information is in the zeroth index. The redundant join step can be removed without effecting the test.

RabiyaF (changes requested):
I tested it. The code looks good and works as expected 👍🏼 I had one minor change to suggest for the test.

RabiyaF (approved):

PR #1307, opened 2024-08-19
Added number of iterations/function evaluations to fitting report

### LINKED ISSUE #1144: Broken link in Minuit docs
The [Minuit 2](https://root.cern.ch/doc/master/Minuit2Page.html) link in the [minuit section](https://fitbenchmarking.readthedocs.io/en/latest/users/options/minimizer_option.html#minuit-minuit) of the docs is broken, so this link should be removed/replaced.

### LINKED ISSUE #1222: Iteration count on the fitting report
It might be useful for debugging if we could output an iteration count and add it to the fitting report.

### LINKED ISSUE #1339: Fix broken minuit link in the documentation
Screenshot 2024-10-09 at 12 35 28

Possible alternatives are:

- https://root.cern.ch/download/minuit.pdf
- https://en.wikipedia.org/wiki/MINUIT
- https://root.cern.ch/root/htmldoc/guides/minuit2/Minuit2.html
- https://scikit-hep.org/iminuit/

jess-farmer:
## Description of work

Fixes #1222 
Fixes #1339 
Fixes #1144 

Adds either iteration count or number of function evaluations to the fitting report if this information is available

## Testing Instructions

1.
2.
3.

## Checklist


jess-farmer:
I've added either an iteration count or number of function evaluations to the fitting report for all minimizers that provide this information apart from the MATLAB minimizers - I've tried installing MATLAB to run these however am getting the following error message:
`matlab.engine.MatlabExecutionError: Python Error: AttributeError: module 'sip' has no attribute 'setapi'`

jess-farmer:
> I've added either an iteration count or number of function evaluations to the fitting report for all minimizers that provide this information apart from the MATLAB minimizers - I've tried installing MATLAB to run these however am getting the following error message: `matlab.engine.MatlabExecutionError: Python Error: AttributeError: module 'sip' has no attribute 'setapi'`

Tried this again in a fresh venv and all worked okay so this must have been a problem on my end

fitbenchmarking/controllers/scipy_go_controller.py
```diff
@@ -100,4 +99,6 @@ def cleanup(self):
         else:
             self.flag = 2
 
-        self.final_params = self._popt
+        self.final_params = self._result.x
+        self.iteration_count = self._result.nit
+        self.func_evals = self._result.nfev
```
RabiyaF, line 104: I noticed `self.iteration_count` and `self.func_evals` are sometimes set in the **cleanup** method and sometimes in the **fit** method. I personally like it when it is done in the **cleanup** method. But I know for the Matlab stuff it is easier to do it in the **fit** method because the results are not saved in a class attribute. Do you think it would be better doing it consistently so it is much clearer when a new controller is added ?

fitbenchmarking/controllers/scipy_go_controller.py
```diff
@@ -79,11 +79,10 @@ def fit(self):
         fun = self.cost_func.eval_cost
         bounds = self.value_ranges
         algorithm = getattr(optimize, self.minimizer)
-        result = algorithm(fun, bounds, **kwargs)
-        self._popt = result.x
-        if result.success:
+        self._result = algorithm(fun, bounds, **kwargs)
+        if self._result.success:
```
RabiyaF, line 83: There is a possibility to clean up the code a little bit here as well. Now that you have made `results` a class variable. You can do this is in **cleanup** method and remove `self._status` from the class.

```
if self._result.success:
      self.flag = 0
elif "Maximum number of iteration" in self._result.message:
      self.flag = 1
else:
       self.flag = 2
```

RabiyaF (changes requested):
Thank you Jess for adding this feature to `fitbenchmarking` 💯  This will be really useful in evaluating the minimizers down the line.

fitbenchmarking/controllers/scipy_go_controller.py
jess-farmer: Good point - will make this change!

fitbenchmarking/controllers/scipy_go_controller.py
jess-farmer: I did think this as I was working through the different controllers - where possible I have set `self.iteration_count`/`self.func_evals` in cleanup, but as you say for some controllers like the Matlab ones and RALFit this information is not saved in a class attribute so I thought it would be neater to set these within fit rather than create additional class attributes. Happy to change this though as I agree it would make it clearer for writing new controllers if these are always set within cleanup

RabiyaF: I will leave it up to you to decide how to do the refactoring part. Maybe there are more things that can be optimized and another issue can be created to look at the controllers. But it would be nice to define a standard whatever you feel is best. Because I just noticed it while reviewing but I am not sure how many things you might have to change.

I also noticed that a link was broken in the docs for minuit while reviewing the PR. Maybe it can be fixed here but it is not urgent. It can also be done in another PR.

jess-farmer: Sounds good - regarding the broken minuit link, I think I opened an issue for this last year so happy to fix it in this PR so that it is actually fixed and not forgotten about again!

RabiyaF (dismissed):
Thank you so much for adding this feature. It looks good to me. We can merge it and fix the tests later.

RabiyaF (approved):
Thank you once again Jess for adding this feature 😃 Everything look good. I am happy to merge it.

PR #1316, opened 2024-09-04
Updated the PyPI install check gitHub actions

### LINKED ISSUE #1314: Fix nightly tests
Currently the nightly test is failing due to missing `distutils` on the latest version of python.

We should: 
- fix this so that the tests pass
- add a nightly build badge to the readme so that it's clear when the nightly build fails
- add a simple regression test to the nightly test, just so that we can see that things are indeed working

RabiyaF:
## Description of work

Fixes #1314 

Added badge for pip install checks. These checks will now run weekly. They will check the install on **Linux**, **Windows** and **Mac**. The tests will verify `fitbenchmarking` is installable on the least supported python version (3.8.x).

The test for checking pip install on python 3.x can be added when the next version of fitbenchmarking is released that supports python 3.12.

At least, it will temporarily fix the nightly notifications for a while.

## Testing Instructions

1. I verified that the PyPI install checks are passing. Can be verified [here](https://github.com/fitbenchmarking/fitbenchmarking/actions/runs/10818404150/job/30013947475)

## Checklist


RabiyaF:
Yes, I just noticed that. I was trying to see why the main workflow is working with 3.12 and this is not. Tyrone raised it in the meeting. But I am also thinking maybe it might be worth considering reducing the frequency to every week instead of nightly.

AndrewLister-STFC (approved):
🚀

PR #1331, opened 2024-09-23
Updated workflows to stop scheduled tests running on forks

RabiyaF:
## Description of work

I work through a fork and I noticed that the scheduled weekly GitHub actions for `Tests` and `PYPI` are running on my fork as well. And here are the OS usage summaries for the latest runs. 

- `Tests`
      image
- `PYPI`
      image

The [PYPI](https://github.com/RabiyaF/fitbenchmarking/actions/runs/10991099166) install passes but I don't think these tests need to be run on forks as well.

The [Tests](https://github.com/RabiyaF/fitbenchmarking/actions/runs/10991041833) fail because of the **Publish Test Results** step. Again, I don't think these need to be duplicated on forks as long as there is activity on the main repo.

Turning off these unnecessary runs might save a small amount of ⚡ 

## Testing Instructions

1. Can only be tested after merging, the test will not run on my fork next week. But everything else should remain the same.

## Checklist

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   package_install_check:
     name: Package install check
+    if: github.event.repository.fork == false
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


jess-farmer (approved):
Thanks for making this change Rabiya - always good to reduce energy usage where we can!

PR #1336, opened 2024-10-02
Fix issue where failures were incorrectly reported in performance profile

### LINKED ISSUE #1240: Acc and runtime performance profiles show different n failures for same solver
When running the NIST/average_difficulty problems, one can see that in the performance profile plots, certain solvers are marked as having a certain number of failures based on the accuracy plot, but no failure based on the runtime plot. See 'dream [bumps]' and 'newton [bumps]' in the picture below, they are marked as having failed in the legend of the accuracy plot but not in the runtime plot.

The number of failures (to display in the plots' legends) is determined in function create_plot in performance_profiler. The upper bound to determine failures is set to 1.0e20, which might be causing the problem in these examples.

pp

AndrewLister-STFC:
## Description of work

Fixes #1240

Counts failures as an `np.inf` rather than `> 1e20`.

## Testing Instructions

1. Run the test example from the comment in the issue and verify no failures
2. Run it with gradient_free as well and verify all fail correctly (gradient free requires bounds)

## Checklist

- [ ] ~I have updated the documentation in the relevant places to cover the changes.~
- [x] I don't think this requires any doc changes

### DIFF
--- a/fitbenchmarking/results_processing/performance_profiler.py
+++ b/fitbenchmarking/results_processing/performance_profiler.py
@@ -277,13 +277,11 @@ def adjust_values_to_plot(step_values: 'list[np.ndarray]',
         label = f"{solver}"
         plot_points = np.linspace(0.0, 1.0, solver_values.size)
         plot_points = np.append(plot_points, 1.0)
-        inf_indices = np.where(solver_values > huge)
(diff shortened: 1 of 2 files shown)


jess-farmer (approved):
Looks good to me and tests run as expected!

PR #1271, opened 2024-04-11
Fixed corruption in checkpoint file after an exception

AndrewLister-STFC:
## Description of work

Fixes #1265 

Guards the cp.finalise() call with a try-finally block. This re raises any exceptions after the checkpoint file is finalised.

## Testing Instructions

1. raise an exception from within results_processing
2. run fitbenchmarking on your favourite test set
3. revert the change from (1.)
4. run `fitbenchmarking-cp report` or `fitbenchmarking --load-checkpoint`

## Checklist

- [x] The title is of a format appropriate for a line in future release notes.
- [x] I have added the appropriate tags to the PR.
- [x] My PR represents one logical piece of work.
- [x] My PR fully fixes the issue linked. If new issues have been created, what are they?
- [ ] I have added the appropriate tests to cover code that has been added.
- [x] I have updated the documentation in the relevant places to cover the changes.

PR #1337, opened 2024-10-02
Fix issue where perf profiles could be out of sync in dash

### LINKED ISSUE #1256: Colours in Dash plots in compare table disagree if either plot has already been modified
Once the Dash app is running, if one plays with the Dash plot in either accuracy or runtime table (i.e., removing and adding minimizers), and then switches to the compare table, the colors in two plots (accuracy and runtime) will no longer agree.  Specifically, the minimizers that were previously removed will be shown in different colours in the two plots.

AndrewLister-STFC:
## Description of work

Fixes #1256 

Copies the formatting from the first plot chosen when multiple are displayed via dash.

## Testing Instructions

1. Go to accuracy table page and click online plots (while dash is running)
2. Remove all lines and add them back in a random order so the colours are different
3. Check against the online plots on the runtime table page to make sure they disagree
4. Open the compare table and check that they update to match

## Checklist

### DIFF
--- a/fitbenchmarking/core/results_output.py
+++ b/fitbenchmarking/core/results_output.py
@@ -588,8 +588,12 @@ def check_max_solvers(opts, solvers, max_solvers):
     return opts
 
 
-def display_page(pathname, profile_instances_all_groups,
-                 layout, max_solvers):
+def display_page(
+    pathname: str,
+    profile_instances_all_groups: "dict[str, dict[str, DashPerfProfile]]",
+    layout: "list",
+    max_solvers: int
+):
     """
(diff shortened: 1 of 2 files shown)


AndrewLister-STFC:
Not sure why this fails but the windows tests on #1271 pass. As far as I can tell, the deps and versions are all the same...

RabiyaF:
Rerunning it will hopefully fix it. It happened on my PR as well. But I think it happens very rarely but I don't think is an issue with fitbenchmarking.

tyronerees (approved):
This works as expected for me 👍

PR #1338, opened 2024-10-03
Added error message for online plots that are unavailable

### LINKED ISSUE #1253: When using load_checkpoint on old results (previous to last run), Dash pp plots are not shown
When one uses load_checkpoint to visualise results that are older than the one from the last run, the Dash app fails to display the performance profiles. When one clicks the online button, a blank space is shown in place of the pp plot, and no warning is given. 

From the terminal it's possible to see that a Dash app is running, but we get an error in display_page (see traceback attached).
traceback

Checking the url for the broken pp plot, I saw that the Dash app available at that url was broken (it just appeared as a blank page), while the only Dash app still running was the one associated with the latest run, so at a different URL.

AndrewLister-STFC:
## Description of work

Fixes #1253 

Options now creates a uuid for every instance, and this is embedded in the dash urls.
If the uuid of the options file does not match the dash url, a message about using `--load-checkpoint` is displayed.

Existing plots will get an error saying the url is in the wrong format.

## Testing Instructions

1. Try to recreate the issue as explained in my comment

## Checklist

### DIFF
--- a/fitbenchmarking/core/results_output.py
+++ b/fitbenchmarking/core/results_output.py
@@ -589,7 +589,7 @@ def check_max_solvers(opts, solvers, max_solvers):
 
 
 def display_page(pathname, profile_instances_all_groups,
-                 layout, max_solvers):
+                 layout, max_solvers, run_id):
     """
     Update the layout of the dash app.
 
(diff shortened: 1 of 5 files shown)


fitbenchmarking/core/results_output.py
```diff
@@ -601,18 +601,25 @@ def display_page(pathname, profile_instances_all_groups,
...
     :return: The updated layout
     :rtype: html.Div
     """
 
     try:
-        _, group, plot, metric_str = pathname.split('/')
+        _, plot_id, group, plot, metric_str = pathname.split('/')
     except ValueError:
         return ("404 Page Error! Path does not have the expected shape. "
                 "Please provide it in the following form:  \n"
```
RabiyaF, line 615: Just had a one word change suggestion to the error msg.
```suggestion
        return ("404 Page Error! Path does not have the expected format. "
                "Please provide it in the following form:  \n"
```

RabiyaF (changes requested):
It all looks good 👍🏼 I tested it and it works as expected.

RabiyaF (approved):
Everything looks good. I will merge it as soon as the tests are done :)

PR #1354, opened 2024-10-30
Added integration workflow

### LINKED ISSUE #1051: Mac tests
We should test the default tests on Mac

### LINKED ISSUE #1292: Test against recent python versions
Currently we only test against python 3.7, we should update at least the basic tests so that more recent versions of python are tested

### LINKED ISSUE #1340: Add an integration workflow
Right now, [main.yml](https://github.com/fitbenchmarking/fitbenchmarking/blob/master/.github/workflows/main.yml) reruns the same workflow when a PR is merged into master. Instead we can add a different integration workflow that runs the default tests on `Windows` and `Mac` instead.

And we can remove the `Windows` tests from main.yml. So they don't run with every push to the PR. Again this might save a little bit of ⚡ because the `Windows` tests take longer to run than the `Linux` tests.

RabiyaF:
## Description of work

Fixes #1340 
Fixes #1051
Fixes #1292

The full tests take slightly longer now, therefore I have removed the `windows` tests from the main workflow. This means that they won't run with every commit push. This will reduce the runtime to the main workflow by ~15 minutes each time.

image

Instead, I have moved the `windows` and `mac` tests to an integration workflow. This will only run when a PR is approved. It will be less frequent and will save energy overall.

## Testing Instructions

1. The integration workflow should pass. This can be verified [here](https://github.com/fitbenchmarking/fitbenchmarking/actions/runs/11599330825)

## Checklist

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -6,7 +6,7 @@ on:
     types: [submitted]
 
 jobs:
-  default_unit_tests:
+  default_tests_win_mac:
     name: Default Unit Tests
     if: github.event.review.state == 'approved' && github.event.repository.fork == false
     strategy:
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,41 @@
+name: 'Integration'
+
+on:
+  push:
+    branches:
+      - 'master'
+      - 'release-**'
+  release:
+    types: [published]
+  workflow_dispatch:
+
+jobs:
+  default_unit_tests:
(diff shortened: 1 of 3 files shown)


RabiyaF:
I want to make a few changes to this because now that I have bumped the python version, the linux test for python 3.13 pass but ones for windows fail. The issue is with the GitHub runner. So I will get the test to run when the PR is approved.

tyronerees (approved):
This seems sensible to me -- thanks, Rabiya

PR #1365, opened 2024-11-06
Added python and numpy version tags to checkpoint.json

### LINKED ISSUE #1318: Add python and numpy version metadata tags
The checkpoint regression tests were failing on **python** `3.12.5` with **numpy** `2.1.0` and passing on **python** `3.8.10` with **numpy** `1.24.4`.

This was because of the parameters that are encoded and decoded using the following lines of code: 

- `a85encode(pickle.dumps(value)).decode('ascii')`
- `pickle.loads(a85decode(value.encode('ascii')))`

Those parameters are `params = ['fin_y', 'r_x', 'jac_x', 'params', 'ini_y', 'x', 'y', 'e', 'sorted_idx', 'ini_params']`.

For now, these parameters are not matched in the `CheckpointTests`.

This might possibly be due to a recent change in the numpy backend. I am looking at the changelogs for numpy which mention a change in precision. But I haven't been able to pinpoint the issue yet.

There is a recommended migration [guide](https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#numpy-2-migration-guide) for migrating to numpy 2.0. It can be done automatically using `ruff`.

However, It would be nice if the **python** and **numpy** version used to generate the checkpoint file is stored either in the checkpoint file itself or a config file as mentioned in #1267.

This will allow users to know if they are trying to decode/merge an incompatible checkpoint file.

RabiyaF:
## Description of work

Fixes #1318 

Added `python` and `numpy` tags to the fitbenchmarking report. Also updated the style of the report. In this PR, I am not trying to handle for the case when checkpoint files (generated on different numpy versions) are merged. This needs a little while to understand the issue and properly handle that case. In my experience, there are very few cases that will cause an issue. 

## Testing Instructions

1. Run fitbenchmarking a verify that python and numpy versions are visible.

image

2. Also verify that the **checkpoint.json** in `fitbenchmarking_results` contains the config.
image

3. Delete the config from the **checkpoint.json** in `fitbenchmarking_results`. Run `fitbenchmarking-cp report`. Verify info_unavaliable is shown in the report.
image

4. Run `pytest fitbenchmarking/utils/tests/test_checkpoint.py -k CheckpointTests`. Verify tests pass.

4. I generated some results in python `3.10`, `3.13` and `3.8`.

image
image
image

I was able to open most of the checkpoint files in other environments without issues (in one case there was a warning). However, I was not able to open a checkpoint file generated in python `3.13` in python `3.8`. I got this error.

image

## Checklist


fitbenchmarking/templates/css/custom_style.css
```diff
@@ -1,5 +1,5 @@
 body {
-    background-color:#4D90BA;
+    background-color:#2e2f31;
```
jess-farmer, line 2: Is there a reason for this colour change?

fitbenchmarking/templates/table_template.html
```diff
@@ -164,15 +160,18 @@ <h2 class="handle">
...
+                                                <span class='pill'>n * m</span> <b>steps</b> on the y-axis.</li>
+                                            <li>This is a Dolan-Moré performance profile.
+                                                The solvers appearing in the <b>top left corner</b> may be considered the <b>best</b>
+                                                performing on this test set.</li>
+                                            <li>See <a href="https://doi.org/10.1007/s101070100263">Dolan and Moré (2001)</a> for more information.</li>
+                                        </ul>
+                                    </div>
                                     <br>
                                     The plot is such that lines can be removed by clicking on the respective minimizer name in the dropdown
                                     or hidden by unselecting the minimizer in the legend. If a solver is removed through the dropdown,
```
jess-farmer, line 177: Is it worth also adding this paragraph to the important information box? Maybe with the final line "We recommend only analysing graphs where lines are removed via the dropdown." in bold?

fitbenchmarking/templates/table_template.html
```diff
@@ -164,15 +160,18 @@ <h2 class="handle">
...
+                                    <div class="info-box">
+                                        <p><b>Important Information:</b></p>
+                                        <ul>
+                                            <li> Fits are taken from all benchmarks, so if FitBenchmarking is run with
+                                                <span class='pill'>n</span> <b>problems</b> and <span class='pill'>m</span> <b>cost functions</b>, the resulting profile plots will have
+                                                <span class='pill'>n * m</span> <b>steps</b> on the y-axis.</li>
+                                            <li>This is a Dolan-Moré performance profile.
+                                                The solvers appearing in the <b>top left corner</b> may be considered the <b>best</b>
+                                                performing on this test set.</li>
+                                            <li>See <a href="https://doi.org/10.1007/s101070100263">Dolan and Moré (2001)</a> for more information.</li>
```
jess-farmer, line 172: I probably wouldn't have put this last line as a separate bullet point 

```suggestion
                                            <li>This is a Dolan-Moré performance profile.
                                                The solvers appearing in the <b>top left corner</b> may be considered the <b>best</b> performing on this test set.
                                                See <a href="https://doi.org/10.1007/s101070100263">Dolan and Moré (2001)</a> for more information.</li>
```

jess-farmer (commented):
This looks great @RabiyaF! The python and numpy tags work as expected and I really like the changes that you've made to the fitting report - I just have a couple of minor comments

fitbenchmarking/templates/table_template.html
RabiyaF: I think that is a good suggestion. I will try to push those changes now.

fitbenchmarking/templates/css/custom_style.css
RabiyaF: This was just a suggestion. I will leave it up to you to decide. I personally think that because the fitbenchmarking tables are very colourful, having a neutral background will be better. Having a grey background was making both the logo and the table pop more. It just focuses your attention on these elements. So that is why I updated this too.

fitbenchmarking/templates/table_template.html
RabiyaF: image

Looks much better in my opinion 👍🏼

fitbenchmarking/templates/css/custom_style.css
jess-farmer: This makes sense to me, happy to change the background to grey :)

PR #1411, opened 2025-01-10
Added check to compare jacobian computed by fitbenchmarking with that obtained through finite differences

### LINKED ISSUE #1360: Add Jacobian check
We should check Jacobians before fitting (outside of timed loop) and use scipy if not any good

Letizia97:
## Description of work
Adds a check to verify the jacobian computed through Fitbenchmarking is close enough to the one obtained through a finite diff approximation of the function.
Fixes #1360 . However, we decided to just add a warning when the relative error is large, and not add the fallback to scipy. We thought this gives the user more flexibility, as, after seeing the warning, the user should go back and check their jacobian (if they have provided one). And if the user has provided a jacobian, most likely they would not be interested in using scipy instead.

## Testing Instructions

1. Run NIST problems with analytic jacobians, but make the jacobians wrong to check warning is thrown
2. Run the above with both the option --check_jacobian and the option --dont_check_jacobian
3.

## Checklist

- [x] My PR fully fixes the issue linked. If new issues have been created, what are they? #1448 


fitbenchmarking/core/tests/test_fitting_benchmarking.py
```diff
@@ -519,19 +519,46 @@ def test_loop_over_jacobians_sparsity_check_false(
...
+                (
+                    "A relative error larger than 10^-3 was detected between "
+                    "the jacobian computed by Fitbenchmarking and the one "
+                    "obtained through a finite difference approximation. "
+                    "This might depend on either the initial parameters "
+                    "provided or the jacobian function, if this has also "
+                    "been provided by the user."
+                )
+                in log.output[0]
+            )
```
RabiyaF, line 560:
```suggestion
    @patch(f"{FITTING_DIR}.check_grad")
    @patch(f"{FITTING_DIR}.np.ptp")
    def test_check_jacobian_raises_warning(self, mock_ptp, mock_check_grad):
        """
        The test checks _check_jacobian method raises
        a warning when the jacobian check fails
        """
        with self.assertLogs(LOGGER, level="WARNING") as log:
            mock_check_grad.return_value = 0.01
            mock_ptp.return_value = 1
            self.controller.cost_func.jacobian = Analytic(
                self.controller.cost_func.problem
            )
            self.fit._check_jacobian(
                func=self.controller.cost_func.eval_cost,
                jac=self.controller.cost_func.jac_cost,
                params=[94.9, 0.009, 90.1, 113.0, 20.0, 73.8, 140.0, 20.0],
            )
            self.assertTrue(
                (
                    "A relative error larger than 10^-3 was detected between "
                    "the jacobian computed by Fitbenchmarking and the one "
                    "obtained through a finite difference approximation. "
                    "This might depend on either the initial parameters "
                    "provided or the jacobian function, if this has also "
                    "been provided by the user."
                )
                in log.output[0]
            )
```

It would be better to call `_check_jacobian` directly in this method. The original test calls `_check_jacobian` twice and this will call it once. That should be enough to verify the log contains the error msg.

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -629,3 +642,41 @@ def _perform_fit(self, controller):
...
+        max_range_finite_diff = abs(
+            max(finite_diff_grad) - min(finite_diff_grad)
+        )
+        max_range = (max_range_analytical + max_range_finite_diff) / 2
+
+        error_from_check_grad = check_grad(func, jac, params)
+        normalized_error = error_from_check_grad / max_range
+
+        if normalized_error > 10**-3:
+            LOGGER.warning(
```
RabiyaF, line 674: The warning can also include the value of the **normalized_error**.

RabiyaF (changes requested):
@Letizia97 Thank you for adding the jacobian check.I had a few comments about the changes.

RabiyaF (approved):
LGTM 👍🏼  Thank you for adding the check.

PR #1413, opened 2025-01-21
Install h5py with --no-binary in container

### LINKED ISSUE #1412: Failing Full System Tests
The full system tests are now failing down to what seems to be an issue with `h5py`: `h5py/h5t.pyx:52: ValueError: Not a datatype (not a datatype)` (full traceback [here](https://github.com/fitbenchmarking/fitbenchmarking/actions/runs/12884334841/job/35920426264))

Could be due to a version mismatch with HDF5, as the following warning is also output: `h5py is running against HDF5 1.10.10 when it was built against 1.14.4, this may cause problems`

AndrewLister-STFC:
## Description of work

Fixes #1412

The new sscanss install was installing h5py but without the required `--no-binary` flag. Hopefully this solves it!

## Testing Instructions

1.
2.
3.

## Checklist

### DIFF
--- a/Container/Containerfile
+++ b/Container/Containerfile
@@ -49,6 +49,8 @@ RUN sudo apt-get install -y cmake
 
 WORKDIR /
 
+RUN apt-get install -y pkg-config libhdf5-dev
+
 RUN git clone --depth 1 https://github.com/ISISNeutronMuon/SScanSS-2.git
 WORKDIR /SScanSS-2
 RUN python -m pip install --no-binary=h5py h5py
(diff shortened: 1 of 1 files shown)


Container/Containerfile
```diff
@@ -51,6 +51,7 @@ WORKDIR /
```
jess-farmer, line 51: It looks like the container build is failing at the moment, I think adding this line should fix things?

```suggestion

RUN apt-get install -y pkg-config libhdf5-dev

```

jess-farmer (changes requested):

jess-farmer (approved):

PR #1431, opened 2025-02-18
Added check/processing on dash component ids

### LINKED ISSUE #1304: Dash filename error
**Description of the error**
I got this when running with a problem directory in `../examples/...` I guess we can just strip leading `.._` triplets from the path?

> dash.exceptions.InvalidComponentIdError: The element `visual .._examples_benchmark_problems_InverseKinematics_-Accuracy` contains `.` in its ID.
Characters `.`, `{` are not allowed in IDs.

Letizia97:
## Description of work

This is to check whether dash component ids contain any of the characters not allowed by Dash ( "." and "{" ) and remove these characters before creating the plot.

Fixes #1304 .

## Testing Instructions

To check this, the path to the problems being run (provided in the command line) needs to include e.g. ".."  . Fitbenchmarking uses the path to determine the id to assign to Dash components. So the changes in this PR aim at removing characters not allowed by Dash (like ".") from the Dash component id, before creating the component.

I checked the changes by:
1. moving a folder with specific problems (e.g. SIF) to outside the folder "examples" . 
2. then running:    fitbenchmarking -p /home/letizia/fitbenchmarking/examples/../SIF

This only works through the changes in this PR. Otherwise Dash raises an error.

## Checklist

### DIFF
--- a/fitbenchmarking/results_processing/performance_profiler.py
+++ b/fitbenchmarking/results_processing/performance_profiler.py
@@ -433,6 +433,10 @@ class DashPerfProfile:
         self.group_label = group_label
         self.id = self.group_label + "-" + self.profile_name
 
+        dash_not_allowed_characters = [".", "{"]
+        for character in dash_not_allowed_characters:
+            self.id = self.id.replace(character, "")
+
         self.default_opt = []
         for solver in self.data.columns:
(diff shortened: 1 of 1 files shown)


RabiyaF (approved):
Thank you for adding this check 👍🏼

PR #1433, opened 2025-02-19
Updated comparison plots

### LINKED ISSUE #1374: Update the Comparison Plots
This was a point raised by Anders in one of the commercial analysis meetings at the start of the year. He pointed out that this plot is not as clear as it can be. 

image

Maybe we can consider updating either the legend or the line trace styles to make to the demarcation between the `best fit` and `others` clearer.

Letizia97:
## Description of work

Fixes #1374 

## Testing Instructions

1.
2.
3.

## Checklist

### DIFF
--- a/fitbenchmarking/results_processing/plots.py
+++ b/fitbenchmarking/results_processing/plots.py
@@ -30,7 +30,7 @@ class Plot:
         "bgcolor": "rgba(0,0,0,0.1)",
     }
     _summary_best_plot_line = {"width": 2}
-    _summary_plot_line = {"width": 1}
+    _summary_plot_line = {"width": 1, "dash": "dash"}
     _subplots_line = {"width": 1, "color": "red"}
     _error_dict = {"type": "data", "array": None, "thickness": 1, "width": 4}
 
@@ -448,6 +448,7 @@ class Plot:
         for (key, results), colour in zip(categories.items(), plotly_colours):
             # Plot category
             for result in results:
+                minim = result.minimizer
                 if result.params is not None:
                     line = (
                         cls._summary_best_plot_line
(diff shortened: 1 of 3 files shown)


jess-farmer:
Thanks @Letizia97 for making these improvements to the plots. There's just a couple of things I noticed when testing:

residual_plots

When I run FitBenchmarking with two cost functions, the legends of the residual plots overlap - it would be good to increase the spacing between the plots to avoid this if possible?

no_plots_option

When I run with the no plots option selected, I get this output where the plot would be - not sure if this is a result of the changes in this PR, but I think it would be good to only display the descriptive text when the plot is being generated, and in the case of no plots maybe an alternative sentence to say that plots haven't been generated due to selected options.

The only other comment I have is that there could be an extra test to check whether `_create_empty_residuals_plot_spinw` is called when `spinw_plot_info` has been set, but also if these functions are likely to be changed in the refactor of `plots.py` then it's probably okay to leave the tests as they are for now

Letizia97:
Thanks @jess-farmer ! I've made the changes you suggested, apart from the one on the tests. The function _create_empty_residuals_plot_spinw will indeed change in the new refactoring PR, as I'm trying to make things more general in that one (so hopefully we won't need a separate spinw function for plotting residuals), so I guess we could leave it for now. But I'll check whether a similar/equivalent check is needed in the other PR! 

Also, about the issue occurring with make_plots is set to no, in the comparison plots part I'm now displaying the same message we display for all other plots, but happy to add some text above the message if you think it needs to give more info.

jess-farmer (approved):
Thanks for the latest changes @Letizia97! All looks good to me 👍

PR #1445, opened 2025-02-25
Fixed the partial cell rendering when error flag is 2

### LINKED ISSUE #1289: Partial cell colouring for errors in compare table
When testing I found a bit of problem with displaying results. If you look at the combined table when something has errored (but with no code), you get a strange partial colouring of the cell. This is probably its own issue 

image

_Originally posted by @AnthonyLim23 in https://github.com/fitbenchmarking/fitbenchmarking/issues/1278#issuecomment-2147698398_

RabiyaF:
## Description of work

Fixes #1289

As mentioned in the issue, the cells are partially coloured when the accuracy in **infinity** and runtime is **finite**. This happens when `controller.eval_chisq` returns **infinity** because the fitting benchmark did not result in finding good parameters. The error code in this case is 2 which means that the software did not converge to a solution. 

image

This is technically not an error. I have updated the tables to show the results in this case as follows.
image

This is done so that the user can update the problem definition file to check the algorithm performance with different ranges of data or different starting conditions etc. Doing it this way also means that the result summary page is linked and the user can view the results of the fitting benchmark even though they are not great.

image

 

## Testing Instructions

I was able to recreate the bug using the `mantid` software with the `DampedGaussian` algorithm on the **MGH17** dataset.

1. Create a options file called **tmp.ini** in the main folder with the following content.
```
[FITTING]

software: mantid
          scipy_ls

num_runs: 1
max_runtime: 600

[MINIMIZERS]

mantid: Damped GaussNewton
```
2. Run `fitbenchmarking -o tmp.ini`
3. Verify the tables show the following
Screenshot 2025-02-25 at 15 15 04

## Checklist

### DIFF
--- a/fitbenchmarking/results_processing/base_table.py
+++ b/fitbenchmarking/results_processing/base_table.py
@@ -281,7 +281,11 @@ class Table:
             error_str = self.get_error_str(
                 result, error_template="<sup>{}</sup>"
             )
-            if "inf" in val_str and self.table_name != "local_min_table.":
+            if (
+                "inf" in val_str
+                and self.table_name != "local_min_table."
+                and "2" not in error_str
+            ):
                 val_str = f'<span class="blank">Error {error_str}</span>'
             elif "N/A" in val_str:
                 val_str = '<span class="blank">N/A</span>'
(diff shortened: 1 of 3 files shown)


jess-farmer (approved):
Thanks for fixing this @RabiyaF!

PR #1456, opened 2025-03-11
Refactored plots

### LINKED ISSUE #1453: Refactor plots.py

Letizia97:
## Description of work

Fixes #1453 

Plots.py has been refactored to make it easier for us to add 2D data visualisations for SpinW.

To do:
- [x] Refactor plots.py 
- [x] Remove SpinW specific functions - have these within the more general functions.
- [x] Update tests
- [x] Add new tests
- [x] Change spinw_plots_info to plots_info, and update the tests accordingly

## Testing Instructions

1. Run fitbenchmarking
2. Check comparison plots
3. Check fitting reports
4. Check plots looks okay also in SpinW case

## Checklist


fitbenchmarking/results_processing/plots.py
```diff
@@ -79,145 +74,59 @@ def write_html_with_link_plotlyjs(fig, figures_dir, htmlfile, options):
...
-                    mode="markers",
-                    name="Data",
-                    marker=self._data_marker,
-                )
-            )
-            fig.update_layout(legend=self._legend_options)
+        if self.result.plot_info is not None:
+            n_plots = self.result.plot_info["n_plots"]
+            subplot_titles = self.result.plot_info["subplot_titles"]
+            ax_titles = self.result.plot_info["ax_titles"]
```
jess-farmer, line 100: Is it worth have a separate function for setting n_plots and titles, as these lines of code are repeated a few times in different functions?

fitbenchmarking/utils/fitbm_result.py
```diff
@@ -83,6 +83,13 @@ def __init__(
             self.params = controller.final_params[dataset]
             self.accuracy = accuracy[dataset]
 
+        # if SpinW, make sure data_x is correct
+        if self.name == "Triangular AFM Powder Spectrum":
```
jess-farmer, line 87: Similar to in the parser, is there another way we can identify SpinW problems other than using the problem name? Maybe here we could check whether `ebin_cens` is a key in `problem.additional_info`?

jess-farmer (changes requested):
Thanks @Letizia97 for refactoring the plotting functions, this looks much better! I've had a look through and made a couple of comments, but I still need to try running fitbenchmarking with these changes to check plots etc.

fitbenchmarking/parsing/horace_parser.py
```diff
@@ -313,27 +313,45 @@ def _set_plot_type(self) -> None:
...
         """
-        Add q_cens and n_cuts to fitting_problem.additional_info.
+        Add plot info needed for 1d cuts (e.g. n_plots and subplot_titles)
+        to fitting_problem.additional_info.
         """
-        if self.fitting_problem.additional_info["plot_type"] == "1d_cuts":
+
+        if (
+            self.fitting_problem.additional_info["plot_type"] == "1d_cuts"
+            and self.fitting_problem.name == "Triangular AFM Powder Spectrum"
```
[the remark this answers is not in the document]
Letizia97: I think I had added that to the if condition because I thought there may be cases in the future where we want 1d cuts but the problem may not be SpinW. But I guess we can deal with this in the future if needed. I've removed that check on the problem name now

fitbenchmarking/utils/fitbm_result.py
Letizia97: I've changed this to use ebin_cens as you suggested (this however doesn't work in the case of the parser, so I've just removed the check on the name there)

PR #1476, opened 2025-04-15
Added new dropdown menus for filtering the software and cost function results

### LINKED ISSUE #993: Toggle `Cost Functions` and `Softwares` in tables
**Description**
It will be possible to toggle on/off the minimizers and problem sets in a table when #991 is merged. However, it might also be useful to toggle on/off certain cost functions and softwares.

RabiyaF:
## Description of work

Fixes #993 

https://github.com/user-attachments/assets/ba9686c2-231c-4f46-ae97-2a9273b47f09

## Testing Instructions

1. Create an options file with the following content. Name the file `tmp.ini`. And run `fitbenchmarking -o tmp.ini`.
```
[FITTING]

software: bumps
          scipy
          scipy_ls

cost_func_type: weighted_nlls
                nlls

num_runs: 1
max_runtime: 600

[MINIMIZERS]

bumps: lm-bumps
       scipy-leastsq

scipy: Powell
       CG
       TNC

scipy_ls: lm-scipy
```
2. Verify new dropdown menus are visible for all the table.
      image
3. Uncheck options from the new menus and verify that everything is updated as expected.

## Checklist

      #1478

### DIFF
--- a/fitbenchmarking/results_processing/base_table.py
+++ b/fitbenchmarking/results_processing/base_table.py
@@ -435,7 +435,7 @@ class Table:
         minimizer_template = (
             '<a class="minimizer_header" col={0} '
             'title="{1}"'
-            'data-software="{2}"'
+            ' data-software="{2}" '
             'href="https://fitbenchmarking.readthedocs.io/'
             "en/latest/users/options/minimizer_option.html"
             '#{2}" target="_blank">{3}</a>'
@@ -751,7 +751,7 @@ class Table:
         items = [
             f'        <li><label class="noselect"><input '
             f'type="checkbox" checked=true value="{minimizer}"'
-            f"onclick=\"toggle_minimizer('{software}', "
+            f" onclick=\"toggle_minimizer('{software}', "
             f"'{minimizer}')\"/> {minimizer}</label></li>"
             for software, minimizer in minimizers
         ]
--- a/fitbenchmarking/results_processing/base_table.py
+++ b/fitbenchmarking/results_processing/base_table.py
@@ -540,11 +540,17 @@ class Table:
         html = {}
         for name in [self.name, self.options.comparison_mode]:
             descrip = FORMAT_DESCRIPTION[name]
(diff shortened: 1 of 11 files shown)


jess-farmer (changes requested):
Thanks @RabiyaF for adding this functionality, I think it's a nice feature. I just have a couple of comments:

Firstly, it doesn't look like Error cells are being handled correctly - in the screenshot below I ran `fitbenchmarking` and then removed the `scipy` software from the drop down. For the Misra1c problem, Powell produced an error, and these cells haven't been removed, causing the results from each of the `scipy-ls` solvers for this problem to be shifted over.

shifted columns

Also, I know that there's already an issue addressing the fact that relative values don't recalculate when options are toggled on/off, but it might be worth adding a line at the top of the table to inform users that this is the case?

Finally, something that could probably be addressed in a different PR: at the moment a user is able to unselect all cost functions, software packages, etc. which means it's possible to completely remove the table - it might be nice to prevent the last remaining option from each drop down from being unselected/in cases where a user has only run fitbenchmarking with a single software package or cost function, the drop down menu doesn't appear.

RabiyaF:
I will have a look at a the error. I know what is causing the issue. It is looking at the href links to figure out what cells correspond to what cost functions, softwares and minimizers. In this case, the error cell does not have a href. This might not be an easy fix. This will need changing the tags and maybe refactoring the base table logic. I will have a look at this next week.

I can add the additional statement to mention that relative values are calculated by comparing all the minimizers and are not updated with they are unselected.

That would be nice to have but can be address in a different PR. I also want to remove the menus in certain cases like the Select Cost Function menu should not be visible when running benchmarking on just one cost function etc. At the moment, I intentionally implemented the same functionality as the ones from the Select Problems and Select Minimizer menus. All of these features can be added after the frontend is properly tested because it is hard to manual tests and find edge cases atm.

RabiyaF:
The issue with the error cell should be fixed now. But before adding more interactivity to the frontend, we should focus on adding tests to make sure there isn't a regression in functionality in the future.

PR #1484, opened 2025-05-08
Fix legend at the bottom of plots

### LINKED ISSUE #1473: The legend style should be updated for plots in the fitting report
In cases where the minimizer name is long, the legend takes more space on the page than the plot. 

Image

Letizia97:
## Description of work

Fixes #1473 

I've updated the legend style so that the legend appears at the bottom of plots. I've done this for both fitting report and summary page, as I think it's good for consistency and to have nicer bigger plots in both cases.

## Testing Instructions

1. Run Fitbenchmarking with minimizers having long names
2. Check plots and legends in problem summary page and fitting reports
3.
4.

## Checklist

### DIFF
--- a/fitbenchmarking/results_processing/plots.py
+++ b/fitbenchmarking/results_processing/plots.py
@@ -522,9 +522,7 @@ class Plot:
             if row_ind == 1:
                 fig.update_traces(row=row_ind)
 
-        fig.update_layout(
-            title=title + ": residuals", legend=cls._legend_style
-        )
+        fig.update_layout(title=title + ": residuals")
 
         html_fname = f"residuals_plot_for_{first_result.sanitised_name}.html"
         cls.write_html_with_link_plotlyjs(
--- a/fitbenchmarking/results_processing/plots.py
+++ b/fitbenchmarking/results_processing/plots.py
@@ -33,6 +33,12 @@ class Plot:
         "color": "rgba(0,0,0,0.4)",
     }
     _default_ax_titles = {"x": "x", "y": "y"}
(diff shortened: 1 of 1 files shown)


fitbenchmarking/results_processing/plots.py
```diff
@@ -514,7 +522,9 @@ def plot_residuals(cls, categories, title, options, figures_dir) -> str:
             if row_ind == 1:
                 fig.update_traces(row=row_ind)
 
-        fig.update_layout(title=title + ": residuals")
+        fig.update_layout(
+            title=title + ": residuals", legend=cls._legend_style
+        )
```
RabiyaF, line 527:
```suggestion
        fig.update_layout(title=title + ": residuals")
```

Is it possible to revert this change ? Because in this case having the legend on the right side results in a bigger plot. The issue with the extended names will not occur for the residual plots. Because the minimizer names won't be appended with the software and jacobian type (as in the other plots).

Screenshot 2025-05-09 at 15 30 41


Screenshot 2025-05-09 at 15 34 41

Screenshot 2025-05-09 at 15 23 43

RabiyaF (changes requested):
Thank you so much for updating this very quickly 👍🏼 I tested it and the changes look good for the plots in fitting report and the problem summary pages. I just had one small comment.

fitbenchmarking/results_processing/plots.py
Letizia97: Makes sense, thanks for checking! Commited this now

PR #1586, opened 2026-03-20
Removed restriction on codecarbon version

### LINKED ISSUE #1324: Errors when running codecarbon v2.7.1
Running FitBenchmarking with codecarbon version 2.7.1 gives the following error message: 

image

jess-farmer:
## Description of work

We have been using an old version of codecarbon because of errors reported with v2.7.1. Codecarbon have now released v3.2.5, and we should be using this latest version providing the errors are resolved. This PR fixes a bug that meant that the emissions tracker was being stopped after the first benchmark problem.

The latest version of codecarbon has been tested locally on Windows and Linux and now runs without errors. Changes should also be tested on a Mac before merging.

Fixes #1324 

## Testing Instructions

1. Upgrade codecarbon to the latest version (3.2.5) and run `fitbenchmarking`. Ensure that code runs with no errors or warnings, and the energy usage table is produced.
2.
3.

## Checklist


jfowkes:
@jess-farmer works perfectly on Mac however two major caveats:

1. The `scipy < 1.16.0` pin needs to be removed from `pyproject.toml` as there are no longer any precompiled wheels (python binaries) for SciPy this old for Mac and building SciPy from source inevitably fails
2. The removal of the above SciPy pin requires the SciPy controller to be updated (a couple of small API changes required to the controller: `maxiter` option seems to have been changed, COBYLA has changed its `func_evals` return type)
3. codecarbon asks for root privileges to access the hardware sensors on Mac, this just takes the form of a `Password:` prompt that comes out of the blue so there should be a warning displayed to the user that it may ask for a root password

I tested with the latest codecarbon release 3.2.6

jfowkes:
@Letizia97 @SamC-STFC could one of you guys create an issue for the SciPy 1.17 API bugs above if it doesn't already exist? Just try running `fitbenchmarking` without the `scipy < 1.16.0` pin in `pyproject.toml`.

jess-farmer:
> @jess-farmer works perfectly on Mac however two major caveats:
> 
> 1. The `scipy < 1.16.0` pin needs to be removed from `pyproject.toml` as there are no longer any precompiled wheels (python binaries) for SciPy this old for Mac and building SciPy from source inevitably fails
> 2. The removal of the above SciPy pin requires the SciPy controller to be updated (a couple of small API changes required to the controller: `maxiter` option seems to have been changed, COBYLA has changed its `func_evals` return type)
> 3. codecarbon asks for root privileges to access the hardware sensors on Mac, this just takes the form of a `Password:` prompt that comes out of the blue so there should be a warning displayed to the user that it may ask for a root password
> 
> I tested with the latest codecarbon release 3.2.6

@jfowkes I've now added a warning for Mac users - can you check that this appears as expected?

jfowkes:
@jess-farmer yes the warning appears as expected on Mac. Some tests are now failing though in an unrelated issue :(

jess-farmer:
> @jess-farmer yes the warning appears as expected on Mac. Some tests are now failing though in an unrelated issue :(

Thanks for checking @jfowkes! Yeah, I think the tests are failing because I also changed the `allow_multiple_runs` option to false (to prevent a separate warning about multiple instances of CodeCarbon being allowed to run at the same time) - I'll have a look into this today

pyproject.toml
```diff
@@ -32,7 +32,7 @@ dependencies = [
     'pandas>=1.3',
     'jinja2',
     'configparser',
-    'codecarbon>=2.5.1,<2.7',
+    'codecarbon>=2.5.1',
```
SamC-STFC:
```suggestion
    'codecarbon>=3.2.6',
```
If we know that 2.7.1 throws errors, should we ensure that users are not able to use it?

Do we want to use >=3.2.6 (the latest version) in the above or >2.7.1?

fitbenchmarking/core/fitting_benchmarking.py
```diff
@@ -71,6 +72,12 @@ def __init__(self, options, data_dir, checkpointer, label="benchmark"):
         self._emissions_tracker = None
         self._logger_prefix = "    "
         if "energy_usage" in options.table_type:
+            if platform.system() == "Darwin":
+                LOGGER.info(
+                    "Please be aware that you may be prompted by "
+                    "CodeCarbon to provide a password to give sudo rights "
+                    "so that the powermetrics tool can be used."
+                )
```
SamC-STFC, line 80: I'm not sure if this should be tested at all, but I've noticed that its not covered in coveralls

SamC-STFC (changes requested):
I've tested these changes locally and they seem to work with the latest codecarbon version. The changes look good with a couple of minor comments.

pyproject.toml
jess-farmer: Good point - I think >=3.2.6 would be best as I haven't tested versions between 2.7.2 and 3.2.4

fitbenchmarking/core/fitting_benchmarking.py
jess-farmer: Good point - I'll have a look to see if there's a suitable place to add a test for this

jess-farmer: I've added in a test for this, and also a few other tests to check that the EmissionsTracker object gets created and started/stopped as expected

fitbenchmarking/core/tests/test_fitting_benchmarking.py
```diff
@@ -146,6 +146,56 @@ def set_up_controller(file, options):
...
+        with self.assertLogs(LOGGER, level="INFO") as log:
+            _ = Fit(options=options, data_dir="test", checkpointer=cp)
+            self.assertTrue(
+                (
+                    "Please be aware that you may be prompted by "
+                    "CodeCarbon to provide a password to give sudo rights "
+                    "so that the powermetrics tool can be used."
+                )
+                in log.output[0]
+            )
```
SamC-STFC: Would it be possible to split this into a separate test for macos and windows since the behaviour is a little different?

SamC-STFC (changes requested):

fitbenchmarking/core/tests/test_fitting_benchmarking.py
```diff
@@ -146,6 +146,62 @@ def set_up_controller(file, options):
...
+        """
+        options = Options()
+        cp = Checkpoint(options)
+
+        _ = Fit(options=options, data_dir="test", checkpointer=cp)
+        assert emissions_mock.call_count == 1
+
+    @patch("fitbenchmarking.core.fitting_benchmarking.EmissionsTracker")
+    @patch("fitbenchmarking.core.fitting_benchmarking.platform.system")
+    def test_emissions_tracking_log_output_on_mac(self, sys_mock, _):
```
SamC-STFC, line 184: Why is `fitbenchmarking.core.fitting_benchmarking.EmissionsTracker` mocked, but then discarded as `_`?

Letizia97 (dismissed):
Thanks for adding the tests @jess-farmer ! Looks good to me .

Letizia97 (approved):
Thanks Jess, fingers crossed the integration tests will pass now

Letizia97 (approved):

SamC-STFC (approved):

PR #1591, opened 2026-03-30
Switched docs to use more recently generated plots

### LINKED ISSUE #1210: Update plots in the docs to the plotly version
Plots haven't been updated in the docs, they should be replaced with their corresponding plotly versions.

SamC-STFC:
## Description of work

Fixes #1210 
note that `nmsimplex2` was removed from GSL, so where appropriate those plots have been regenerated using `nmsimplex`.

## Testing Instructions

view the "[why fitbenchmarking](https://fitbenchmarking.readthedocs.io/en/stable/concept/why.html)" page on the docs - outdated plots should now all be plotly versions and body text should reference the updated information for the relevant plots.

Old:
image

New:
GSL_lmsder_VESUVIO

## Checklist


docs/source/concept/why.rst
```diff
@@ -30,7 +30,7 @@ Example workflow
 The black crosses on the plot below are data obtained from an experiment
 at the VESUVIO beamline at ISIS Neutron and Muon source:
 
-.. figure:: ../../images/start_for_EVS14188-90_processed_Gaussian_peaks_1_1.png  
+.. figure:: ../../images/VESUVIO_raw.png
```
jess-farmer, line 33: This first plot should also include the starting guess, rather than just the data

docs/source/concept/why.rst
jess-farmer: This line will probably need to be updated, as the default colour for the starting guess is now green

docs/source/concept/why.rst
```diff
@@ -70,28 +70,27 @@ less suited to a problem, depending on factors such as the architecture
...
 
    GSL's :code:`lmsder` (Levenberg-Marquardt) algorithm on the data
 
 We also solved the nonlinear least squares problem using GSL's implementation of
-a Nelder-Mead simplex algorithm (:code:`nmsimplex2`), which again claimed to solve
-the problem, this time in a faster 1.5 seconds.  However, this time the Chi-squared error was
-0.8505, and we plot the curve obtained in green below.  The previous curve
-is in dotted-blue, for comparison.
+a Nelder-Mead simplex algorithm (:code:`nmsimplex`), which claimed to solve
+the problem in 0.0072 seconds.  However, this time with a :math:`\chi^2` error of
```
jess-farmer: a minor comment, but I guess with the previous `nmsimplex2` minimizer it was quite nice that it claims to have solved the problem in a faster time but had a lower accuracy, which I suppose highlights that it's important to not just look at a single metric. I don't know if there are any other GSL minimizers that exhibit that behaviour?

jess-farmer (changes requested):
Many thanks @SamC-STFC for making these updates to the docs, it's been long overdue! In addition to the few minor comments, it'd also be good to know either if you can save slightly higher resolution images of the new plots (I'm not sure if you've taken screenshots or used the 'download plot as png' button?), or if we could make them smaller on the page? I think at the moment they look a bit pixelated when viewing on the docs page

docs/source/concept/why.rst
SamC-STFC: I couldnt find anything with the same behaviour in GSL, so [I have rewritten that section](https://github.com/fitbenchmarking/fitbenchmarking/pull/1591/changes/9bc135deb2be18b924b98fd9b72468f0cbe1c464) to talk about SciPy's COBYLA implementation, which does have the same behaviour (i.e. it is less accurate but faster)

jess-farmer (approved):
Looks great, happy to get this merged!

PR #1595, opened 2026-04-07
Added SASfit parser

jess-farmer:
## Description of work

Fixes #1555

This PR adds a parser for SASfit single dataset problems, with functionality to call form factors from any SASfit plugin. 

Todo:

- [x] add tests
- [x] add docs

## Testing Instructions

1. Run FitBenchmarking with the new problem set `fitbenchmarking -p examples/benchmark_problems/SAS_modelling/SASfit/`
2.
3.

## Checklist


jess-farmer:
Current questions regarding adding option for including size distributions:

- It looks like the size distribution can only be assigned to a single form factor parameter - need to figure out how to define this, and also which parameter this should be for the example problems?
- Each size distribution also has a number of parameters - need to work out how to also include these in the fitting?
- In the SASStudio functions, the `compute_scattering_contribution` function doesn't currently work when a size distribution has been set, as the variables `x` and `index_of_form_factor_param_to_smear` aren't defined - need to speak to Jase about this.

jess-farmer:
I have removed the options for adding in size distributions and structure factors as this code isn't yet fully functional in SASStudio. Instead, this PR will focus on adding functionality to run SASfit problems which contain one or more form factors.

One outstanding issue is that this parser doesn't currently deal with the case where there are duplicate parameter names across plugins - this isn't an issue for this example problem that has been added but could be an issue with other SASfit problems. This might be quite a bit of extra work, so not sure if this should be addressed in a separate PR?

PR #1596, opened 2026-04-09
Added the ability to see various metrics and run status in a tooltip over the results tables

### LINKED ISSUE #1175: Add hover text to tables
We need to pass lots of data to users to explain, for example, which Jacobians are being used.  This could be added as hover text to avoid cluttering up the tables visually. This could also be used to explain what the numbers in the cells mean.

SamC-STFC:
## Description of work

Fixes #1175 

## Testing Instructions

When the mouse is moved over any table, it should show a tooltip with the following information
image

When hovering over a cell that shows an error result, it should only show the error information
image

## Checklist


fitbenchmarking/results_processing/base_table.py
```diff
@@ -226,13 +234,25 @@ def get_str_dict(self, html=False):
...
-                self.get_str_result(r, t, html)
-                for r, t in zip(results, text_arrs)
+                self.get_str_result(result, color, html)
+                for result, color in zip(results, text_colors)
             ]
         return str_dict
 
+    def get_hover_str_dict(self):
+        """
+        Create a dict with the tooltip for each cell froom self.sorted_results
```
Letizia97:
```suggestion
        Create a dict with the tooltip for each cell from self.sorted_results
```

fitbenchmarking/results_processing/base_table.py
```diff
@@ -336,6 +356,32 @@ def get_hyperlink(self, result, val_str, text_col):
         )
         return val_str
 
+    @staticmethod
+    def get_hover_text(result: FittingResult) -> str:
+        """
+        Generate The tooltip text for a given fitting result.
```
Letizia97:
```suggestion
        Generate the tooltip text for a given fitting result.
```

fitbenchmarking/results_processing/base_table.py
```diff
@@ -384,32 +430,51 @@ def create_pandas_data_frame(self, html=False):
         :return: DataFrame with string representations of results
         :rtype: pandas.DataFrame
         """
-        str_results = self.get_str_dict(html)
         # Check all rows incase first row has missing information
```
Letizia97: This comment should be probably moved inside the format_results_indicies function since the code it refers to was moved to that function

fitbenchmarking/results_processing/base_table.py
```diff
@@ -515,7 +586,26 @@ def to_html(self):
             }
         )
 
-        return table_style.to_html(table_uuid="table")
+        hover_table = pd.DataFrame(
+            list(self.get_hover_str_dict().values()),
```
Letizia97: After seeing this line I'm wondering: is there any specific reason the result of self.get_hover_str_dict() is a dict rather than being a list? Are the keys used anywhere / should they be used?

Letizia97 (commented):
Thanks Sam for these changes! I tried running this locally and looks great. Just submitting some comments for now, but need to have a further look, especially at the formatting of the string returned by get_hover_text

fitbenchmarking/results_processing/base_table.py
SamC-STFC: Thanks for pointing this out, the keys are not used, so I have committed a change to strip them before sending the list back from `get_hover_str_dict`

fitbenchmarking/results_processing/base_table.py
```diff
@@ -2,18 +2,21 @@
...
 import matplotlib as mpl
 
 from fitbenchmarking.cost_func.cost_func_factory import create_cost_func
+from fitbenchmarking.utils.fitbm_result import FittingResult
 
 mpl.use("Agg")
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
+import pandas.io.formats.style as style
```
jess-farmer: Is there a reason that `style` is being imported separately? You should already have access to it through the pandas import on the line above

jess-farmer (commented):
This looks great, thanks @SamC-STFC! Just one very minor comment to look at

fitbenchmarking/results_processing/base_table.py
SamC-STFC: It was a a quick fix to allow me to reduce the line length when defining the return type for `style_df_for_html`. I think it now all fits within a good line length when using `pd.io.formats.style.Styler` so I have now removed the redundant import

jess-farmer (approved):
Thanks for adding this feature @SamC-STFC, all looks good to me!

PR #1603, opened 2026-04-14
Re-added error flags to runtime and compare tables

### LINKED ISSUE #1597: Runtime and Compare tables do not show error superscript
**Description of the error**
Despite having the key for the error superscript on the page, the runtime table and the compare table only show the error superscript for results where the error flag is 3 (which is a special case since it replaces the text of the cell with the word Error)

Either there should be no key for the error superscript if it is not supposed to be on those pages, or, the superscript should show as intended

There is a key, but the error flags are not showing

**Suggested fix**
Show the error flags, or remove the table with the key and rely on the tooltips added by #1175 to convey the important information

**Additional details**

Example of a case where the error flag would not represent success, but the superscript is not visible
Image

Example of the same type of error, but this time it is visible because it is on the acc table instead of the runtime table
Image

SamC-STFC:
## Description of work

Fixes #1597 

## Testing Instructions

1. View runtime or compare table
2. Observe that error superscript now appears
image

## Checklist


Letizia97 (approved):
Thanks Sam! All looks good to me.

PR #1611, opened 2026-04-30
Refactored and improved test coverage of CLI argument handling

### LINKED ISSUE #1199: Test options from options file are respected.
Recently #1197 highlighted a gap in our tests.

`cli/main.py` and other entry points.

We should test that when a user does not pass a cli argument, no value is passed to additional options.

**Is this a unit, system or functional test?**
Unit test

SamC-STFC:
## Description of work

Fixes #1199 

Moved code from main.py to a function to allow it to be unit tested in isolation. Added unit tests to test the new function.

## Testing Instructions

1. Run pytest on test_main.py

## Checklist


fitbenchmarking/cli/tests/test_main.py
```diff
@@ -57,6 +60,76 @@ def mock_func_call(*args, **kwargs):
...
+        "external_output": "",
+        "load_checkpoint": False,
+        "run_dash": False,
+        "dont_run_dash": False,
+        "check_jacobian": False,
+        "dont_check_jacobian": False,
+    }
+
+
+class cli_option_mapping:
```
Letizia97: For consistency with other classes, it'd be good to change to PascalCase i.e. `CliOptionMapping`

fitbenchmarking/cli/main.py
```diff
@@ -366,6 +366,79 @@ def get_parser():
     return parser
 
 
+def parse_options_from_cli(args) -> dict:
+    """
+    parse the command line options, and apply any options which should override
+    the options file or default options.
+    :param args: The command line arguments.
```
Letizia97, line 373: It'd be good to add a line to specify the type of the parameter args (like it's done in "run") if we can

fitbenchmarking/cli/tests/test_main.py
```diff
@@ -57,6 +60,76 @@ def mock_func_call(*args, **kwargs):
     return results, failed_problems, unselected_minimizers
 
 
+def get_default_args():
```
Letizia97: Wondering whether this would benefit from a docstring explaining these are the default _**cli**_ args, as the cli is not mentioned in the function name

Letizia97 (changes requested):
Thanks for this Sam! I've pulled your branch and checked the tests work as expected (i.e., that they fail if I change the tested function in main). I added a few very minor comments, plus one around the use of the class, and will have a better look at the code / logic once I'm back from leave. Also probably worth creating an issue to say that get_parser in main should also be tested and then link the issue in this PR.

fitbenchmarking/cli/tests/test_main.py
Letizia97: As we are using this class mainly to set 4 attributes, do you think we could simplify things and avoid using the class? Looking at the use of parametrized in FitBenchmarking, it seems that it's generally used to take in list of tuples, and I wonder whether using a class is overcomplicating the logic of the tests (though it's a very clever solution!) . Copilot is suggesting a frozen dataclass , but I don't think this will simplify things much. Probably needs more time to properly look into this

fitbenchmarking/cli/tests/test_main.py
SamC-STFC: addressed here: [37148c6](https://github.com/fitbenchmarking/fitbenchmarking/pull/1611/commits/37148c63d489ccf4402179db025e1f2c813e3e71)

fitbenchmarking/cli/tests/test_main.py
SamC-STFC: I've switched to a `dataclass` to simplify the logic - I did try moving to a tuple, but I thought that addressing the specific elements lacked clarity compared to addressing the attributes of a class (e.g. `options[0]` is not as clear as `options.input_key`)

see: [29dd34d](https://github.com/fitbenchmarking/fitbenchmarking/pull/1611/commits/29dd34d9a374d2b94975f206e34ccfb523e8a722)

fitbenchmarking/cli/main.py
SamC-STFC: I think I've now addressed this here [ef1a658](https://github.com/fitbenchmarking/fitbenchmarking/pull/1611/commits/ef1a658652225406008d5aea1b862cba97b9c85e)

fitbenchmarking/cli/tests/test_main.py
SamC-STFC: Thanks for pointing this out, I've replaced it with `OptionMapping` in latest changes (I dropped the `CLI` since the class is now also defined in side the main test case, so within the same scope it should be hard to confuse with anything else - dropping `CLI` also lets me define all of the test data without exceeding the line length limit later in the file so it looks a little nicer)

fitbenchmarking/cli/tests/test_main.py
```diff
@@ -128,3 +168,92 @@ def test_checkpoint_file_on_fail(self, get_problems, save_results):
 
         # Check that it's not empty
         self.assertTrue(contents)
+
+    @dataclass
+    class OptionMapping:
```
Letizia97, line 173: It'd be great to have a docstring here, just to make it easier in the future to understand what the class is for . I  think the docstring you had on the previous version of this class would work

Letizia97 (dismissed):
Looks great now, thanks Sam ! I just added a minor comment regarding adding a docstring for the new class, but everything else looks great.

Letizia97 (approved):
Thanks Sam!

PR #1613, opened 2026-05-05
Added a scatter plot to compare normalised accuracy and runtime

### LINKED ISSUE #1609: Create a static scatter plot for the compare table page
Create a static scatter plot that displays accuracy vs runtime for all minimizer/problem pairs, and add this as a new section on the compare table page.

### LINKED ISSUE #1616: Write tests for the new scatter plot

SamC-STFC:
## Description of work

Fixes #1609 and fixes #1616
Adds a scatter plot to fitbenchmarking which can be used to compare normalised runtime and normalised accuracy. It has been designed in such a way that it should not be complicated to introduce logic to allow the user to select which values are being compared.

The design is based on user feedback surrounding the complexity of the compare table, Ideally this new plot should provide a more user friendly way of showing the same data as the compare table

## Testing Instructions

To get to the plot:
1. Run FitBenchmarking with Dash
2. Navigate to any table page (e.g. compare table)
3. See new plot before the performance profile and after the compare table

Expected behaviours:
- click on a point to be taken to the relevant fitting report
- hover over a point to see detailed information about the point
- legend should do intersection logic when items selected from both problems and minimisers (e.g. if you have problem1 and minimiser1 selected, then only a single fitting result should be highlighted)
- when only items from one column are selected then it should do the union of all of the points with match the selected items (e.g. if you have minimiser 1 and minimiser 2 selected, all points relating to minimiser 1 and two should be selected)
- each problem should have a unique visually distinct shape
- each minimiser should have a unique color
- plot should act like a normal plotly plot (excluding the legend)
- points at 0,0 should be distinguishable and not overlap when zoomed in on
- minimizers that failed for some problems should show a warning toast when selected on the legend
## Checklist

- #1633
- #1632 
- #1640


jess-farmer:
A couple of initial thoughts (after just running the testing instructions):

1. When viewing on a laptop screen, the legend takes up a large portion of the page, meaning the plot is quite squashed (see screenshot below). I wonder if we could have the text for minimizers over multiple lines to cut down the width of the legend (e.g. a new line for Jacobiab/Hessian options)? We could also have the legend above/below the plot, or add a scroll bar to avoid it taking up too much space? 

image

2. I notice that the comparison plot has been added to every table - I think it probably makes sense to include only on the compare table as all other tables are only displaying a single metric? E.g. when clicking on the energy usage table, it's then a bit confusing to have an accuracy vs runtime plot included on the page.

docs/source/users/output/index.rst
```diff
@@ -38,10 +38,23 @@ This can be set in the option file using the
...
+Below the table there is a comparison scatter plot.
+The comparison scatter plot will only function while dash is running.
+
+The comparison scatter has a two column legend. If only items from one column
+are selected, then the union of all valid points are highlighted. If items from
+both columns are selected, then the intersection of the two colums will be 
+highlighted.
+
+Clicking on a point will take you to the fitting report page for that point.
+
```
jess-farmer: As mentioned in my previous comment, I think that the comparison scatter plot should only be featured on the compare table page, so these docs would need be moved to the docs for the compare table.

fitbenchmarking/core/results_output.py
jess-farmer, line 595: Please could you add results to the param list in the doc string?

test_options_tests.ini
jess-farmer, line 1: I think this file should be removed?

fitbenchmarking/core/results_output.py
```diff
@@ -627,7 +631,34 @@ def open_browser(output_file: str, options, pp_dfs_all_prob_sets) -> None:
...
+        max_solvers = 15
+        app = prepare_dash_app_for_performance_profiles(
+            app, options, pp_dfs_all_prob_sets, max_solvers
+        )
+        profile_instances_all_groups = create_performace_profile_instances(
+            pp_dfs_all_prob_sets
+        )
+        performance_profile_layout = get_performance_profile_layout()
+        compare_scatter = CompareScatter(app, options, results=results)
+        compare_scatter_layout, app = compare_scatter.get_layout()
```
jess-farmer: Probably not something to address in this PR, but I think it would be nice if we could have compare scatter and performance profiles set up in a similar way, i.e. rather than having functions in this file specific to the performance profile set up, moving these into the performance profile class.

fitbenchmarking/core/results_output.py
```diff
@@ -627,7 +631,34 @@ def open_browser(output_file: str, options, pp_dfs_all_prob_sets) -> None:
...
+        app = Dash(
+            __name__,
+            suppress_callback_exceptions=True,
+            external_stylesheets=[dbc.themes.BOOTSTRAP],
+        )
+        max_solvers = 15
+        app = prepare_dash_app_for_performance_profiles(
+            app, options, pp_dfs_all_prob_sets, max_solvers
+        )
+        profile_instances_all_groups = create_performace_profile_instances(
```
jess-farmer:
```suggestion
        profile_instances_all_groups = create_performance_profile_instances(
```

fitbenchmarking/core/results_output.py
```diff
@@ -802,9 +941,11 @@ def prepare_dash_app_and_run(options, pp_dfs_all_prob_sets) -> None:
         ),
     ]
 
+
+def create_performace_profile_instances(pp_dfs_all_prob_sets):
```
jess-farmer:
```suggestion
def create_performance_profile_instances(pp_dfs_all_prob_sets):
```

jess-farmer (changes requested):
Adding a few more comments, as I've now looked through `results_output.py` - will continue to add more comments as I look at more of the changes :)

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        contains the compare scatter at runtime, and to add the handling for
+        the clickthrough links
+
+        :param app: The existing dash app to add the callbacks to
+        :type Dash:
+        :param legend_items: A list of minimizer names or IDs
+        :type list[str]:
+
+        :return: The app with callbacks added
+        :rtype Dash:
```
Letizia97:
```suggestion
        :param app: The existing dash app to add the callbacks to
        :type app: Dash
        :param legend_items: A list of minimizer names or IDs
        :type legend_items: list[str]

        :return: The app with callbacks added
        :rtype: Dash
```
I've added only a couple on comments on this issue with type and rtype in the docstrings, but most docstrings in compare_scatter present the same issue. If you could fix them that would be great. Sorry that it's in many lines!

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        """
+        Given a string name of a minimizer, return a bool representing if it
+        failed to complete at least one problem (i.e. it had and error flag of
+        3)
+
+        :param item: the minimizer to check
+        :type str:
+
+        :return: whether that item threw any errors
+        :rtype bool:
```
Letizia97:
```suggestion
        :param item: the minimizer to check
        :type item: str

        :return: whether that item threw any errors
        :rtype: bool
```

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+                    Output("dummy-height", "children"),
+                    Input("compare_scatter", "figure"),
+                    prevent_initial_call=False,
+                )
+
+        else:
+            raise ValueError(
+                f"warning plot type is: {type(self.view.plot)} but go.Figure "
+                "was expected"
+            )
```
Letizia97: This kind of `if else` is typically better to have as a guard.  So, at the start of this function you would put instead:

```
if not isinstance(self.view.plot, go.Figure):
    raise ValueError(....)
```

then you can continue outside the if block, with the block starting from the line `for i, legend_item in enumerate(legend_items):`. No `else` is then needed afterwards. This is to handle invalid cases first, and keep the main logic outside the if block.

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+    inactive_opacity = 0.2
+
+    active_error_template = (
+        f"""<sup style="opacity:{active_opacity}">"""
+        """<b>{0}</b></sup>"""
+    )
+    inactive_error_template = (
+        f"""<sup style="opacity:{inactive_opacity}">"""
+        """<b>{0}</b></sup>"""
+    )
```
Letizia97, line 288: From the indentation it looks like this is not part of init?  In which case it should be moved to the top of the class, before the init

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        tooltips: list[list[str]],
+        errors: list[int],
+        minimizers: list[str],
+        problems: list[str],
+        report_pages: list[str],
+    ):
+        """
+        Get a div containing the compare scatter and legend.
+        Not that has the side effect of setting self.plot for the class as well
+        as returning the plot in a Div.
```
Letizia97:
```suggestion
        Note that it has the side effect of setting self.plot for the class as well
        as returning the plot in a Div.
```
Am I understanding the meaning correctly? Otherwise feel free to modify the docstring as needed

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+            marker={
+                "line": {
+                    "width": 0.6,
+                    "color": "#e5ecf6",  # colour of plot background
+                },
+                "size": 13,
+            },
+            showlegend=False,
+        )
+        self.plot = plot
```
Letizia97: Any reason for doing this here rather than calling it self.plot from the start, i.e. `self.plot = px.scatter ...` ?

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        legend = self.get_legend(
+            symbol_groups=problems,
+            symbol_map=self.valid_symbols,
+            colour_groups=minimizers,
+            colour_map=colour_groups,
+        )
+        self.legend = legend
```
Letizia97: I think this line should be removed and legend called self.legend from the moment it's created

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+                    )
+                else:
+                    warning_text_by_minimizer[minimizer] = (
+                        f"Warning: this minimizer failed to run on "
+                        f"{n_failed}/{n_runs} problems. Only succesful runs"
+                        " have been plotted."
+                    )
+        return warning_text_by_minimizer
+
+    problem_legend = {}
```
Letizia97: Is this line a leftover and so should it be removed?
```suggestion
```

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+            e.g. ["min1", "min1", "min2", "min2"]) each instance represents one
+            run of that minimizer
+        :type list[str]:
+
+        :return errors: A dict where the key is the minimizer name, and the
+        value is the number of times that minimizer had an error flag of 3
+        :rtype list[str,int]:
+        :return runs: A dict where the key is the minimizer name, and the
+        value is the number of times that minimizer ran
+        :rtype list[str,int]:
```
Letizia97: We should avoid having 2 returns in the docstring. Something like this would be better:

```
:return: 
    errors: A dict where ...
    runs: A dict where ...
:rtype: tuple[dict[...], dict[...]]
```
Also rtype should be dict rather than list

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        )
+        none_button_style = (
+            self.active_button_style
+            if not focus
+            else self.inactive_button_style
+        )
+        return state, all_button_style, none_button_style, plot
+
+    def set_focus_for_group(self, group, state, return_new_state=False):
+        pass
```
Letizia97: What's the reason for having this function? If it's needed somewhere else than we need to specify that in a docstring

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+            if n_failed:
+                if n_failed == n_runs:
+                    warning_text_by_minimizer[minimizer] = (
+                        "Warning: this minimizer failed to run on every "
+                        "problem and could not be plotted."
+                    )
+                else:
+                    warning_text_by_minimizer[minimizer] = (
+                        f"Warning: this minimizer failed to run on "
+                        f"{n_failed}/{n_runs} problems. Only succesful runs"
```
Letizia97:
```suggestion
                        f"{n_failed}/{n_runs} problems. Only successful runs"
```

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        the value, and the runs dictionary uses the number of occurences of the
+        minimizer name in the minimizer_names list as the value.
+
+        :param error_flags: list of error flags in same order as minimizer
+        names
+        :type list[int]:
+        :param minimizer_names: list of minimser names, including duplicates (
+            e.g. ["min1", "min1", "min2", "min2"]) each instance represents one
+            run of that minimizer
+        :type list[str]:
```
Letizia97: I can see you put in some indentation for minimizer_names but not for error_flags . Indentation really helps to read, so ideally we would have some indentation like you did with minimizer_names, both in this docstring and others

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+                    warning_text_by_minimizer[minimizer] = (
+                        f"Warning: this minimizer failed to run on "
+                        f"{n_failed}/{n_runs} problems. Only succesful runs"
+                        " have been plotted."
+                    )
+        return warning_text_by_minimizer
+
+    problem_legend = {}
+
+    def set_focus_for_all_items(self, focus, state):
```
Letizia97: It seems to me that the word focus in this function is not very clear. It's more like a state really, active or not, highlighted or not. In fact, the function builds a "state" dicitonary. I fear "focus" may be a bit misleading, for someone reading the function, because one might think about "in focus" and "out of focus", which is different than highlighted or not. Reading through the docstring, it took me some time to understand what you were referring to. Maybe the function could be renamed to something like "toggle_all_items"? Perhaps worth having a chat to find the right nomenclature to use in this function

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+            "problem": {"problem_name":True},
+        }
+
+        :param state: Dictionary with the structure described above
+        :type dict[str,dict[str,bool]]:
+
+        :return all_button_style: the updated style for the select all button
+        :rtype dict[str,any]:
+        :return none_button_style: the updated style for the select all button
+        :rtype dict[str,any]:
```
Letizia97: Similarly to a previous docstring, we shouldn't have 2 returns

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+                self.set_trace_opacity(t, self.inactive_opacity)
+        return plot
+
+    @staticmethod
+    def set_trace_opacity(t, new_opacity):
+        """
+        Given a trace, update the opacity of the text field and points.
+
+        :param t: The trace to modify
+        :type:
```
Letizia97: Missing type

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        "text-align": "left",
+    }
+
+    inactive_button_style = {
+        "display": "flex",
+        "background-color": "white",
+        "border": "none",
+        "opacity": 0.5,
+        "text-align": "left",
+    }
```
Letizia97: These definitions for `active_button_style` and `inactive_button_style` should go at the top of the class, before all the function definitions

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+
+    def get_legend(self, symbol_groups, symbol_map, colour_groups, colour_map):
+        """
+        Given a list of items which should be grouped by either symbol or
+        colour, and the colours/ symbols which they should be grouped by
+        create a legend using dash html components, that includes the name
+        of each legend item and an icon representing each hape or colour.
```
Letizia97:
```suggestion
        Receives a list of items which should be grouped by either symbol or
        colour and the colours/symbols they should be grouped by.
        Creates a legend using dash html components, that includes the name
        of each legend item and an icon representing each shape or colour.
```
Splitting into 2 sentences for readability.

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+            style={
+                "display": "flex",
+                "justify-content": "center",
+                "align-items": "center",
+            },
+        )
+
+        legend.append(html.Div(problem_legend))
+        legend.append(html.Div(minimizer_legend))
+        legend.append(dcc.Store(id="legend-status", data=legend_status))
```
Letizia97:
```suggestion
        legend = [
            html.Div(problem_legend),
            html.Div(minimizer_legend),
            dcc.Store(id="legend-status", data=legend_status),
        ]
```
For simplicity and avoiding repeating append

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        """
+
+        unique_symbol_groups = list(dict.fromkeys(symbol_groups))
+        unique_colour_groups = list(dict.fromkeys(colour_groups))
+
+        legend = []
+        legend_status = {"minimizer": {}, "problem": {}}
+
+        problem_legend = []
+        problem_legend.append(html.H2("Problem"))
```
Letizia97: In a few places you create an empty list and then use append. If you know what the list needs to contain, at least initially, use something like:


```suggestion
        problem_legend = [html.H2("Problem")]
```

Unless there's another reason for using append obviously!

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        :type list[str]:
+
+        :return complete_legend: A div containing the created legend
+        :type html.Div:
+        """
+
+        unique_symbol_groups = list(dict.fromkeys(symbol_groups))
+        unique_colour_groups = list(dict.fromkeys(colour_groups))
+
+        legend = []
```
Letizia97: `legend = []` will need to be removed if applying the suggestion regarding how `legend` is populated

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        """
+
+        return "".join([char for char in to_sanitize if char.isalnum()])
+
+    @staticmethod
+    def get_point(symbol="circle-x", colour="rgba(150,150,150,1)"):
+        """
+        Get a html div containing a single point, which is an example of the
+        provided input values. The point is non interactible, intended for
+        use in legends and embeded within other elements.
```
Letizia97: Probably worth adding a sentence here to explain why this is needed i.e. because the shapes we need are complex and need to be built this way? Otherwise in the future we might look at this and wonder why it was needed

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+        """
+        Given a string (metric), retreive the value of that metric from each
+        fitting result. Works for attributes and callables (with args able to
+        be passed using func_kwargs).
+
+        Note: this function caches the values provided to it, meaning that
+        repeated calls do not cause signifigant slowdown. In the case of
+        functions, it caches the function - not the return value, this still
+        improves performance, but means that if the function output changes,
+        the returned values are still correct.
```
Letizia97:
```suggestion
        Given a string (metric), retrieve the value of that metric from each
        fitting result. Works for attributes and callables (with args able to
        be passed using func_kwargs).

        Note: this function caches the values provided to it, meaning that
        repeated calls do not cause significant slowdown. In the case of
        functions, it caches the function - not the return value. This still
        improves performance, but means that if the function output changes,
        the returned values are still correct.
```

fitbenchmarking/results_processing/base_table.py
```diff
@@ -18,7 +18,7 @@
 import pandas as pd
 from pandas.io.formats import style
 
-from fitbenchmarking.utils.misc import get_js
+from fitbenchmarking.utils.misc import get_hover_text, get_js
```
Letizia97: Now that `get_hover_text` is imported, the old function definition in this file should be removed

fitbenchmarking/results_processing/scripts/compare_scatter/handle_link.js
```diff
@@ -0,0 +1,10 @@
+function(data){
+    if (!data) {
+        return null;
+    }
+    var newpath = data.points?.[0]?.customdata?.[3];
+    if (typeof(newpath) !== "undefined") {
+        window.parent.postMessage({path:newpath},"*");
```
Letizia97, line 8: "*" allows sending messages to any origin, while ideally we'd want to put here there the specific url we want to target, if we can. Again both Both here and in resize_observer.js

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1095 @@
...
+                    style=self.active_button_style,
+                    id=self.sanitize_for_id(symbol_mapped_value),
+                ),
+            )
+            legend_status["problem"][symbol_mapped_value] = True
+            problem_legend.extend(legend_item)
+            problem_legend.append(html.Br())
+
+        minimizer_legend = []
+        minimizer_legend.append(html.H2("Minimizer"))
```
Letizia97:
```suggestion
        minimizer_legend = [html.H2("Minimizer")]
```

Letizia97 (commented):
Thanks Sam for all your hard work! I've started having a look through this, especially at compare_scatter, but I haven't quite finished yet. Just leaving some comments based on what I've checked so far (mostly minor corrections), but need to spend some more time looking at this. Probably worth having a chat around this next week, both so I can ask some questions and explain some of the comments

fitbenchmarking/core/results_output.py
```diff
@@ -734,20 +835,26 @@ def display_page(
...
     opts = group_profiles["acc"].default_opt
 
     layout[1].options = opts
     layout[1].value = [i["label"] for i in opts[:max_solvers]]
+
     return html.Div(new_layout)
 
 
-def prepare_dash_app_and_run(options, pp_dfs_all_prob_sets) -> None:
+def prepare_dash_app_for_performance_profiles(
```
Letizia97: Calling this function this way may be a bit misleading. There are some steps done in here that prepare the dash app in general. The bit of preparation for performance profiles you've already placed in "add_perfomance_profile_callbacks", so I'd suggest either of:

- removing this function altogether and placing its content in the if block starting with `if options.run_dash:` where this function gets called
- keeping this function but moving "add_perfomance_profile_callbacks" outside of it and then renaming this function to "prepare_dash_app"

fitbenchmarking/core/results_output.py
```diff
@@ -759,7 +866,39 @@ def prepare_dash_app_and_run(options, pp_dfs_all_prob_sets) -> None:
...
+    app.layout = html.Div(
+        [
+            dcc.Location(id="url", refresh=False),
+            html.Div(id="page-content", children=[]),
+        ]
+    )
+
+    app = add_perfomance_profile_callbacks(app, max_solvers)
+
+    # Create the callback to handle multiple pages
```
Letizia97: This comment should be removed

fitbenchmarking/core/results_output.py
```diff
@@ -675,17 +706,21 @@ def check_max_solvers(opts, solvers, max_solvers):
 
 def display_page(
     pathname: str,
-    profile_instances_all_groups: "dict[str, dict[str, DashPerfProfile]]",
-    layout: "list",
+    performance_profile_instances_all_groups: dict[
+        str, dict[str, DashPerfProfile]
+    ],
+    performance_profile_layout: "list",
```
Letizia97: "list" should be changed to list without " ", both here and below (the compare_scatter_layout entry). Sorry I think this mistake was already there in the code, but we should change it since we are changing this line.

Update: needs to be figured out

fitbenchmarking/utils/misc.py
```diff
@@ -98,3 +99,48 @@ def get_js(options, working_directory):
     }
 
     return js_dict
+
+
+@staticmethod
+def get_hover_text(result, include_title=False, newline="""\\a """) -> str:
```
jess-farmer: I don't know if misc feels like the right place for this function, as it's very much part of the results processing workflow - would somewhere like base table be more appropriate?

SamC-STFC:
> A couple of initial thoughts (after just running the testing instructions):
> 
> 1. When viewing on a laptop screen, the legend takes up a large portion of the page, meaning the plot is quite squashed (see screenshot below). I wonder if we could have the text for minimizers over multiple lines to cut down the width of the legend (e.g. a new line for Jacobiab/Hessian options)? We could also have the legend above/below the plot, or add a scroll bar to avoid it taking up too much space?
> 
> image

I'm testing some ways to fix this issue. Right now I have set it so that problem names use line breaks when there are commas, and minimizer names use them for the jacobian and hessian information, I have also set the minimum width of the plot to be 2/3 of the viewport width

image


Do these changes look like enough or do you think I should also look into other things like moving the legend below the plot (which would look like the following image):
image

fitbenchmarking/core/results_output.py
SamC-STFC: seems like it was not needed - fixed here:
[2509435](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/2509435fee84241021f526ed93c880253148f287)

fitbenchmarking/core/results_output.py
SamC-STFC: resolved here using the first solution proposed: [42f8a38](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/42f8a38f1d48a9d9050c73979420eaa1c8cff382)

fitbenchmarking/core/results_output.py
SamC-STFC: resolved here: [42f8a38](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/42f8a38f1d48a9d9050c73979420eaa1c8cff382)

fitbenchmarking/results_processing/base_table.py
SamC-STFC: Resolved here: [dac6cfa](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/dac6cfa5a910d420413686adefa0e2735500ea75)

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: Does this adequately explain the reasoning? Happy change if not
[5414ae3](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/5414ae3767869d46c7c025098af0ba6f47ce76c5)

jess-farmer:
I think that adding the line breaks looks good @SamC-STFC - I would probably test on some other problem sets though to check whether this a good solution for other cases, e.g. the problem names in `SAS_modelling` are much longer, so may still mean that the plot looks quite squashed

fitbenchmarking/results_processing/scripts/compare_scatter/handle_link.js
SamC-STFC: I've looked into this a bit more, and I think since we control both the parent window and the Iframe source, and the information is not sensitive (its just stuff about the window size or URL) this doesn't seem to expose us to any risk. 

To fix it we would have to work out how to pass the dash URL into this JS file.

We also do some validation at the other end to ensure that it came from the right place, which I think is the more security critical part here. (https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/changes/BASE..12582cf8da7b926f69f137903c2c4ba64f35f0b0#diff-a01575787ef23dd0384be9cc098fdf03fbec2e71d9786d557b5ccb56c093c83fR155)

- Possibly a good one for the discussion tomorrow

test_options_tests.ini
SamC-STFC: Yep, this was a mistake, removing now

fitbenchmarking/core/results_output.py
SamC-STFC: Thanks for suggesting this - I've raised #1632 based on this comment

fitbenchmarking/utils/misc.py
SamC-STFC: I'm not sure about base table, since the plots arent tables, maybe it makes more sense to have something on the fitting result object?

I havent looked into the consequences of doing it that way, but to me it makes the most sense

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: resolved - see: [0ecb75c](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/0ecb75c38b5bf70be2662739605fb9f5b83e9505)

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: resolved - see: [0ecb75c](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/0ecb75c38b5bf70be2662739605fb9f5b83e9505)

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: Resolved - see [7693446](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/76934469b70e7683490118ba7d28bffac2ea4a0d)

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: Resolved here: [7fc949a](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/7fc949a31a9d7b85586f86cacfecd6e97fa21e93)

fitbenchmarking/core/results_output.py
SamC-STFC: Resolved here: [9a27170](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/9a27170547b7586e1bab2292bc16230b07407f86)

fitbenchmarking/core/results_output.py
SamC-STFC: Resolved here: [9a27170](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/9a27170547b7586e1bab2292bc16230b07407f86)

docs/source/users/output/index.rst
SamC-STFC: Updated to only feature on the compare table page, Still need to do docs update

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -874,6 +874,11 @@ def get_point(symbol="circle-x", colour="rgba(150,150,150,1)"):
         provided input values. The point is non interactible, intended for
         use in legends and embeded within other elements.
 
+        This is needed because plotly does not expose the shape objects for use
+        in situations like these, which means that the only way we can display
+        a single with a specific symbol (such as you would see on the legend)
```
SamC-STFC: "a single point"

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: Will update docstring to explain what focus means wherever it is mentioned

docs/source/users/output/index.rst
SamC-STFC: It looks like we dont already have manually written documentation for each table `/docs/source/users/output/compare.rst` seems to be an auto generated file from the `CompareTable` class, do we want to add content directly to this page or to create a new page, or update the docstrings in the compare table class?

jess-farmer: I was thinking that we'd update the docstring for the compare table class

SamC-STFC: I've added a short bit about it to the compare table class, and put the full text seen here in the compare scatter class (since I think it makes more sense to talk about implementation details there than in the class for the table)

fitbenchmarking/results_processing/tests/test_compare_scatter.py
```diff
@@ -0,0 +1,310 @@
...
+        for mock_result in test_data:
+            mock_result.fitting_report_link = "test/support_pages/test_link"
+        cs = CompareScatter(app, options, test_data)
+
+        urls = cs.get_fitting_report_urls()
+
+        self.assertEqual(urls[0], "support_pages/test_link")
+        self.assertEqual(urls[1], "support_pages/test_link")
+
+    def test_get_fitting_report_urls_returns_index_when_none_provided(self):
```
Letizia97: The name of this test, ending in "when_none_provided", seems to imply that get_fitting_report_urls should receive some optional input parameter (other than None), while the function in reality doesn't require any inputs. Also I'd replace "index" with "index_page" or "index_html". So overall , something along the lines of :
`test_get_fitting_report_urls_returns_index_html_when_link_missing`

fitbenchmarking/results_processing/tests/test_compare_scatter.py
```diff
@@ -0,0 +1,310 @@
...
+            {"mock_solver_0": 0, "mock_solver_1": 1},
+            None,
+        )
+
+        cs = CompareScatter(app, options, test_data)
+        self.assertFalse(cs.item_should_have_warning_toast("mock_solver_0"))
+        self.assertTrue(cs.item_should_have_warning_toast("mock_solver_1"))
+
+    # TODO: split this into one per callback - should include the following:
+    # all button, none button, mock, testMinimiser, clickthrough link,
```
Letizia97: There's a TODO here. Should this be done in this PR or should we open an issue and make the TODO changes in another PR?

fitbenchmarking/results_processing/tests/test_compare_scatter.py
```diff
@@ -0,0 +1,310 @@
...
+        self.assertEqual(
+            resize_observer_callback_args[1],
+            Output("dummy-height", "children"),
+        )
+        self.assertEqual(
+            resize_observer_callback_args[2],
+            Input("compare_scatter", "figure"),
+        )
+
+    def test_get_layout_uses_correct_information(self):
```
Letizia97, line 274: Quite coupled to the implementation order of get_values() calls. We could use assert_has_calls() or assert_any_call() where possible, as that would be more resilient to refactoring.

if :
get_values("problem_tag")
get_values("fitting_report_link")

rather than :
get_values("fitting_report_link")
get_values("problem_tag")

the test would fail.

Example improvement:

expected_calls = [
    call("hover_text", style="html", include_title=True),
    call("norm_runtime"),
    call("norm_acc"),
    call("error_flag"),
    call("modified_minimizer_name", with_software=True),
    call("problem_tag"),
    call("fitting_report_link"),
    call("modified_minimizer_name", with_software=True, unique=True),
    call("problem_tag", unique=True),
]

cs.model.get_values.assert_has_calls(expected_calls)

fitbenchmarking/results_processing/tests/test_compare_scatter_data_model.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        model = CompareScatterDataModel(self.many_result_dataset)
+        values = model.get_values("name")
+        self.assertEqual(
+            values, [result.name for result in self.many_result_dataset]
+        )
+
+    def test_get_values_works_for_callables(self):
+        """
+        Get values for axis can be provided with an axis name that links to a
```
Letizia97:
```suggestion
        get_values can be provided with an axis name that links to a
```

fitbenchmarking/results_processing/tests/test_compare_scatter_data_model.py
```diff
@@ -0,0 +1,176 @@
...
+            values,
+            [
+                result.modified_minimizer_name()
+                for result in self.many_result_dataset
+            ],
+        )
+
+    def test_get_values_respects_callable_arguments(self):
+        """
+        Get values for axis can be provided with an axis name that links to a
```
Letizia97:
```suggestion
        get_values can be provided with an axis name that links to a
```

fitbenchmarking/results_processing/tests/test_compare_scatter_view.py
```diff
@@ -0,0 +1,541 @@
...
+        self.assertEqual(sanitized, view.sanitize_for_id(sanitized))
+
+    def test_get_point(self):
+        plot_div = CompareScatterView.get_point("circle", "rgba(0,255,0,1)")
+        self.assertIsInstance(plot_div, html.Div)
+        self.assertIsInstance(plot_div.children[0], dcc.Graph)
+
+        plot = plot_div.children[0]
+
+        # check that the plot cannot be interacted with
```
Letizia97: I think this needs a bit more clarification. I know the "why" it's obvious to us now, but may not be to future developers. Just needs a sentence explaining why this should be **not** interactable.

fitbenchmarking/results_processing/tests/test_compare_scatter_view.py
```diff
@@ -0,0 +1,541 @@
...
+
+        # check that it can be called to focus
+        for i in range(num_traces):
+            _ = view.apply_state(view.plot, start_state, select)
+            self.assertEqual(
+                mock_trace_opacity.call_count, (i + 1) * num_traces
+            )
+            self.assertEqual(mock_trace_opacity.call_args.args[1], new_opacity)
+
+        mock_trace_opacity.assert_called()
```
Letizia97: Why repeating this num_traces i.e. 100 times (which then forces checking cumulative count)? Can we not check once only?  mock_trace_opacity.assert_called() will only check the last call.
Same observations apply to the test following this

fitbenchmarking/results_processing/tests/test_compare_scatter_view.py
```diff
@@ -0,0 +1,541 @@
...
+        Ensure that the function called when the All/None buttons are clicked
+        updates the "Focus" (i.e. if a minimizer/problem is selected on the
+        legend) for every minimizer and problem
+        """
+        view = CompareScatterView()
+        minimizers = ["mySolver", "mySolver", "otherSolver", "otherSolver"]
+        problems = ["problem1", "problem2", "problem1", "problem2"]
+
+        plot = self._create_test_plot(view)
+        print(plot)
```
Letizia97:
```suggestion
```

fitbenchmarking/results_processing/tests/test_compare_scatter_view.py
```diff
@@ -0,0 +1,541 @@
...
+
+        num_traces = 10
+
+        view.plot = Mock(spec=go.Figure)
+        view.plot.data = [Mock(spec=go.Trace)] * num_traces
+
+        start_state = {
+            "minimizer": dict.fromkeys(["test"] * num_traces, True),
+            "problem": dict.fromkeys(["test"] * num_traces, True),
+        }
```
Letizia97, line 513: Not sure this is having the intended effect, i.e. `dict.fromkeys(["test"] * 10, True)` is not creating 10 entries, so dict collapses to:

```
    start_state = {
        "minimizer": {"test": True},
        "problem": {"test": True},
    }
```

fitbenchmarking/results_processing/tests/test_compare_scatter_view.py
```diff
@@ -0,0 +1,541 @@
...
+        """
+        test that when apply_state is called with the value all or none, it
+        sets the opacity/focus for every trace on the plot
+        """
+        view = CompareScatterView()
+
+        num_traces = 10
+
+        view.plot = Mock(spec=go.Figure)
+        view.plot.data = [Mock(spec=go.Trace)] * num_traces
```
Letizia97: This is the same mock object repeated 10 times

fitbenchmarking/results_processing/tests/test_compare_scatter_view.py
```diff
@@ -0,0 +1,541 @@
...
+            "someFails": "1/2 failed",
+            "allFails": "all failed",
+        }
+        toasts = view.create_warning_toasts(warnings)
+
+        self.assertEqual(len(toasts), 2)
+        self.assertEqual(toasts[0].id, "allFails_toast")
+        self.assertEqual(toasts[0].children, "all failed")
+        self.assertEqual(toasts[1].id, "someFails_toast")
+        self.assertEqual(toasts[1].children, "1/2 failed")
```
Letizia97: Could we make these asserts order-independent ?

fitbenchmarking/results_processing/tests/test_compare_scatter.py
```diff
@@ -0,0 +1,310 @@
...
+        a compare scatter with two points.
+
+        App and options are Mocks of Dash and Options.
+        """
+        app = Mock(spec=Dash)
+        options = Mock(spec=Options)
+        test_data = []
+        for i in range(2):
+            # we need to set this since the model tries to access the name for
+            # sorting
```
Letizia97: Probably worth chatting about what this comment means here, and whether it should be moved to the docstring of make_mock_fitting_result instead, where the name is set

fitbenchmarking/results_processing/tests/test_compare_scatter.py
```diff
@@ -0,0 +1,310 @@
...
+from fitbenchmarking.results_processing.compare_scatter import (
+    CompareScatter,
+    CompareScatterDataModel,
+    CompareScatterView,
+)
+from fitbenchmarking.utils.fitbm_result import FittingResult
+from fitbenchmarking.utils.options import Options
+
+
+def make_mock_fitting_result(i, alt_function=False):
```
Letizia97: It would be good if we can make alt_function more descriptive

fitbenchmarking/results_processing/tests/test_compare_scatter_data_model.py
```diff
@@ -0,0 +1,176 @@
...
+            values,
+            [
+                result.modified_minimizer_name(with_software=True)
+                for result in self.many_result_dataset
+            ],
+        )
+
+    # This test is non deterministic
+    @parameterized.expand(["name", "modified_minimizer_name"])
+    def test_get_values_cache_is_faster(self, axis):
```
Letizia97: This is more of a benchmark/performance test rather than a unit test. Also, it's inherently timing-sensitive and may be flaky under CI load. Since get_values() exposes cache state via _cache_* and _unique_cache_*, could we replace this with a deterministic test that verifies cache creation/reuse directly rather than inferring it from execution time?

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1119 @@
...
+        selected_problems = [
+            g for g in state["problem"] if state["problem"][g]
+        ]
+        for t in plot.data:
+            if deselect_all:
+                visible = False
+            elif select_all:
+                visible = True
+            else:
+                minimizer = t.customdata[0][1]  # type: ignore
```
Letizia97: I think it'd be good to have a comment above this to explain why the ` # type: ignore` . We generally do in this kind of scenarios

fitbenchmarking/results_processing/tests/test_compare_scatter_view.py
```diff
@@ -0,0 +1,541 @@
...
+        self.assertEqual(num_c_grp_1, 2)
+        num_c_grp_2 = len(re.findall("colour_group_2", legend_string))
+        self.assertEqual(num_c_grp_2, 2)
+        num_s_grp_1 = len(re.findall("symbol_group_1", legend_string))
+        self.assertEqual(num_s_grp_1, 2)
+        num_s_grp_2 = len(re.findall("symbol_group_2", legend_string))
+        self.assertEqual(num_s_grp_2, 2)
+
+        # convert the legend to a string so that we can check that it contains
+        # the expected information without caring about specific structure
```
Letizia97: Comment to remove

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1119 @@
...
+            minimizer_names=self.model.get_values(
+                "modified_minimizer_name", with_software=True
+            ),
+        )
+        if item in errors:
+            return bool(errors[item])
+        else:
+            return False
+
+    def add_callbacks(
```
Letizia97: In a separate PR, probably worth splitting this into multiple subfunctions , which would also make testing easier (I think there is a TODO in the tests currently for this reason), e.g.:

```
self._register_legend_callbacks(...)
self._register_all_none_callbacks(...)
self._register_clientside_callbacks(...)
```

For example: the focus_callback function could be moved into a factory/helper function to make the captured variables more obvious when reading in the future

```
def make_focus_callback(group, has_run_failures):

    def focus_callback(_, state):
        ...
        return ...

    return focus_callback


for legend_item in legend_items:
    app.callback(...)(
        make_focus_callback(
            legend_item,
            has_run_failures
        )
    )
```

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1119 @@
...
+        self, plot: html.Div, app: Dash, legend_items: list[str]
```
Letizia97: Is `app` needed as input if we have self.app?
Later there is a line of the type : `self.app = self.add_callbacks(plot, self.app, legend_items)`, which seems to suggest we shouln't need to pass self.app

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1119 @@
...
+            self.active_button_style if focus else self.inactive_button_style
+        )
+
+        for item_type in state:
+            for item in state[item_type]:
+                state[item_type][item] = focus
+                set_props(self.sanitize_for_id(item), {"style": style})
+
+        plot = self.apply_state(
+            plot_div.children[1].figure, state, "all" if focus else "none"
```
Letizia97: Accessing the graph via plot_div.children[1] relies on the layout order remaining unchanged. Could either pass in the figure directly or locate the graph by id to reduce coupling between the layout structure and the callback logic.

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1119 @@
...
+
+        :return: The modified plot
+        :rtype: go.Figure
+        """
+
+        valid_group_types = ["all", "none"]
+        if group is not None and group not in ["all", "none"]:
+            raise ValueError(
+                f"Apply state only supports group = {valid_group_types}or None"
+                ", '{group}' was provided"
```
Letizia97:
```suggestion
                f"Apply state only supports group = {valid_group_types} or None, "
                f"'{group}' was provided"
```
f needed to avoid bug in error string

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -0,0 +1,1111 @@
...
+        """
+        self.results = results
+        # ensure consistent processing between runs
+        self.results.sort(key=self.get_sort_key)
+
+    @staticmethod
+    def get_sort_key(result: FittingResult):
+        return result.name
+
+    def get_values(self, attribute: str, unique=False, **func_kwargs) -> list:
```
Letizia97: Need to change function name to e.g. get_values_from_result

fitbenchmarking/results_processing/tests/test_compare_scatter_data_model.py
```diff
@@ -0,0 +1,176 @@
...
+
+        # if we cached the return values, the output would be the same for both
+        self.assertNotEqual(first_values, values_after_result_change)
+
+    def test_get_unique_values_gets_unique_values(self):
+        model = CompareScatterDataModel(self.duplicate_name_dataset)
+        unique_values = model.get_values("name", unique=True)
+        self.assertEqual(unique_values, ["mock_result_1"])
+
+    def test_get_unique_values_uses_different_cache(self):
```
Letizia97: Can be removed

fitbenchmarking/results_processing/tests/test_compare_scatter_data_model.py
```diff
@@ -0,0 +1,176 @@
...
+
+        first_values = model.get_values("modified_minimizer_name")
+        values_after_result_change = model.get_values(
+            "modified_minimizer_name"
+        )
+
+        # if we cached the return values, the output would be the same for both
+        self.assertNotEqual(first_values, values_after_result_change)
+
+    def test_get_unique_values_gets_unique_values(self):
```
Letizia97: get_unique_values to get_values

fitbenchmarking/results_processing/compare_scatter.py
Letizia97: Created an issue based on this

fitbenchmarking/results_processing/tests/test_compare_scatter_data_model.py
SamC-STFC: these are the timing values I get in testing this function - the speed up is substantial, but the actual times are very small both with and without it on the test dataset of 100 mock fitting results

Getting an attribute
1st run    2nd run    speed up
1.69e-05 1.31e-06 12.9x
1.61e-05 1.48e-06 10.9x
2.34e-05 1.71e-06 13.7x
1.76e-05 1.23e-06 14.3x

Getting a function
1st run    2nd run    speed up
7.68e-05 2.36e-05 3.25x
7.23e-05 2.46e-05 2.94x
6.86e-05 2.43e-05 2.82x
7.73e-05 2.36e-05 3.28x

we could possibly remove the caching logic to reduce the complexity

fitbenchmarking/results_processing/tests/test_compare_scatter.py
SamC-STFC: raised #1638

fitbenchmarking/results_processing/tests/test_compare_scatter_view.py
SamC-STFC: In this case it makes sense to check the cumulative call count because the function being tested should set the opacity for all of the traces (therefore there must be one call to set the opacity per trace)

For clarity I have removed `mock_trace_opacity.assert_called()` as it duplicates the testing inside the for loop and updated the line `self.assertEqual(mock_trace_opacity.call_args.args[1], new_opacity)` to check that it calls for the correct object

SamC-STFC: @Letizia97  could you have a look at these to check if it is ok now?  
[e59d70f](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/e59d70f8d65fc85503cb7ada233b5b6bd354d8e6)
[8847d44](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/8847d44307f52c02f96a4255adab92b4225aae75)

Letizia97: @SamC-STFC The changes in the commits you linked above look good!

Letizia97: Probably need to chat together with @jess-farmer regarding the initial comment in this thread on the for loop on num_traces instead (once she's back). For context, the commits linked above are for improving naming/docstrings in other functions

SamC-STFC: Sorry I think I linked the wrong commits, could you check the version of the test shown at the latest commit?
[78e4bc9](https://github.com/fitbenchmarking/fitbenchmarking/pull/1613/commits/78e4bc941a22f3c3954a54fbe872235765be651e)
this should be the one where I improved the testing logic

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: This ignore was supressing two warnings - the first was caused by the fact that not all traces have a `customdata` attribute. In our case the trace is always a `go.Scatter`, which does have the attribute, so we can supress this with an assertion:
```py
assert(type(t) is go.Scatter)
```

The second cannot be supressed with an assertion since it is caused by what appears to be an issue with plotly - essentially `t.customdata` is defined it have a type of `list[float]`, but various examples in the docs show it being used with strings and more complex data structures. In our case it is a list of three lists, `customdata[0]` is the hover text for the given point, `customdata[1]` is the minimizer name, `customdata[2]` is the problem name.

see: https://plotly.com/python/hover-text-and-formatting/ ("Advanced Hover Template" and "Adding other data to the hover with customdata and a hover template")

I have added the following comment to explain:
```py
# uses type: ignore here because plotly defines customdata as
# a list[float], but in our case it is a list[list[str]]
# The documentation (linked below) shows how using customadata
# to store information other than floats is idiomatic.
# https://plotly.com/python/hover-text-and-formatting/
```
@Letizia97

fitbenchmarking/results_processing/tests/test_compare_scatter_data_model.py
SamC-STFC: @jess-farmer these are the test results I mentioned in the meeting

fitbenchmarking/results_processing/compare_scatter.py
Letizia97: As discussed needs adding a bit more on what customdata[0] , etc. are

Letizia97 (approved):

jess-farmer (approved):

PR #1614, opened 2026-05-13
Added SASfit to full-tests container

Letizia97:
## Description of work

This PR adds sasfit to the full-tests Containerfile. 

The same error we get when installing SASfit locally also occurs in the container, so I had to add `|| true` to bypass that.
Building the container locally, I saw that libsasfit.so gets created as expected, which is good as that's what we need to then call the sasfit functions. 

## Testing Instructions

1.
2.
3.

## Checklist


jess-farmer (approved):
Looks good to me - many thanks for adding this @Letizia97!

PR #1622, opened 2026-06-15
Removed polyfill lines from html files

### LINKED ISSUE #1621: Weird polyfill pop-up
When trying to view FitBenchmarking result I'm getting this pop-up asking for credentials:

Image

Apparently polyfill it's a compatibility helper for old browsers, and in recent years, there have been security concerns around the polyfill.io domain. The pop-up might be caused by :

- browser extensions
- network/proxy rewriting requests
- the polyfill service behaving oddly

And shouldn't be there at all. To get rid of it / avoid this issue occurring for users it's best to remove lines like `<script src="https://cdn.polyfill.io/..."></script>` which we do have in some html files. Those lines are really not needed and can cause this issue. Removing them fixed the issue for me locally.

Letizia97:
## Description of work

Fixes #1621 

This removes the unneeded polyfill lines from the htmls. 
Polifyll is a compatibility helper for old browsers, and in recent years, there have been security concerns around the polyfill.io domain. For me locally it is causing a weird pop-up to appear, asking for credentials. Removing it from the htmls fixes the issue.

For similar reasons, also removed the js file for mathjax, and replaced it so that it is pulled from the cdn instead.

After a new release is made , we will need to modify the README to highlight security patches (#1626 ).

## Testing Instructions

1.
2.
3.

## Checklist


jess-farmer (approved):
Thanks for fixing this issue @Letizia97 - the changes look good to me!

PR #1641, opened 2026-07-27
Updated test function to use difflib library instead of custom written diff implemenation

### LINKED ISSUE #1640: Update compare_files to use inbuilt python methods
test_tables.py

test_tables.py has a custom function to create a diff between two tables, but the resulting diff isnt hugely informative and the function itself needs maintenance etc. Python has a library called `difflib` designed to do exactly this and output git style diffs (with a `+` to indicate added and `-` to indicate removed etc.)

see:
https://docs.python.org/3/library/difflib.html#difflib.Differ

It would improve the unit test output readability to use this instead, and mean that we dont have to maintain as much custom diff code

**Is this a unit, system or functional test?**
Test infrastructure

**Additional context**
Raised while working on #1613

SamC-STFC:
## Description of work

Updates `compare_files` in `test_tables.py` to use difflib to produce a git style unified diff between files instead of manually iterating though it

Fixes #1640 

## Testing Instructions

1. Run tests when files match, see that assertion passes
2. Run tests when files do not match, see that actual.out and actual.diff are generated and assertion fails

## Checklist


SamC-STFC:
I've noticed that there are may be places in our codebase where we do similar checks that could benefit from us making a helper function- see: `test_regression_two_files` in `test_checkpoint_handler`

jess-farmer (dismissed):
Thanks for making this change @SamC-STFC - having the diff file is really helpful! As discussed, it would be good to also extend this to other tests

fitbenchmarking/results_processing/tests/test_tables.py
```diff
@@ -139,49 +142,55 @@ def test_dropdown_html_correct(self):
...
+        # to pass on windows need to first do this before comparing
+        actual_output = actual_output.replace('href="..\\', 'href="../')
+
+        out_file_dir = os.getcwd() + "/actual.out"
+        diff_file_dir = os.getcwd() + "/actual.diff"
+
+        diff = list(
+            difflib.unified_diff(
+                expected_output,
+                actual_output.splitlines(keepends=True),
```
jess-farmer: I think having `keepends=True` here is causing the Windows tests to fail, as we then have `\r\n` vs the expected `\n`.

SamC-STFC: Thanks for pointing this out, I've pushed a change to format it correctly on windows

jess-farmer (approved):

PR #1642, opened 2026-07-27
Added pre commit check for common spelling/phrasing errors

### LINKED ISSUE #1624: CI Test for minimizer spelling
Currently we use an assortment of 'minimser', 'minimizer' and 'solver' - we should add a CI check/pre commit hook to standardise this. Since currently 'minimzer' is the most common, it would make sense to standardise to that where possible.

SamC-STFC:
## Description of work

Fixes #1624 

## Testing Instructions

1. See pre-commit ci test is passing for this pr because exclusions were added for the banned words
2. Create a file with either the word solver or minimiser anywhere in it and attempt to commit
3. it should raise a warning

## Checklist


.pre-commit-config.yaml
```diff
@@ -15,4 +15,16 @@ repos:
       args: []
     # Run the formatter.
     - id: ruff-format
-      args: [--check]
\ No newline at end of file
+      args: [--check]
+
+repos:
```
Letizia97: This is redefining repos, which is already defined above, effectively dropping all hooks above. This "repos" should be removed and the lines below this just appended to the above

.gitignore
```diff
@@ -13,6 +13,7 @@ docs/source/contributors/module_index
 *.DS_Store
 .vscode/
 .eggs/
+.custom_hooks/
```
Letizia97: .custom_hooks/ is added to .gitignore, but minimizer_spelling.py lives there and is committed. It works now because the file was already tracked, but any new file added to that dir will be ignored

.custom_hooks/minimizer_spelling.py
```diff
@@ -0,0 +1,65 @@
...
+        return None
+
+    for word in INVALID_WORDS:
+        index = lowercase_text.find(word)
+        if index != -1:
+            return index, word
+
+    return None
+
+def process_file(file_contents: str, filename: str) -> str | None:
```
Letizia97, line 25: Only the first error in the first file is reported — process_file returns on first match and main calls sys.exit(1) on the first file causing the problem, meaning if there's many issues one would need many fix-run cycles. It'd be better if all issues were found through a single run.

Letizia97 (changes requested):
Thanks for these changes! I've added a couple of comments on things to have a look at

jess-farmer:
As discussed with @Letizia97, it might be better to not restrict the use of the word solver, as there are some places where it's needed and maintaining `ALLOWED_NAMES` might be more hassle than it's worth. Instead I think we should just use this check to prevent inconsistent spelling of the word minimizer. Will discuss with others at the next weekly check in

jess-farmer (approved):
Happy for this to be merged!

PR #1644, opened 2026-07-27
Added ruff check to ensure that a logger is used wherever possible instead of printing

### LINKED ISSUE #1326: Include flake8-print (T20) ruleset in ruff linting tests.
[T20](https://docs.astral.sh/ruff/rules/#flake8-print-t20) will check if `print` or `pprint` statements are in the repo. There are currently about 30 `print` statements in the code. These can be replaced by using logging instead.

image

SamC-STFC:
## Description of work

Fixes #1326 

## Testing Instructions

1. use ruff with a file containing a pprint or print statement
2. see error

## Checklist


fitbenchmarking/jacobian/tests/test_jacobian.py
```diff
@@ -342,8 +342,8 @@ def test_analytic_cutest_poisson(self):
             self.fitting_problem.data_y,
             self.params,
         )
-        print(str(actual))
-        print(str(eval_result))
+        LOGGER.info(str(actual))
+        LOGGER.info(str(eval_result))
```
jess-farmer: I think these print statements can just be removed - it looks to me like they are left over from debugging something?

fitbenchmarking/results_processing/problem_summary_page.py
```diff
@@ -136,7 +136,8 @@ def _create_summary_page(
     :type options: utils.optons.Options
     """
     categories, results, descriptions = zip(*categorised_best_results)
-
+    # ignore: spelling minimiser # ignore: spelling
+    # solver
```
jess-farmer: Don't think this change belongs in this PR?

fitbenchmarking/utils/tests/test_misc.py
```diff
@@ -61,7 +64,7 @@ def test_get_css(self):
         Test that the right css files are returned
         """
         options = Options()
-        print(options.results_dir)
+        LOGGER.info(options.results_dir)
```
jess-farmer: I think this print statement can also just be removed rather introducing the logger here

jess-farmer (changes requested):
Thanks for making adding in this check @SamC-STFC!

I do think that maybe we should make all tests exempt from this rule (e.g. by adding `"**/tests/*.py"` to the per-file-ignores) as I don't think the logger ever gets set up for running the tests, so the desired output may not actually get printed? There's also a couple of print statements that I think we can just delete from the tests

jess-farmer (approved):

PR #1656, opened 2026-08-05
Fixed emissions tracker

### LINKED ISSUE #1655: stop_task() called without start_task() when a fit raises before timing starts
In Fit._perform_fit (fitbenchmarking/core/fitting_benchmarking.py), the codecarbon emissions tracker is stopped in a cleanup block that runs unconditionally:

  if self._emissions_tracker and not tracker_stopped:
      _ = self._emissions_tracker.stop_task()

  But tracker.start_task() is only reached after controller.validate() and controller.prepare() have both succeeded. If either raises — a ValidationException, or a MaxRuntimeError — stop_task()
  is called on a task that was never started.

  This only shows up when energy_usage is in table_type, which is in the default table set.

Letizia97:
## Description of work

Fixes #1655 

This PR fixes a bug in Fit._perform_fit (fitbenchmarking/core/fitting_benchmarking.py). The codecarbon emissions tracker is stopped in a cleanup block that runs unconditionally:

```
if self._emissions_tracker and not tracker_stopped:
_ = self._emissions_tracker.stop_task()
```

But tracker.start_task() is only reached after controller.validate() and controller.prepare() have both succeeded. If either raises a ValidationException, or a MaxRuntimeError ,  stop_task() is called on a task that was never started.

## Testing Instructions

1.
2.
3.

## Checklist


fitbenchmarking/core/tests/test_fitting_benchmarking.py
```diff
@@ -282,6 +282,26 @@ def test_perform_fit_error_handling(self, exp, mock):
...
+    ):
+        """
+        The test checks the emissions tracker task is not stopped when
+        the fit raised an exception before the task was started.
+        """
+        controller = set_up_controller("Gauss3.dat", self.options)
+        controller.minimizer = "Nelder-Mead"
+
+        fit = Fit(options=self.options, data_dir="test", checkpointer=self.cp)
+        assert fit._emissions_tracker is not None
```
SamC-STFC:
```suggestion
```
This assertion seems slight out of place here since it means that the test is testing if the emissions tracker is set when the constructor is called as well as testing if stop is called when an exception was thrown.

I would suggest removed it or moving to a seperate test unless there is another reason to have it here.

SamC-STFC (changes requested):

SamC-STFC (approved):

PR #1637, opened 2026-07-16
Added multifit capability (fitting multiple datasets) to more minimizers other than Mantid

Letizia97:
## Description of work

Fixes #1605 

This PR adds the multifit capability for more minimizers other than Mantid.
The only options currently incompatible are:

- analytic jacobians
- MCMC minimizers

## Testing Instructions

1. Check running a mantid multifit problem with a minimizer other than mantid
2. Check that supplying an analytic jacobian raises an error
3. Check that running with an MCMC minimizer raises an error

## Checklist

- [x] The title is of a format appropriate for a line in future release notes.
- [x] I have added the appropriate tags to the PR.
- [x] My PR represents one logical piece of work.
- [x] My PR fully fixes the issue linked. If new issues have been created, what are they?
- [x] I have added the appropriate tests to cover code that has been added.
- [x] I have updated the documentation in the relevant places to cover the changes.

PR #1701, opened 2026-08-13
Fixed bugs with non-Mantid Multifit implementation

jess-farmer:
## Description of work

This PR addresses a few bugs with the recently merged #1637:

- Adds more checks for incompatible options, as multifit problems should not currently be run with Hessians, or any cost function other than nlls ones.
- Moves the call to `multifit_init` in Fitting Benchmarking so that it is called once per minimizer, rather than once per software. `multifit_cleanup` now also clears all multifit attributes before the next minimizer is called. These functions also now work for the case where a problem has multiple sets of starting values.
- Fixes a bug in the bumps controller where incorrect x values were being passed to the solver
- Adds a helper function for more robust checking of whether a problem is a multifit one or not in `eval_r`

## Testing Instructions

1.
2.
3.

## Checklist

### DIFF
--- a/docs/source/users/problem_definition_files/mantid.rst
+++ b/docs/source/users/problem_definition_files/mantid.rst
@@ -155,6 +155,18 @@ against the data in the corresponding input file.
    the status "Validation of the provided options failed".
 
 .. warning::
+   :ref:`Hessians <hessian_option>` are not available for MultiFit problems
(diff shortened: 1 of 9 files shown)


Letizia97 (approved):
Thanks Jess, happy to merge this!

Issue #1705, opened 2026-08-13
Test callbacks for toggling metrics in the compare scatter


SamC-STFC:
**Which module and class/method/function does this relate to?**
Please provide the module and class, method or function name(s) that the tests will apply to.
CompareScatter
**What aspect requires additional tests?**
Please provide a description of the specifics of the test - what is the behaviour that the tests will inspect.  Ideally this will include a description of the setup (e.g. Pytest fixtures) and the expected results.
no tests were added during #1610 for the new callbacks added due to the complexity described in #1703 
**Is this a unit, system or functional test?**
Simply state what type of test you are expecting is required.
new test to check if the correct callbacks were added to handle the dropdowns added in #1610 

**Additional context**
Add any other context about the tests here.

SamC-STFC:
Stale, decided to add testing to pr

Issue #1703, opened 2026-08-13
Refactor add_callbacks


SamC-STFC:
**Description of the error**
The `add_callbacks` function handles multiple responsibilities in a single function, making it difficult to test individual components and maintain the code. `test_add_callbacks_adds_callbacks` is tightly coupled to the current implementation, which makes refactoring difficult.

**Describe the expected result**
- The `add_callbacks` function should be split into smaller, functions
- Each function should handle one specific type of callback addition
- The test case `test_add_callbacks_adds_callbacks` should be updated to test the refactored functions independently

Issue #1700, opened 2026-08-13
Fix minimizer spelling warnings in table_template.html


SamC-STFC:
**Description of the error**
This file uses the word solver instead of minimizer in the following locations:

```
203: data-value3="{{ n_solvers_large }}"
                       ^^^^^^

207: data-value3="{{ n_solvers_large }}"
                       ^^^^^^

214: src="{{ offline_plot }}" height={{ 100 if n_solvers_large else 650 }}></iframe>
                                                 ^^^^^^

219: <p id="profiles_info" style={{"display:none" if n_solvers_large else "display:block"}}>
                                                       ^^^^^^

227: The solvers appearing in the <b>top left corner</b> may be considered the <b>best</b> performing on this test set.
         ^^^^^^

232: or hidden by unselecting the minimizer in the legend. If a solver is removed through the dropdown,
                                                                ^^^^^^

233 the graph gets recalculated based on the best minimizer among the ones shown. On the other hand, if a solver
                                                                                                          ^^^^^^
```

note: `n_solvers_large` comes from `tables.py` which will also need to be fixed to ensure that the checks pass

**Describe the expected result**
What is the result you expect from running the steps described above?

**Describe the actual result**
What was the actual result.

**Suggested fix**

**Additional details**

SamC-STFC:
raised while working on #1610 as that required an update to the table template html

PR #1706, opened 2026-08-13
Enabled users to toggle metrics in compare scatter

### LINKED ISSUE #1610: Allow a user to toggle metrics displayed in the scatter plot
Users should be able to toggle which metrics are displayed, e.g. showing energy usage vs accuracy rather than having a fixed comparison of accuracy vs runtime.

SamC-STFC:
## Description of work

Fixes #1610

## Testing Instructions

1. Open compare table page
2. See compare scatter with two new dropdown boxes below
3. Selecting an attribute from one of the boxes should change what the corresponding axis on the scatter plot is plotting

## Checklist

- #1703 
- #1700 
- #1705 
- #1709

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/fitbenchmarking/results_processing/compare_scatter.py
+++ b/fitbenchmarking/results_processing/compare_scatter.py
@@ -240,7 +240,6 @@ class CompareScatter:
         """
         default_x = "norm_runtime"
         default_y = "norm_acc"
-
         # hover text needs to have the <extra/> tag to remove the grey box
         # that would normally show the trace name
         hover_text = [
@@ -273,6 +272,7 @@ class CompareScatter:
         ]
 
         self.add_callbacks(self.view.plot, legend_items)
+
         return plot, self.app
 
 
@@ -1126,14 +1126,14 @@ class CompareScatterDataModel:
         return result.name
 
     def get_plottable_attributes(self) -> list[str]:
-        """
-        Get a list of attributes which can be plotted on the compare scatter
-        plot. This is determined by checking if the attribute is present in the
-        FittingResult class and if it is a number or a callable function.
 
-        :return: List of attributes which can be plotted
-        :rtype: list[str]
-        """
+        excluded_attributes = [
+            "error_flag",
+            "get_n_data_points",
+            "get_n_parameters",
+            "init_blank",
+        ]
+
         all_attributes = []
         for result in self.results:
             all_attributes.extend(dir(result))
@@ -1141,20 +1141,30 @@ class CompareScatterDataModel:
         unique_attributes = list(dict.fromkeys(all_attributes))
         plottable_attributes = []
         for attribute in unique_attributes:
-            if not attribute.startswith("_"):
+            if (
+                not attribute.startswith("_")
+                and attribute not in excluded_attributes
+            ):
                 try:
                     values = self.get_values_from_results(attribute)
+                    if attribute == "norm_acc":
+                        pass
+                        # print(values)
+                        # print([type(value) for value in values])
 
-                    if any(
-                        isinstance(value, numbers.Number) for value in values
-                    ) and not all(value == np.inf for value in values):
+                    if self.list_contains_plottable_types(values):
                         plottable_attributes.append(attribute)
-
                 except TypeError:
                     # callable, but requires arguments, so we cannot plot it
                     continue
-
         return plottable_attributes
+
+    @staticmethod
+    def list_contains_plottable_types(list):
+        return any(
+            isinstance(value, numbers.Number) and not isinstance(value, bool)
+            for value in list
+        ) and not all(value for value in np.isinf(list))
 
     def get_values_from_results(
         self, attribute: str, unique=False, **func_kwargs
--- a/fitbenchmarking/results_processing/compare_scatter.py
+++ b/fitbenchmarking/results_processing/compare_scatter.py
@@ -1195,40 +1195,33 @@ class CompareScatterDataModel:
         :rtype: list[str]
         """
 
-        excluded_attributes = [
-            "error_flag",
-            "min_accuracy",
-            "min_energy",
-            "min_first_runtime",
-            "min_harmonic_runtime",
-            "min_maximum_runtime",
-            "min_mean_runtime",
-            "min_median_runtime",
-            "min_minimum_runtime",
-            "min_trim_runtime",
-            "init_blank",  # needed to avoid accidentally clearing every result
+        plottable_attributes = [
+            "accuracy",
+            "energy",
+            "first_runtime",
+            "func_evals",
+            "get_n_data_points",
+            "get_n_parameters",
+            "harmonic_runtime",
+            "iteration_count",
+            "maximum_runtime",
+            "mean_runtime",
+            "median_runtime",
+            "minimum_runtime",
+            "norm_acc",
+            "norm_energy",
+            "norm_runtime",
+            "runtime",
+            "trim_runtime",
         ]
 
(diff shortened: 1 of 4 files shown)


fitbenchmarking/results_processing/tests/test_compare_scatter_view.py
```diff
@@ -556,3 +588,17 @@ def test_apply_state(self, start_state, mock_trace_opacity: Mock):
             self.assertEqual(mock_trace_opacity.call_args.args[1], new_opacity)
 
         mock_trace_opacity.assert_called()
+
+    def update_fig_axes_makes_correct_changes(self):
```
Letizia97: This test is missing the "test" prefix so unittests skips it

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -502,6 +570,27 @@ def get_per_minimizer_errors_and_runs(error_flags, minimizer_names):
 
         return errors_by_minimizer, runs_by_minimizer
 
+    def update_fig_axes(
```
Letizia97: Looks like this function is doing more than updating the axes, as it's basically replotting things. Maybe worth improving the name and adding a docstring?

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -502,6 +570,27 @@ def get_per_minimizer_errors_and_runs(error_flags, minimizer_names):
...
+        self, x_title=None, x_data=None, y_title=None, y_data=None
+    ):
+        if x_title is not None:
+            assert x_data is not None, (
+                "x_data must be provided when x_title is provided"
+            )
+            self.plot.update_layout(xaxis_title=x_title)
+            for i, t in enumerate(self.plot.data):
```
Letizia97: This for loop assumes one trace per result, in results order. It's fine whenever every (minimizer, problem) pair is unique, but if multiple cost functions are used it might become a problem

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -403,12 +443,41 @@ def get_plot(
...
+            ),
+            html.Div(
+                [
+                    html.Div(
+                        "X axis attribute:",
+                        style={"padding-right": "5px", "padding-left": "5px"},
+                    ),
+                    dcc.Dropdown(
+                        plottable_attributes,
+                        value=[x_title],
```
Letizia97:
```suggestion
                        value=x_title,
```
Dropdown doesn't set multi=True, so value should be the string itself. If left as is, the dropdown won't show the default metric as selected, even though the plot is drawn with norm_runtime/norm_acc. Once you pick something the callback receives a plain string and everything downstream behaves, so it's a first-paint problem rather than a persistent one.

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -403,12 +443,41 @@ def get_plot(
...
+                        id="x-dropdown",
+                        clearable=False,
+                    ),
+                    html.Div(
+                        "Y axis attribute:",
+                        style={"padding-right": "5px", "padding-left": "5px"},
+                    ),
+                    dcc.Dropdown(
+                        plottable_attributes,
+                        value=[y_title],
```
Letizia97:
```suggestion
                        value=y_title,
```

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -1040,6 +1129,119 @@ def __post_init__(self):
...
+        :param values: list of values to check
+        :type values: list
+
+        :return: if the list can be plotted on a scatter plot
+        :rtype: bool
+        """
+        return any(
+            isinstance(value, numbers.Number) and not isinstance(value, bool)
+            for value in values
+        ) and not all(value for value in np.isinf(values))
```
Letizia97: Right now, this excludes  func_evals and iteration_count when these appear like mixed lists containing None, e.g. [20, 15, 20, None, 20, 20, 20, 15, 20, 20, None, 20, 20, 15, None] . So if a minimiser reports None for n_evals, then no matter whether the other minimizers do, n_evals is removed from the plot.
Not sure whether there's any way around this

fitbenchmarking/results_processing/compare_scatter.py
```diff
@@ -1040,6 +1129,119 @@ def __post_init__(self):
     def get_sort_key(result: FittingResult):
         return result.name
 
+    def get_plottable_attributes(self) -> list[str]:
```
Letizia97, line 1190: I wonder whether we should invert the logic and have a list of "compare_scatter_attribute" rather than a list of excluded ones, just because of the potential consequences in the case where we add more attributes which are a no for this in the future. What do you think?

Letizia97 (changes requested):
Thanks Sam! I've added a few comments on things to have a look at. Also looks like the stack needs rebasing now, can you check whether it's as simple as clicking on rebase or whether more needs to be done to rebase the stack? Thanks!

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: Personally I prefer it the way it is for the following reasons:
- Renaming variables in the fitting result class wont have a chance of accidentally removing them for the compare scatter if the developer is not aware of this relationship
- New attributes will be automatically plotted and formatted semi (again this happens automatically so doesnt require developers to know the implementation details of this class)
- If we accidentally plot something which doesnt make sense (e.g. error flag) that is much less annoying than if we accidentally dont plot something which we should (e.g. runtime)

All this assumes that we follow naming conventions (using _ for private variables) and are more likely to add attributes that we do want to be plotted than to add ones that we dont want to be plotted - you have a lot more experience with fitbechmarking than me, so I think you should get the final call on whether we need to make any changes here

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: Im not sure that this is how it works. To my understanding the `any` function means that it will return true if any of the values in `values` is not an number and not a bool whereas the `all` function would behave as you are suggesting

image

Letizia97: No, it's the np.isinf that's causing the problem here

SamC-STFC: I see - this commit should fix it [c6c719a](https://github.com/fitbenchmarking/fitbenchmarking/pull/1706/commits/c6c719a2cd66ff520e8e3cf1ed24b9bb39a9877a)

SamC-STFC: I'll add test coverage for this in another commit

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: @jess-farmer We have discussed this offline and would like your opinion on whether attributes should be excluded via a blacklist or if a whitelist of permitted attributes should be used instead

fitbenchmarking/results_processing/compare_scatter.py
SamC-STFC: I've renamed it to `update_axes_data` and added a docstring

Letizia97 (approved):
Thanks Sam, looks good to me!

PR #1729, opened 2026-08-24
Updated docutils call to use `writer` attribute instead of deprecated `writer_name`

### LINKED ISSUE #1727: Deprecation warning in unit tests
Unit test runs give this warning
```
/__w/fitbenchmarking/fitbenchmarking/fitbenchmarking/results_processing/base_table.py:620: PendingDeprecationWarning: Argument "writer_name" will be removed in Docutils 2.0.  Specify writer name in the "writer" argument.
```

It does not cause the test to fail but we should look at doing what it says

SamC-STFC:
## Description of work

Fixes #1727 

## Testing Instructions

1. Check that docutils build succeeds
2. Ensure that docs are in format expected for html

1. Check `Default System Tests (Linux)`
2. Ensure that the following warning is not present
```
/__w/fitbenchmarking/fitbenchmarking/fitbenchmarking/results_processing/base_table.py:620: PendingDeprecationWarning: Argument "writer_name" will be removed in Docutils 2.0.  Specify writer name in the "writer" argument.
```
## Checklist


SamC-STFC:
The docs build took an unusually long time, which caused the other tests to be cancelled automatically  - I will look into if the change is causing that before marking as ready for review

SamC-STFC:
> The docs build took an unusually long time, which caused the other tests to be cancelled automatically - I will look into if the change is causing that before marking as ready for review

Tested it a second time and the build took a normal amount of time, I think it may have just failed to report it correctly

Letizia97 (changes requested):
Thanks Sam! Looks like this requires docutils>=0.22, so we should add that to the dependencies list in pyproject.toml

SamC-STFC:
The docs build is failing on this PR, I'm not sure why so I've converted this to a draft while we investigate

SamC-STFC:
I've fixed the Docs build issues - it was caused by the fact that the sphinx version used on readthedocs doesnt support the latest docutils version

PR #1738, opened 2026-08-26
Added cost function to hover text when more than one is in use

### LINKED ISSUE #1732: Cant differentiate between multiple cost functions on compare scatter
**Description of the error**
A clear and concise description of what the problem is, including steps to reproduce it and the environment you are running FitBenchmarking in.
The compare scatter plots one point for each cost func/minimizer/problem combination, but there is no iconography or text on the scatter itself explaining what cost function relates to what point.

What is the result you expect from running the steps described above?
Ideally you should be able to determine cost function by looking at the hover text or some other iconography surrounding the point

SamC-STFC:
## Description of work

Fixes #1732 

## Testing Instructions

1. Run fitbenchmarking with two cost functions (e.g. `-c weighted_nlls nlls`)
2. See cost functions in hover text when hovering over point on compare scatter
3. 
image
image

1. Run fitbenchmarking with one cost function
2. See that the cost function is not included in the hover text

## Checklist


SamC-STFC:
I've added unit tests for all of the hover text function as well as just the `include_cost_func` param since they were missing

fitbenchmarking/utils/fitbm_result.py
```diff
@@ -598,16 +604,33 @@ def hover_text(self, include_title=False, style="html") -> str:
...
+                    f"{bold_start}"
+                    f"{self.modified_minimizer_name(with_software=True)}"
+                    f"{bold_end}"
+                    f" | "
+                    f"{bold_start}"
+                    f"{self.problem_tag}"
+                    f"{bold_end}"
+                    f"{line_break}"
+                    f"{hover_text}"
+                )
```
Letizia97:
```suggestion
            title_parts = [
                self.modified_minimizer_name(with_software=True),
                self.problem_tag,
            ]
            if include_cost_func:
                title_parts.append(self.costfun_tag)

            title = " | ".join(
                f"{bold_start}{part}{bold_end}" for part in title_parts
            )
            hover_text = f"{title}{line_break}{hover_text}"                
```

fitbenchmarking/utils/fitbm_result.py
Letizia97: Just realised this should be `self.iteration_count` rather than `self.func_evals`. Could probably be fixed here as there's a matching comment in the test for this

fitbenchmarking/utils/tests/test_fitbm_result.py
```diff
@@ -641,6 +643,117 @@ def test_fitting_result_gets_mask_from_problem(self):
...
+
+        self.assertIn(f"Status: {result.status}", actual)
+        self.assertIn(f"Accuracy: {result.accuracy:.4g}", actual)
+        self.assertIn(
+            f"{result.runtime_metric.capitalize()} runtime: "
+            f"{result.runtime:.4g}",
+            actual,
+        )
+        self.assertIn(f"Energy usage: {result.energy:.4g}", actual)
+        self.assertIn(f"Iterations: {result.func_evals}", actual)
```
Letizia97:
```suggestion
        self.assertIn(f"Iterations: {result.iteration_count}", actual)
```

fitbenchmarking/utils/fitbm_result.py
```diff
@@ -572,6 +574,10 @@ def hover_text(self, include_title=False, style="html") -> str:
             newline used in tables
         :type newline: str
```
Letizia97: We should update this docstring to:
- remove the `result ` and `newline` parameters as they don't appear in the function signature anymore
- add `style` instead

Letizia97 (changes requested):

Letizia97 (approved):

PR #1770, opened 2026-09-09
Fixed Poisson cost func error around residuals plots

### LINKED ISSUE #1612: FitBenchmarking fails when poisson cost_function selected
I've tried running the NIST low_difficulty problems with the poisson cost function and a number of minimizers. In all cases, I get the following error (this was obtained with bumps, newton).

Image

It looks like the error occurs at plotting level, as no error is printed while minimisers are running.
So far, I've encountered this with all minimizers I tried:

- newton and amoeba from bumps
- Newton-CG and BFGS from scipy
- migrad from minuit

SamC-STFC:
## Description of work

Fixes #1612

## Testing Instructions

1. run fitbenchmarking with `-c poisson` `-p some_set_of_problems_with_negative_values`
2. see warnings in the log explaining that the poisson cost func is incompatible
<br>

1. run fitbenchmarking with `-c poisson` `-p some_set_of_problems_without_negative_values`
2. runs without error
3. residuals plot not plotted
4. info box explaining why is shown
<br>

1. run fitbenchmarking with `-c poisson nlls` `-p some_set_of_problems_without_negative_values`
2. runs without error
3. residuals plot is plotted, but only for nlls cost func
4. info box explaining the missing plot is shown

## Checklist


docs/source/users/cost_func.rst
```diff
@@ -40,6 +40,10 @@ The cost functions that are currently supported are:
     .. currentmodule:: fitbenchmarking.cost_func.poisson_cost_func
     .. autoclass:: fitbenchmarking.cost_func.poisson_cost_func.PoissonCostFunc
                :noindex:
+    .. note::
+         The Poisson deviance cost function is only valid for problems with non-negative data.
+         In addition to this, the Poisson deviance cost function cannot calculate residuals,
+         so the residuals plot will not be available when using this cost function.
```
Letizia97:
```suggestion
         In addition to this, FitBenchmarking does not report residuals for this cost function.
         The Poisson deviance is a sum of per-point contributions rather than a sum of squares,
         so its residuals are not directly comparable with those of the least-squares cost 
         functions. Therefore, the residuals plot will not be available when using this cost 
         function.
```

fitbenchmarking/templates/problem_summary_page_template.html
```diff
@@ -214,17 +214,25 @@ <h3>Summary</h3>
...
+                                hide/show the respective data.
+                            </p>
+                            <div align="center" class="figure align-center">
+                                <iframe width="100%" height="500" frameborder="0" seamless="seamless" scrolling="no" src="{{ residuals_plot }}"></iframe>
+                            </div>
+                        {% endif %}
+                        {% if any_residuals_plot_missing %}
+                            <div class="info-box">
+                                <p><b>Important Information:</b></p>
+                                Residuals are are not calculated for non-nlls cost functions, so they could not be plotted.
```
Letizia97:
```suggestion
                                Residuals are not calculated for non-nlls cost functions, so they could not be plotted.
```

fitbenchmarking/cost_func/poisson_cost_func.py
```diff
@@ -63,7 +66,13 @@ def eval_cost(self, params, **kwargs):
             )
         f_xp = self.problem.eval_model(x=x, params=params)
 
-        # Penalise nagative f(x, p)
+        if (f_xp[f_xp <= 0.0]).any():
```
Letizia97:
```suggestion
        if (f_xp <= 0.0).any():
```
Might be wrong about this,  but I think this would otherwise return the matching values rather than the mask, and .any() on numbers treats 0.0 as false

Letizia97:
Thanks for these changes @SamC-STFC  . There's a couple of things in this PR that will need discussing with Jess or Jari I think, mainly because I think the work in this PR gets slightly affected by the eval_r changes made during the multifit work (as we were discussing earlier). Basically I think the reason the code can't compute/plot residuals for this cost function is partly to do with how we wrote eval_r. So probably worth a discussion with people before we merge it. It's looking good otherwise, consider it done! Will be discussing a few bits with people next week and then just merge this, no more work needed on your end

fitbenchmarking/cost_func/poisson_cost_func.py
```diff
@@ -63,7 +66,13 @@ def eval_cost(self, params, **kwargs):
...
+            LOGGER.error(
+                "Found negative values while evaluating poisson cost "
+                "function. Negative values have been replaced with "
+                "np.finfo(float).max so that the cost cannot be evaluated."
```
Letizia97:
```suggestion
                "Found negative values while evaluating poisson cost "
                "function. Zero or negative values have been replaced with "
                "the largest representable float, so the cost is heavily penalised there."
```

fitbenchmarking/cost_func/poisson_cost_func.py
```diff
@@ -63,7 +66,13 @@ def eval_cost(self, params, **kwargs):
             )
         f_xp = self.problem.eval_model(x=x, params=params)
 
-        # Penalise nagative f(x, p)
+        if (f_xp[f_xp <= 0.0]).any():
+            LOGGER.error(
```
Letizia97: This will tend to flad the logs, maybe better have a "_warned_non_positive" flag defined at class level, and turn it into True once the first warning is raised

PR #1771, opened 2026-09-09
Fixed unusual scroll behaviour on tables

### LINKED ISSUE #1764: Hover text causes unusual scroll behaviour on tables
Image

The pages seem to reserve more space for the tables than needed so that they can fit the hover text inside the same container as the table this means that users can scroll past the edge of the table into the space reserved for the hover text. This is weird UX, but probably not too big of an issue for users.

It would be good if the hover text was seperated from the container somehow and was able to display over the edges, or if was positioned to minimise this effect (e.g. positioned on the left side of a result instead of the right)

SamC-STFC:
## Description of work

Fixes #1764

## Testing Instructions

1. Open any table
2. Scroll to rightmost/bottom column
3. check that hover text is entirely contained within the table area
4. check that you cannot scroll further out to reveal whitespace beyond the bounds of the table

For reference,
Before change:
image

After change:
image

## Checklist


Letizia97 (commented):
Thanks @SamC-STFC ! This fixes the issue on large tables, but on a small table I'm getting this kind of behaviour now:

Image

Letizia97 (changes requested):

SamC-STFC:
```css
tr:nth-last-child(n+10) ~ tr:nth-last-child(-n+5){
	color:red
}
```

Letizia97 (approved):

PR #1780, opened 2026-09-18
Replaced hogben regression test

### LINKED ISSUE #1654: GSL sometimes fails regression test
Most of the time, the full system tests pass as expected, however on some runs GSL fails for the HOGBEN simple sample problem (returning inf instead of the expected accuracy value).

Letizia97:
## Description of work

This is to replace a regression test, because GSL sometimes fails on the simple sample hogben regression test.

Fixes #1654

## Testing Instructions

1.
2.
3.

## Checklist


Letizia97:
Have set tolerance to be slightly higher **specifically and only** on hogben tests, but it's getting to the point where it needs to be 1e-1, which maybe is a bit too high. I've tried replacing the same test with a few different hogben examples and in every case the tolerance needs to be 1e-1 to allow tests to always pass (as opposed to sometimes fail, sometimes pass). Maybe we should consider whether we want to keep testing on these examples

